PR #48, opened 2026-04-09
Codex/sync upstream 20260409

yibie:


FodorDani:
This introduces/references curl ... | bash installation from external repo   (yibie/caveman-codex) and supply-chain style remote fetch/install flow. This is explicit and reviewable, but still high-risk pattern.

FodorDani:
the install.sh script it targets is also a High-risk, lines:

 1. install.sh is meant to run via curl ... | bash from mutable main   - https://github.com/yibie/caveman-codex/blob/main/install.sh#L7-L8
 2. Download source is unpinned (REF_DEFAULT="main")   - https://github.com/yibie/caveman-codex/blob/main/install.sh#L16
 3. Remote base URL is constructed from repo/ref and can be overridden   - https://github.com/yibie/caveman-codex/blob/main/install.sh#L87-L89
 4. It fetches plugin files with curl and no checksum/signature verification   - https://github.com/yibie/caveman-codex/blob/main/install.sh#L112-L113

So exploit is: if upstream repo/branch/content is compromised (or redirected via --raw-base), users execute attacker-controlled install/plugin payload. This is a real RCE supply-chain risk

JuliusBrussee:
Thanks @yibie — Codex delivery has moved to the unified profile/native-pack architecture, and this sync branch targets deleted surfaces. Closing as superseded.

PR #49, opened 2026-04-09
fix(compress): resolve runtime path and add retry failure behavior

xiaolai:
## Summary

- Replace unresolvable `<directory_containing_this_SKILL.md>` placeholder with concrete `caveman-compress/` path and a glob fallback
- Define behavior after 2 failed retries: report error, leave original file untouched
- Align `name: compress` → `name: caveman-compress` to match trigger `/caveman:compress` and directory name

## Why

The path placeholder in step 1 is not resolvable at runtime — the model has no mechanism to derive the skill's install directory from "Find that directory." The missing failure path leaves the user in an undefined state after retries exhaust.

Spotted while running the skills through [nlpm](https://github.com/xiaolai/nlpm-for-claude).

## Test plan

- [ ] Invoke `/caveman:compress` on a test markdown file
- [ ] Confirm `scripts/` directory is found without manual intervention
- [ ] Force a compression failure and verify original file is preserved

### DIFF
--- a/caveman-compress/SKILL.md
+++ b/caveman-compress/SKILL.md
@@ -1,5 +1,5 @@
 ---
-name: compress
+name: caveman-compress
 description: >
   Compress natural language memory files (CLAUDE.md, todos, preferences) into caveman format
   to save input tokens. Preserves all technical substance, code, URLs, and structure.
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Closing — already on main. SKILL.md has concrete `caveman-compress/scripts/` path and retry failure behavior documented.

PR #50, opened 2026-04-09
fix(caveman): replace "user confused" with observable triggers

xiaolai:
## Summary

Auto-Clarity section: "user confused" → "user asks to clarify or repeats question"

## Why

"Confused" is an internal mental state the model can't observe directly. Observable triggers — clarification requests, repeated questions — produce more consistent behavior across sessions.

## Test plan

- [ ] Activate caveman mode, ask an ambiguous question, then say "wait, what?" — verify caveman drops to clear mode
- [ ] Verify caveman stays active when user does not signal confusion

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -49,7 +49,7 @@ Example — "Explain database connection pooling."
 
 ## Auto-Clarity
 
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Closing — already on main. SKILL.md Auto-Clarity trigger already reads "user asks to clarify or repeats question."

PR #62, opened 2026-04-09
fix: correct max_tokens typo from 8096 to 8192

sjhddh:
`max_tokens=8096` is likely a typo for `8192` (2^13). Other scripts in the repo use `4096` (2^12). While the API accepts any integer, this aligns the value with the intended power-of-two convention.


Borda (approved):

JuliusBrussee:
Closing — already on main. `compress.py` has `max_tokens=8192` (fixed in earlier commit).

PR #63, opened 2026-04-09
fix: use consistent o200k_base tiktoken encoding in benchmark.py

sjhddh:
benchmark.py uses cl100k_base while the rest of the project (measure.py, plot.py) uses o200k_base. This causes token counts from benchmark.py to be measured on a different scale than the eval harness. Aligns with the project's documented tokenizer choice.


Borda (approved):

JuliusBrussee:
Closing — already on main. Both `benchmark.py` and `measure.py` use `o200k_base`.

PR #66, opened 2026-04-10
fix: handle nested fences + strip LLM output wrapper in caveman-compress

rookledookle:
## Summary

Two bugs caused `caveman-compress` to reliably fail on files that contain fenced code blocks showing markdown (e.g. report templates, output format examples, docs about markdown). Both are fixed here with no behavior change for the happy path.

### 1. Fence parser mispairs nested / variable-length fences

`validate.py` used:

```python
CODE_BLOCK_REGEX = re.compile(r"```.*?```", re.DOTALL)
```

This pairs fences greedily left-to-right. When a file has an outer 4-backtick fence wrapping an inner 3-backtick snippet (common in docs about markdown and in report templates that illustrate code output), the regex parses it as two separate blocks — so `extract_code_blocks(original) != extract_code_blocks(compressed)` even when nothing inside changed. Tilde fences (`~~~`) weren't supported at all.

**Fix:** replaced with a line-based fence parser following CommonMark rules. A fenced block opens on a line starting with 3+ backticks or tildes (with up to 3 leading spaces), and closes on a line with a matching fence char of at least the same length and no info string. Variable-length fences, tilde fences, and nested blocks all parse correctly.

### 2. Claude wraps full output in `` ```markdown ... ``` ``

`compress.py` passes Claude's stdout straight through without stripping any wrapper. The model frequently wraps its entire response in a fence (even when prompted not to), which:

- fails code-block validation (inner real blocks no longer match the original)
- burns the retry budget on a fix that can't succeed
- ends with the file restored and the user thinking compression "just doesn't work on this file"

**Fix:** added `strip_llm_wrapper()` that removes an outer `` ``` ``-or-`~~~` fence only when it wraps the entire output. Inner fences preserved exactly. Applied to both the `anthropic` SDK path and the `claude` CLI fallback. Also strengthened the compression prompt with an explicit "do not wrap the output in an outer fence" rule.

## Test plan

- [x] Unit-style smoke test on a file with three fenced blocks (one ` ``` `, one tilde, one 4-backtick wrapping 3-backtick) — all three extract correctly
- [x] `strip_llm_wrapper()` round-trips ` ```markdown ... ``` ` correctly, preserves inner ` ```py ... ``` ` exactly
- [x] End-to-end run on a real ~6KB command file with two fenced output-format blocks — validation passes on first attempt where the old parser would hit max retries and restore

## Notes

- No change to the extractor's public API — still returns `list[str]`
- `strip_llm_wrapper()` is conservative: it only strips when a single outer fence wraps the *entire* stripped output, so normal responses without a wrapper are untouched
- Happy to split into two commits if preferred

### DIFF
--- a/caveman-compress/scripts/compress.py
+++ b/caveman-compress/scripts/compress.py
@@ -7,9 +7,29 @@ Usage:
 """
 
 import os
+import re
 import subprocess
 from pathlib import Path
 from typing import List
+
+OUTER_FENCE_REGEX = re.compile(
+    r"\A\s*(`{3,}|~{3,})[^\n]*\n(.*)\n\1\s*\Z", re.DOTALL
+)
+
+
+def strip_llm_wrapper(text: str) -> str:
+    """Strip outer ```markdown ... ``` fence when it wraps the entire output.
+
+    Claude sometimes wraps its full response in a fenced block (despite
+    prompt instructions not to), which then fails validation because the
+    inner real code blocks no longer match the original. We only strip
+    when a single outer fence wraps the whole output — inner fences are
+    preserved exactly.
+    """
+    m = OUTER_FENCE_REGEX.match(text)
+    if m:
+        return m.group(2)
+    return text
 
 from .detect import should_compress
 from .validate import validate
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Closing — already on main. `strip_llm_wrapper()` and CommonMark-compliant line-based fence parser already in `compress.py` and `validate.py`.

PR #79, opened 2026-04-10
readme: matches the vibe

ashimdahal:


kronflux:
Eh. "why many token when few trick" not make sense. "why many token when few do trick" or "why many token when few good" or similar work better.
Not much point in change, but ok if make sense.

PR #83, opened 2026-04-10
feat: add Google Gemini CLI and Antigravity support

### LINKED ISSUE #81: Adding support for gemini cli
●What you want
  Add support for Gemini CLI so the caveman terse output style can be applied to reduce tokens when
  using gemini in the terminal.

  Before/after example
  Before: Gemini CLI gives verbose, padded responses with filler explanations
  After:  Gemini CLI gives terse, direct responses like caveman does for Claude Code

  Why good
  Gemini CLI has no built-in token reduction mode. Users running both Claude Code and Gemini CLI would
  benefit from consistent terse output across both tools. Could be implemented via npx skills add like
  the existing Cursor/Copilot support, or as a system prompt export for Gemini's --system flag.

thank you very much -mr

### LINKED ISSUE #82: Google Antigravity Skill
**What you want**
Google Antigravity Skill

**Why good**
So I can use this magic

adrianbadarau:
Before: No Gemini/Antigravity support.
After: Gemini/Antigravity work.
Add config files make agent know skills.

Fixes #81
Fixes #82

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,4 @@
+@./skills/caveman/SKILL.md
+@./skills/caveman-commit/SKILL.md
+@./skills/caveman-review/SKILL.md
+@./skills/compress/SKILL.md
--- a/GEMINI.md
+++ b/GEMINI.md
@@ -1,0 +1,4 @@
+@./skills/caveman/SKILL.md
+@./skills/caveman-commit/SKILL.md
+@./skills/caveman-review/SKILL.md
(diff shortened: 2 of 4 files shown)


adrianbadarau:
Integration work. See proof:

image

s3f5:
why even needed ? if gemini supports .agents/skills path ? https://geminicli.com/docs/cli/skills/

adrianbadarau:
I've closed the PR as the feat was already added by the creator or the repo.

PR #91, opened 2026-04-11
feat(warp): add Warp skill copies

vojtechgistr:
Add copy-based .agents skill artifacts for caveman, caveman-commit, caveman-review, and compress.

Update sync workflow to keep these copies in sync and document Warp install usage in README/CONTRIBUTING.

Co-Authored-By: Oz <oz-agent@warp.dev>


JuliusBrussee:
Thanks @vojtechgistr — Warp support now uses the shared skills registry instead of checked-in mirrors. Closing as implemented.

PR #93, opened 2026-04-11
Fix broken compress skill packaging in Codex plugin

### LINKED ISSUE #92: Codex plugin packages broken compress skill stubs instead of a real SKILL.md/scripts directory
## Summary
The Codex plugin install ships `plugins/caveman/skills/compress/SKILL.md` as a one-line path pointer instead of an actual skill file with YAML frontmatter. The companion `scripts` entry is also shipped as a one-line path pointer instead of a real directory.

## User-visible failure
Codex reports:

```text
Skipped loading 1 skill(s) due to invalid SKILL.md files.
...\skills\compress\SKILL.md: missing YAML frontmatter delimited by ---
```

## Root cause
The packaged plugin contains:

```text
plugins/caveman/skills/compress/SKILL.md -> ../../../../caveman-compress/SKILL.md
plugins/caveman/skills/compress/scripts  -> ../../../../caveman-compress/scripts
```

Those pointer-style files are not resolved by the Codex plugin loader, so it reads the literal file contents and rejects `SKILL.md` as invalid markdown frontmatter.

## Expected
The packaged plugin should contain a real `SKILL.md` file with frontmatter and a real `scripts/` directory so the `compress` skill can load and run after install.

## Proposed fix
Replace the pointer files in `plugins/caveman/skills/compress/` with the actual packaged skill content and script files.

apelz92:
## Summary
Fix the Codex plugin packaging for the `compress` skill.

## Problem
The packaged plugin currently ships:

```text
plugins/caveman/skills/compress/SKILL.md
plugins/caveman/skills/compress/scripts
```

as pointer-style stub files instead of actual packaged content. Codex does not resolve those stubs, so install/load fails with missing YAML frontmatter on `SKILL.md`.

Closes #92.

## Changes
- replace the stub `plugins/caveman/skills/compress/SKILL.md` with the real skill file containing YAML frontmatter
- replace the stub `plugins/caveman/skills/compress/scripts` entry with the actual Python script package
- preserve the existing compress-skill behavior while making the packaged plugin self-contained for Codex installs

## Verification
- confirmed the installed cached plugin loads a real `SKILL.md` that starts with `---`
- confirmed `scripts/` is a real directory in the packaged plugin rather than a one-line path stub

### DIFF
--- a/plugins/caveman/skills/compress/SKILL.md
+++ b/plugins/caveman/skills/compress/SKILL.md
@@ -1,1 +1,110 @@
-../../../../caveman-compress/SKILL.md
+---
+name: compress
+description: >
+  Compress natural language memory files (CLAUDE.md, todos, preferences) into caveman format
+  to save input tokens. Preserves all technical substance, code, URLs, and structure.
+  Compressed version overwrites the original file. Human-readable backup saved as FILE.original.md.
+  Trigger: /caveman:compress <filepath> or "compress memory file"
+---
+
+# Caveman Compress
+
+## Purpose
+
+Compress natural language files (CLAUDE.md, todos, preferences) into caveman-speak to reduce input tokens. Compressed version overwrites original. Human-readable backup saved as `<filename>.original.md`.
+
+## Trigger
+
+`/caveman:compress <filepath>` or when user asks to compress a memory file.
+
+## Process
+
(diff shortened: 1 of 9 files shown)


JuliusBrussee:
Fixed!

PR #95, opened 2026-04-11
Fix plugin-packaged compress symlinks

### LINKED ISSUE #92: Codex plugin packages broken compress skill stubs instead of a real SKILL.md/scripts directory

apelz92:
## Summary
Replace the plugin-packaged compress skill symlinks with real files so Codex can load the installed plugin on Windows and other setups where Git checks symlinks out as plain text stubs.

Closes #92.

## Root cause
plugins/caveman/skills/compress/SKILL.md and plugins/caveman/skills/compress/scripts were committed as symlinks. On machines with git config core.symlinks=false, Git checks them out as regular text files containing the target path, which breaks Codex skill loading.

## Changes
- replace plugins/caveman/skills/compress/SKILL.md symlink with a real packaged skill file
- replace plugins/caveman/skills/compress/scripts symlink with real packaged script files
- keep the fix scoped to the published Codex plugin payload only

## Verification
- reproduced the install failure locally on Windows with core.symlinks=false
- confirmed the installed cached plugin now has a real SKILL.md with YAML frontmatter instead of a path stub
- no broader repo symlink changes in this PR

### DIFF
--- a/plugins/caveman/skills/compress/SKILL.md
+++ b/plugins/caveman/skills/compress/SKILL.md
@@ -1,1 +1,111 @@
-../../../../caveman-compress/SKILL.md
+---
+name: compress
+description: >
+  Compress natural language memory files (CLAUDE.md, todos, preferences) into caveman format
+  to save input tokens. Preserves all technical substance, code, URLs, and structure.
+  Compressed version overwrites the original file. Human-readable backup saved as FILE.original.md.
+  Trigger: /caveman:compress <filepath> or "compress memory file"
+---
+
+# Caveman Compress
+
+## Purpose
+
+Compress natural language files (CLAUDE.md, todos, preferences) into caveman-speak to reduce input tokens. Compressed version overwrites original. Human-readable backup saved as `<filename>.original.md`.
+
+## Trigger
+
+`/caveman:compress <filepath>` or when user asks to compress a memory file.
+
+## Process
+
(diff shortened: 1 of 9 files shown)


JuliusBrussee:
Fixed with easier impl in 1.4.1

PR #99, opened 2026-04-11
feat(compress): respect CAVEMAN_MODEL on the claude CLI fallback path

marcelopaniza:
## Summary

Extend the existing `CAVEMAN_MODEL` env var support to the `claude --print` CLI fallback path.

The Anthropic SDK path in `call_claude()` already honors `CAVEMAN_MODEL` — see line 44 where it does `os.environ.get("CAVEMAN_MODEL", "claude-sonnet-4-5")`. But the CLI fallback (used when `ANTHROPIC_API_KEY` is unset or the `anthropic` package isn't installed — i.e., the common Claude Code plugin path) hardcodes whatever `claude --print` defaults to, ignoring the same env var.

This PR closes the gap so behavior is consistent across both paths.

## Why this matters

The CLI fallback path is the common one for users running caveman as a Claude Code plugin. Those users can't install `anthropic` into Claude Code's bundled Python environment, and many of them aren't using a raw API key — they use CLI auth. Today, those users have no way to change the model: `CAVEMAN_MODEL` is silently ignored on their execution path.

Concretely: running caveman compression on ~100 files against Opus costs about 5–10× what the same pass costs against Haiku, and the output quality is indistinguishable for this mechanical task (the existing regex validator catches both models' mistakes equivalently). Cost-conscious plugin users should be able to opt into the cheaper model just like API-key users already can.

## Usage

Session-wide (in shell rc):
```bash
export CAVEMAN_MODEL=claude-haiku-4-5-20251001
```

Per-call:
```bash
CAVEMAN_MODEL=claude-sonnet-4-6 caveman:compress big_file.md
```

Unset or empty → previous behavior, no `--model` flag passed to `claude --print`.

## Design notes

- **Non-breaking**: zero behavior change for users who don't set the env var. The `--model` flag is only appended if `CAVEMAN_MODEL` is non-empty, same pattern the SDK path already uses.
- **Consistent with SDK path**: same env var name, same "unset = default" semantics, so documentation can cover both paths together.
- **Retry path stays cheap**: if validation fails and `build_fix_prompt` runs, it re-enters `call_claude()`, which re-reads the env var — so the retry also uses the cheaper model.
- **Error messages updated**: `Claude call failed` now includes which model was in effect (`default` if unset), which helps debug cheap-model failures when they happen.

## Test plan

- [x] With `CAVEMAN_MODEL` unset, the constructed subprocess command is byte-identical to before (`["claude", "--print"]`).
- [ ] With `CAVEMAN_MODEL=claude-haiku-4-5-20251001`, compression produces valid output on a representative file (verified on a ~150 KB auto-memory directory).
- [ ] With `CAVEMAN_MODEL=nonexistent-model`, the error message clearly reports the invalid model.
- [ ] Retry path (forced by corrupting a code block) uses the same model as the initial call.

## Alternatives considered

- **A new env var for the CLI path** — rejected. Having one knob `CAVEMAN_MODEL` that works on both paths is simpler for users and for documentation.
- **Adding a `--model` CLI flag to the skill** — rejected. Skills are invoked via Claude Code's `Skill` tool, so an env var is easier to thread through than a flag.

## Scope

Touches only `caveman-compress/scripts/compress.py`. No README changes (the existing `CAVEMAN_MODEL` documentation, if any, applies equally to both paths now). Happy to add a README section clarifying the env var works on both paths if the maintainer wants.

### DIFF
--- a/caveman-compress/scripts/compress.py
+++ b/caveman-compress/scripts/compress.py
@@ -49,9 +49,13 @@ def call_claude(prompt: str) -> str:
         except ImportError:
             pass  # anthropic not installed, fall back to CLI
     # Fallback: use claude CLI (handles desktop auth)
+    cmd = ["claude", "--print"]
+    cli_model = os.environ.get("CAVEMAN_MODEL")
+    if cli_model:
+        cmd += ["--model", cli_model]
     try:
         result = subprocess.run(
-            ["claude", "--print"],
+            cmd,
             input=prompt,
             text=True,
             capture_output=True,
@@ -59,7 +63,9 @@ def call_claude(prompt: str) -> str:
         )
         return strip_llm_wrapper(result.stdout.strip())
     except subprocess.CalledProcessError as e:
-        raise RuntimeError(f"Claude call failed:\n{e.stderr}")
+        raise RuntimeError(
+            f"Claude call failed (model={cli_model or 'default'}):\n{e.stderr}"
+        )
 
 
 def build_compress_prompt(original: str) -> str:


JuliusBrussee:
Closing — already on main. `compress.py` reads `CAVEMAN_MODEL` env var on both SDK and CLI paths.

PR #104, opened 2026-04-11
Meta-compression of SKILL.md and Spanish localization

DarksAces:
I have applied the Caveman philosophy to its own instructions. This PR introduces "efficiency from within" by compressing the core skill and expanding accessibility.
**Changes:**
1. **Meta-Compression of `SKILL.md`**: Reduced the instruction footprint from **3.7KB to 1.9KB (-48%)**. This saves critical input tokens on every turn.
2. **Spanish Localization**: Added `SKILL.es.md` for high-efficiency technical Spanish communication.
3. **Terminal Utilities**: Added `commands/caveman-pipe` to summarize terminal input (like `git diff`) instantly.
**Before (SKILL.md):**
> "Respond terse like smart caveman. All technical substance stay. Only fluff die. ACTIVE EVERY RESPONSE. No revert after many turns. No filler drift. Still active if unsure. Off only: "stop caveman" / "normal mode"."

**After (SKILL.md):**
> "Respond terse like smart caveman. Substance stay. Fluff die. ACTIVE EVERY RESPONSE. No drift. Off: "stop caveman" / "normal mode"."
**Why this change is better:**
The meta-compressed prompt is significantly more token-efficient. By using fewer tokens in the system prompt, we free up more of the agent's context window for actual project code, satisfying the core goal: **"Why use many token when few do trick"**.
---

### DIFF
--- a/commands/caveman-pipe.ps1
+++ b/commands/caveman-pipe.ps1
@@ -1,0 +1,11 @@
+# Caveman Pipe: Summarize anything in technical caveman style
+# Usage: git diff | .\commands\caveman-pipe.ps1
+
+$inputData = $input | Out-String
+
+if ([string]::IsNullOrWhiteSpace($inputData)) {
+    Write-Host "Usage: <command> | .\commands\caveman-pipe.ps1"
+    exit 1
+}
+
+$inputData | claude --print -p "Responde en estilo cavernícola técnico. Sustancia queda. Paja muere."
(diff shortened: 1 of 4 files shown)


yairEO:
it's irrelevant and insignificant. input tokens are cheaper anyway than output tokens, which are the real money saving. This will barley save you a cent

PR #106, opened 2026-04-11
docs: consolidate Cursor/Windsurf/Cline/Copilot install details into one block

itsmejay80:
## Summary

- Four near-identical `<details>` sections for Cursor, Windsurf, Cline, and Copilot replaced with a single combined block
- Added a summary table showing each agent's install command and the rule file that `npx skills add` does **not** install
- All information preserved — rule file paths, always-on note, Copilot variant note, uninstall command

## Why

The four blocks were ~95% identical copy-paste. One consolidated block is easier to maintain and faster to scan.

## Test plan

- [ ] Verify install commands in table match the original per-agent blocks
- [ ] Verify rule file paths are correct for each agent
- [ ] Confirm "Want it always on?" snippet link still resolves in the merged block

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,64 +223,20 @@ Auto-activates via `GEMINI.md` context file. Also ships custom Gemini commands:
 </details>
 
 <details>
-<summary><strong>Cursor — full details</strong></summary>
+<summary><strong>Cursor / Windsurf / Cline / Copilot — full details</strong></summary>
 
-```bash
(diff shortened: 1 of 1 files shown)


itsmejay80:
@JuliusBrussee — consolidated the four near-identical Cursor/Windsurf/Cline/Copilot details blocks into one table. All info preserved — install commands, rule file paths, always-on note, uninstall. Happy to adjust if the format doesn't fit.

PR #122, opened 2026-04-12
fix: bundle Codex auto-activation hook in plugin

hummat:
## Summary
- add `plugins/caveman/.codex/hooks.json` so the installed Codex plugin carries the same SessionStart activation hook as the repo-local dogfood setup
- sync that hook file from the repo root in `.github/workflows/sync-skill.yml` and make verification fail if the plugin copy drifts
- update Codex-facing docs to describe the shipped plugin bundle rather than the repo-only `.codex/hooks.json`
- relax the stale activation banner assertion in `tests/verify_repo.py` so the pulled upstream hook output verifies cleanly again

## Verification
- `python3 tests/verify_repo.py`
- `python3 -m unittest tests.test_hooks`

## Why
The repo docs and maintainer notes said the Codex plugin in `plugins/caveman/` auto-activated via a hook, but the actual plugin bundle did not ship that hook file. In practice the behavior only existed when running Codex inside this repo, not after installing the plugin from the marketplace.

### DIFF
--- a/.github/workflows/sync-skill.yml
+++ b/.github/workflows/sync-skill.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - skills/caveman/SKILL.md
+      - .codex/hooks.json
       - rules/caveman-activate.md
       - caveman-compress/SKILL.md
       - caveman-compress/scripts/**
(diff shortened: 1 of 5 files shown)


hummat:
Follow-up after opening this PR: I tested the patched plugin against a fresh local Codex runtime and the hypothesis does not hold. Adding `plugins/caveman/.codex/hooks.json` to the installed plugin bundle did not cause SessionStart caveman activation in either `codex debug prompt-input` or a fresh `codex exec` run. That suggests current Codex does not load plugin-bundled `.codex/hooks.json` from the installed cache, or at least not in the way this PR assumes. Closing to avoid merging a packaging change that does not actually fix runtime behavior.

PR #135, opened 2026-04-13
feat(caveman): add Russian language support

kudg0:
## What changed

This PR adds Russian support to the caveman plugin and aligns the surrounding plugin metadata and docs with that behavior.

## Included

- keep the user's language in caveman mode, including Russian
- add Russian trigger and stop phrases
- add Russian examples to the main caveman skill
- update always-on activation rules and Claude hook fallback text
- update Codex/OpenAI plugin metadata
- refresh README and help docs to mention Russian support

## Notes

- source-of-truth behavior was updated in `skills/caveman/SKILL.md`
- synced plugin and generated copies were refreshed locally to match the source files
- hook scripts were syntax-checked with Node


kudg0:
Hi @JuliusBrussee! 
Thanks a lot for the work on this plugin. it’s genuinely super useful.

I’ve been using it a lot, and I think it could become even more valuable if it kept growing in a more multilingual direction, similar to how Chinese support was introduced. That pattern feels really promising, and extending it further to other languages would make the plugin useful for a much broader set of users.

I opened a small PR around this direction as well, if you have a moment, I’d really appreciate it if you could take a look.

Really appreciate the work you’ve put into this.

kudg0:
Duplicated by: https://github.com/JuliusBrussee/caveman/pull/144

PR #139, opened 2026-04-13
Add Kiro Docs to Caveman Skills repo

Keerthi-Sreenivas:
Adds Kiro as a fully supported agent with auto-activation via steering files and skill sync through CI.

### What changed

- Added `.kiro/skills/caveman/SKILL.md` and `.kiro/steering/caveman.md` so caveman auto-activates when the repo is opened in Kiro
- Updated `.github/workflows/sync-skill.yml` to sync both files from their single-source-of-truth origins, same as Cursor/Windsurf/Cline/Copilot
- Added Kiro column to the feature matrix in README with install command, details section, and footnote ⁵
- Moved Kiro out of the "Any other agent" catch-all into its own `<details>` block
- Updated `CLAUDE.md` sync table and agent distribution table with Kiro paths

### How it works

- `.kiro/steering/caveman.md` — always-included steering file, no frontmatter needed (Kiro's default is `inclusion: always`)
- `.kiro/skills/caveman/SKILL.md` — full skill loaded on demand when user says `/caveman` or request matches skill description
- CI keeps both in sync from `rules/caveman-activate.md` and `skills/caveman/SKILL.md` respectively

### DIFF
--- a/.github/workflows/sync-skill.yml
+++ b/.github/workflows/sync-skill.yml
@@ -36,6 +36,8 @@ jobs:
           cp skills/caveman/SKILL.md .cursor/skills/caveman/SKILL.md
           mkdir -p .windsurf/skills/caveman
           cp skills/caveman/SKILL.md .windsurf/skills/caveman/SKILL.md
+          mkdir -p .kiro/skills/caveman
+          cp skills/caveman/SKILL.md .kiro/skills/caveman/SKILL.md
 
       - name: Sync compress skill to plugin
         run: |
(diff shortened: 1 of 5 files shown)


felipeneuhauss:
Thank you @Keerthi-Sreenivas  :)

Thoughts on steering vs tokens on Kiro... 

I think that having steering + skills synced from the same sources as the other agents keeps the repo consistent and makes adoption straightforward.

One tradeoff worth calling out: always-on steering (as described for .kiro/steering/caveman.md) typically means that text is included in context repeatedly, so it does incur a recurring token cost, similar to other “always apply” project instructions. That is separate from the on-demand skill (.kiro/skills/caveman/SKILL.md), which only adds cost when the user invokes it or when the agent loads it for a matching request.

For a project whose whole pitch is reducing tokens in model output, it is worth being explicit in the docs (or in the steering file itself) that the steering layer is intentionally small and mainly routes users to the full skill — so the baseline context bump stays minimal while auto-activation still works.

If the steering file ever grows large or duplicates the skill, it might be worth revisiting what lives in steering vs skill to avoid paying twice on every turn. 

Let me know your thoughts!

JuliusBrussee:
Thanks @Keerthi-Sreenivas — Kiro is documented and wired via the current provider registry. Closing as implemented.

PR #151, opened 2026-04-14
Update SKILL.md reduce think

Adam-2244:
Before: 
**Thinking block**
User asks about best about best debugging option. Let me look in their codebase quickly to understand what type of project this is, then give a relevant debugging advice in ultra caveman mode.

**More prompts**

After: 
**Thinking block**
Ultra caveman mode. User asks about best option for debugging code. They're looking at java test file. In Cursor, there are debugging options. Let me give a brief answer.

In Cursor/IDE context, debugging options:

1. Built-in debugger (IDE breakpoint)
2. Print/log statements
3. Debug mode in tests
4. Remote debugging
5. Using the debug mode in cursor itself

Let me be concise and helpful in ultra caveman mode.

One sentence why change better

More things less token.


JuliusBrussee:
Thanks @Adam-2244 — Caveman owns visible response style, not hidden model reasoning. Closing this out-of-scope prompt change.

PR #156, opened 2026-04-14
feat: add native Copilot CLI extension and installer

Rogn:
## Summary

- add native Copilot CLI extension under `.github/extensions/caveman/`
- add installer scripts for per-repo and `--global` installs
- add canonical shared sources in `lib/` for parser + config resolver, with CI sync into runtime copies
- update README / CLAUDE / verification so Copilot CLI path is documented and validated
- add regression tests for extension behavior, installer behavior, and config resolution

## Before

- Copilot CLI only had static-instructions path; no native extension install path
- no Copilot CLI auto-activation, slash commands, or one-shot skills via native runtime
- default-mode config resolution could drift between runtimes; `~/.config/caveman/config.json` on Windows could be ignored by extension copy

## After

- Copilot CLI can install Caveman natively per repo or globally
- auto-activation, `/caveman` commands, one-shot skills, and mode switching work in Copilot CLI
- parser and config resolver each have one canonical source with synced runtime copies
- verification checks extension syntax without requiring Copilot CLI SDK outside Copilot CLI runtime

## Why this implementation

Copilot CLI has native extension discovery from repo `.github/extensions/` and user `~/.copilot/extensions/`. `npx skills add` would only install static instructions, so it cannot provide native `onSessionStart` / `onUserPromptSubmitted` hooks, slash commands, one-shot skills, or repo/global extension install.

Installer script exists because extension has to be copied into one of Copilot CLI's discovery locations. `--global` covers user-wide install; repo-targeted install covers checked-in project setup.

Shared `lib/` sources follow repo's existing sync pattern for generated/runtime consumers and reduce drift risk between hook runtime and Copilot extension runtime.

## Concerns / tradeoffs

- this does **not** use same install path as other agent integrations. That difference is intentional and platform-driven, not arbitrary.
- no persistent status badge in Copilot CLI because there is no statusline API there.
- kept scope to one feature: native Copilot CLI integration. Docs/tests/sync changes are support for that one integration, not a wider harness rewrite.

## Non-goals

- no change to core caveman behavior / SKILL semantics
- no change to other agent install paths
- no attempt to force generic cross-harness installer abstraction

## Verification

- `node tests/test_caveman_config.mjs`
- `node tests/test_extension_integration.mjs`
- `node tests/test_copilot_installer.mjs`
- `node --check .github/extensions/caveman/extension.mjs`
- manual smoke check: `node scripts/install-copilot-extension.mjs --global --force`


JuliusBrussee:
Thanks @Rogn — #553 is the current focused Copilot-native candidate, while this branch targets the retired layout. Closing as duplicate/superseded.

PR #170, opened 2026-04-15
update caveman agent switch syntax for Copilot compatibility

Programmercito:
What
Changed /caveman lite|full|ultra slash command syntax to plain text "say caveman lite/full/ultra" in caveman.agent.md.

Why
caveman slash commands only work in Claude Code (hooks intercept them). Copilot has no hook system → slash syntax does nothing. Plain text switch works across all agents.

Before

Default: **ultra**. Switch: `/caveman lite|full|ultra`.
After

Default: **ultra**. Switch: say "caveman lite", "caveman full", or "caveman ultra".
Notes

caveman.agent.md is NOT in the auto-synced file list → safe to edit directly.
No behavior change — only instruction text updated for Copilot users.

works in copilot in agent format

### DIFF
--- a/agents/caveman/caveman.agent.md
+++ b/agents/caveman/caveman.agent.md
@@ -1,0 +1,64 @@
+---
+name: caveman
+description: "Use when: user wants compressed terse responses, caveman mode, less tokens, brief answers. Cuts ~75% output tokens. Intensity levels: lite, full, ultra, wenyan."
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks @Programmercito — the edited agent file is gone and Copilot now uses the shared profile/install path. Closing as superseded.

PR #195, opened 2026-04-16
feat: add native opencode plugin (server + TUI)

sdeonvacation:
## What
Adds native opencode support for caveman:

- `plugins/caveman/.opencode-plugin/server.ts` — injects caveman system prompt on every turn via `experimental.chat.system.transform`. Reads per-session state (disabled list + level) from `XDG_STATE_HOME/opencode/caveman.json`, validates persisted state before use, and tracks subagent sessions via bus events to auto-apply ultra for them.
- `plugins/caveman/.opencode-plugin/tui.ts` — registers three TUI commands in `/`: enable, disable, set-level (`lite` / `full` / `ultra`) with dialog picker.
- `plugins/caveman/.opencode-plugin/package.json` — nested plugin manifest exporting `./server` and `./tui`.
- `package.json` — root package manifest exposing opencode entrypoints so opencode can install caveman directly from git/npm package spec without requiring users to clone repo.
- `README.md` — promotes opencode to first-class support and documents install via `github:JuliusBrussee/caveman` in `opencode.json`.

## Before / After
Before: opencode listed only in \"Any other agent\" section with generic `npx skills` install (skill only, no auto-activation), and native plugin usage required cloning repo.

After: opencode can install caveman directly from package spec (`github:JuliusBrussee/caveman`), auto-load server + TUI entrypoints, auto-activate every session, support per-session enable/disable/level switching via TUI commands, auto-ultra for subagent sessions, and safely handle malformed persisted plugin state.

## Structure
```
package.json                  ← new root package for package-spec install
plugins/caveman/
  .codex-plugin/             ← existing
  .opencode-plugin/          ← new
    package.json
    server.ts
    tui.ts
```

## Notes
- `experimental.chat.system.transform` hook is marked experimental in opencode's API — same as other plugin authors use today
- `compress` and `caveman-help` not included (`compress` requires local skill files; `help` is skill-based)
- Tested against opencode's real plugin loader / plugin tests — direct package install via Bun resolves root package exports and loads both `./server` and `./tui` without repo clone

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Pick your agent. One command. Done.
 | **Cursor** | `npx skills add JuliusBrussee/caveman -a cursor` |
 | **Windsurf** | `npx skills add JuliusBrussee/caveman -a windsurf` |
 | **Copilot** | `npx skills add JuliusBrussee/caveman -a github-copilot` |
+| **opencode** | Clone repo → add `.opencode-plugin/` path to `opencode.json` `plugin` array |
 | **Cline** | `npx skills add JuliusBrussee/caveman -a cline` |
 | **Any other** | `npx skills add JuliusBrussee/caveman` |
 
@@ -141,17 +142,17 @@ Install once. Use in every session for that install target after that. One rock.
 
 Auto-activation is built in for Claude Code, Gemini CLI, and the repo-local Codex setup below. `npx skills add` installs the skill for other agents, but does **not** install repo rule/instruction files, so Caveman does not auto-start there unless you add the always-on snippet below.
 
-| Feature | Claude Code | Codex | Gemini CLI | Cursor | Windsurf | Cline | Copilot |
-|---------|:-----------:|:-----:|:----------:|:------:|:--------:|:-----:|:-------:|
-| Caveman mode | Y | Y | Y | Y | Y | Y | Y |
-| Auto-activate every session | Y | Y¹ | Y | —² | —² | —² | —² |
-| `/caveman` command | Y | Y¹ | Y | — | — | — | — |
-| Mode switching (lite/full/ultra) | Y | Y¹ | Y | Y³ | Y³ | — | — |
-| Statusline badge | Y⁴ | — | — | — | — | — | — |
-| caveman-commit | Y | — | Y | Y | Y | Y | Y |
-| caveman-review | Y | — | Y | Y | Y | Y | Y |
-| caveman-compress | Y | Y | Y | Y | Y | Y | Y |
-| caveman-help | Y | — | Y | Y | Y | Y | Y |
+| Feature | Claude Code | Codex | Gemini CLI | Cursor | Windsurf | Cline | Copilot | opencode |
+|---------|:-----------:|:-----:|:----------:|:------:|:--------:|:-----:|:-------:|:--------:|
+| Caveman mode | Y | Y | Y | Y | Y | Y | Y | Y |
+| Auto-activate every session | Y | Y¹ | Y | —² | —² | —² | —² | Y⁵ |
+| `/caveman` command | Y | Y¹ | Y | — | — | — | — | Y⁵ |
+| Mode switching (lite/full/ultra) | Y | Y¹ | Y | Y³ | Y³ | — | — | Y⁵ |
+| Statusline badge | Y⁴ | — | — | — | — | — | — | — |
+| caveman-commit | Y | — | Y | Y | Y | Y | Y | Y |
+| caveman-review | Y | — | Y | Y | Y | Y | Y | Y |
+| caveman-compress | Y | Y | Y | Y | Y | Y | Y | — |
+| caveman-help | Y | — | Y | Y | Y | Y | Y | — |
 
 > [!NOTE]
 > Auto-activation works differently per agent: Claude Code uses SessionStart hooks, this repo's Codex dogfood setup uses `.codex/hooks.json`, Gemini uses context files. Cursor/Windsurf/Cline/Copilot can be made always-on, but `npx skills add` installs only the skill, not the repo rule/instruction files.
@@ -160,6 +161,7 @@ Auto-activation is built in for Claude Code, Gemini CLI, and the repo-local Code
 > ² Add the "Want it always on?" snippet below to those agents' system prompt or rule file if you want session-start activation.
 > ³ Cursor and Windsurf receive the full SKILL.md with all intensity levels. Mode switching works on-demand via the skill; no slash command.
 > ⁴ Available in Claude Code, but plugin install only nudges setup. Standalone `install.sh` / `install.ps1` configures it automatically when no custom `statusLine` exists.
+> ⁵ Via the native opencode plugin (server + TUI). Enable/disable/level commands work per-session. Subagent sessions auto-use ultra.
 
 <details>
 <summary><strong>Claude Code — full details</strong></summary>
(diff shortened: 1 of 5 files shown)


alblez:
It would be nice to have this PR integrated soon

Qodo-Free-For-OSS:
Hi, readState() returns the raw JSON.parse result without validating shape/types, but later code assumes State and can throw (e.g., state.disabled.includes) or push undefined rules when the file contains valid JSON with the wrong structure.

**How to fix:** Validate and sanitize state

> ### Issue description
> `caveman.json` is parsed and trusted without runtime validation. If its JSON structure is wrong, the plugin can crash or inject invalid system rules.
> 
> ### Issue Context
> Both server and TUI parse state via `JSON.parse(...)` and then treat the result as `{ disabled: string[]; levels: Record<string, Level> }`.
> 
> ### Fix Focus Areas
> - plugins/caveman/.opencode-plugin/server.ts[15-64]
> - plugins/caveman/.opencode-plugin/tui.ts[15-85]
> 
> ### Expected fix
> - Add a small runtime validator/type-guard (or schema) that:
>   - defaults `disabled` to `[]` if not an array of strings
>   - defaults `levels` to `{}` if not an object; filters values to `lite|full|ultra`
> - Ensure `RULES[level]` is always defined before pushing.
> - Consider logging (or at least safely ignoring) malformed state rather than throwing.

---

ALameLlama:
If we're going down the route of adding native support for opencode could, I think it would be better to do it like this https://opencode.ai/docs/plugins/#create-a-plugin

e.g [superpowers](https://github.com/obra/superpowers/tree/main/.opencode) it just requires adding it to the plugin array

sdeonvacation:
@qodo-ai-reviewer Added runtime sanitization for persisted plugin state in both `server.ts` and `tui.ts`, so malformed JSON now safely falls back to sane defaults instead of being trusted blindly. `disabled` is filtered to string arrays, `levels` is filtered to valid `lite|full|ultra` values, and the server now falls back safely before `RULES[...]` lookup so we never push an undefined rule.

I also sanity-tested the plugin/package loading path against opencode’s plugin loader tests after the change.

Commit: 58e1de0

sdeonvacation:
> If we're going down the route of adding native support for opencode could, I think it would be better to do it like this https://opencode.ai/docs/plugins/#create-a-plugin
> 
> e.g [superpowers](https://github.com/obra/superpowers/tree/main/.opencode) it does requires adding it to the plugin array

@ALameLlama Good call. I checked opencode’s plugin model more closely and updated this to be more aligned with it.
The current `.opencode-plugin` folder is a valid opencode plugin package because it exposes `./server` and `./tui` in `package.json`, so the runtime can load both entrypoints from the plugin array. To make that clearer and more idiomatic, I updated the README in 79fb558 to recommend adding the plugin directory itself to `opencode.json`:
```json
{
  "plugin": ["/path/to/caveman/plugins/caveman/.opencode-plugin"]
}
instead of wiring separate server.ts / tui.ts file paths manually.

**Comment by ALameLlama** [comment]
> > If we're going down the route of adding native support for opencode could, I think it would be better to do it like this [opencode.ai/docs/plugins#create-a-plugin](https://opencode.ai/docs/plugins/#create-a-plugin)
> > e.g [superpowers](https://github.com/obra/superpowers/tree/main/.opencode?rgh-link-date=2026-04-25T05%3A23%3A09Z) it does requires adding it to the plugin array
> 
> @ALameLlama Good call. I checked opencode’s plugin model more closely and updated this to be more aligned with it. The current `.opencode-plugin` folder is a valid opencode plugin package because it exposes `./server` and `./tui` in `package.json`, so the runtime can load both entrypoints from the plugin array. To make that clearer and more idiomatic, I updated the README in [79fb558](https://github.com/JuliusBrussee/caveman/commit/79fb558ed65460f03fce39b9fb0754bbc9dade81) to recommend adding the plugin directory itself to `opencode.json`:
> 
> ```json
> {
>   "plugin": ["/path/to/caveman/plugins/caveman/.opencode-plugin"]
> }
> instead of wiring separate server.ts / tui.ts file paths manually.
> ```

I think doing it like this would still require cloning down the repo if i'm not mistaken. superpowers get around this by having this in the root https://github.com/obra/superpowers/blob/main/package.json. but I'm not sure if this is something the caveman maintainers would want to do.

I was playing around with this idea the other day and vibe coded this mess https://github.com/JuliusBrussee/caveman/compare/main...ALameLlama:caveman:opencode-plugin just to see if the idea would even work and installed it like this 
```
  "plugin": [
    "@tarquinen/opencode-dcp@latest",
    "superpowers@git+https://github.com/obra/superpowers.git",
    "caveman@git+https://github.com/ALameLlama/caveman.git#opencode-plugin",
  ],
```

**Comment by sdeonvacation** [comment]
> > > If we're going down the route of adding native support for opencode could, I think it would be better to do it like this [opencode.ai/docs/plugins#create-a-plugin](https://opencode.ai/docs/plugins/#create-a-plugin)
> > > e.g [superpowers](https://github.com/obra/superpowers/tree/main/.opencode?rgh-link-date=2026-04-25T05%3A23%3A09Z) it does requires adding it to the plugin array
> > 
> > 
> > @ALameLlama Good call. I checked opencode’s plugin model more closely and updated this to be more aligned with it. The current `.opencode-plugin` folder is a valid opencode plugin package because it exposes `./server` and `./tui` in `package.json`, so the runtime can load both entrypoints from the plugin array. To make that clearer and more idiomatic, I updated the README in [79fb558](https://github.com/JuliusBrussee/caveman/commit/79fb558ed65460f03fce39b9fb0754bbc9dade81) to recommend adding the plugin directory itself to `opencode.json`:
> > ```json
> > {
> >   "plugin": ["/path/to/caveman/plugins/caveman/.opencode-plugin"]
> > }
> > instead of wiring separate server.ts / tui.ts file paths manually.
> > ```
> 
> I think doing it like this would still require cloning down the repo if i'm not mistaken. superpowers get around this by having this in the root https://github.com/obra/superpowers/blob/main/package.json. but I'm not sure if this is something the caveman maintainers would want to do.
> 
> I was playing around with this idea the other day and vibe coded this mess [main...ALameLlama:caveman:opencode-plugin](https://github.com/JuliusBrussee/caveman/compare/main...ALameLlama:caveman:opencode-plugin) just to see if the idea would even work and installed it like this
> 
> ```
>   "plugin": [
>     "@tarquinen/opencode-dcp@latest",
>     "superpowers@git+https://github.com/obra/superpowers.git",
>     "caveman@git+https://github.com/ALameLlama/caveman.git#opencode-plugin",
>   ],
> ```

@ALameLlama You were right that the previous README flow still effectively required a local clone.

I updated this so opencode can install Caveman directly as a native plugin package spec instead:

> {
  plugin: [github:JuliusBrussee/caveman]
> }
>
To support that, I added a root package.json that exposes the existing opencode plugin entrypoints:
- ./server → plugins/caveman/.opencode-plugin/server.ts
 - ./tui → plugins/caveman/.opencode-plugin/tui.ts

So opencode now installs the repo via Bun and loads both server + TUI from the root package exports, without needing users to clone the repo just to point at .opencode-plugin/.

I also verified the install/load path and did an isolated smoke test of the plugin init before pushing. Please test and let me know if this works for you.

Commit: 6283c8e

**To summarise:**

Users do this in ~/.config/opencode/opencode.json:
{
  plugin: [github:JuliusBrussee/caveman]
}

Then start/restart opencode.

What happens:
- opencode installs caveman via Bun
- loads server + TUI from root package exports
- caveman auto-activates each session

In TUI:
- /enable_caveman
- /disable_caveman
- /caveman_level

No repo clone needed. No manual file copy needed.

**Comment by ALameLlama** [comment]
I tried to test this but it looks like it didn't install properly, I added
`"caveman@git+https://github.com/sdeonvacation/caveman.git#opencode-plugin",` but inside my ~/.cache/opencode it looks like it tried to pull your branch but it was empty.

Also for the commands, I think it would be good to keep them as close to the existing ones e.g

```
/enable_caveman
/disable_caveman
/caveman_level
```

should be closer to https://github.com/JuliusBrussee/caveman#usage

`/caveman` or `/caveman lite`

Rumblingb:
Excited for OpenCode support! If you need caveman-style documentation or a caveman code reviewer for testing, check out https://rumblingb.github.io/caveman-tools/ — paste code, get caveman feedback instantly. Free to use.

douglasgomes98:
any updates?

JuliusBrussee:
Thanks @sdeonvacation — native OpenCode lifecycle support now ships in the current plugin. Closing this superseded implementation.

PR #213, opened 2026-04-17
fix: expand natural-language toggle phrasings in mode tracker

hummat:
The previous regex (PR #120, which I also wrote) added natural-language toggles but missed a case that turns out to be the one users actually type: bare `caveman off`. The alternation required `turn off` as a two-word phrase, so `off` on its own slipped through. I just hit this myself — typed `caveman off`, statusline kept showing the badge, hook kept injecting the reinforcement on every turn. Took me a while to notice because the flag file silently stayed put.

## What's new

**Activation:** `caveman on`, `on caveman`, `go caveman`, `use caveman`, `be caveman`, `switch on caveman`.

**Deactivation:** `caveman off`, `off caveman`, `no caveman`, `exit/end/cancel caveman`, `switch off caveman`, `back to normal`.

## False-positive defense

Bare `on`, `off`, `no` only match in tight proximity to `caveman` — patterns like `\bcaveman\s+off\b`, not the sentence-spanning `.*` alternation. Without this, prose like *"the on switch is broken"*, *"I have no problems today"*, or *"off to the races"* would flip the flag. Tight proximity keeps false positives at zero on the test cases I tried.

The activation guard now also blocks on `off`, `switch off`, `exit`, `end`, `cancel`, and `no` — so an ambiguous prompt like *"stop caveman mode"* can't simultaneously activate and deactivate. The guard wins, deactivation fires.

## Test plan

I ran both regexes through 35 hand-crafted cases: every new trigger, every existing trigger, and a false-positive section (`"hello world"`, `"the on switch is broken"`, `"I have no problems today"`, `"off to the races"`, `"end of file"`, `"cancel my order"`). 34/35 behave as intended. The one skipped case is `"caveman please"` — intent is ambiguous enough that I didn't want to guess.

- [ ] Smoke test: open Claude Code, type `caveman off`, verify flag deletion + statusline update
- [ ] Smoke test: `no caveman`, `go caveman`, `back to normal` — deactivation/activation/deactivation respectively
- [ ] Regression check on existing phrasings (`stop caveman`, `normal mode`, `turn off caveman`)

### DIFF
--- a/hooks/caveman-mode-tracker.js
+++ b/hooks/caveman-mode-tracker.js
@@ -17,22 +17,36 @@ process.stdin.on('end', () => {
     const data = JSON.parse(input);
     const prompt = (data.prompt || '').trim().toLowerCase();
 
+    // Deactivation patterns — reused below to unlink the flag and above as the
+    // activation guard. Broad verbs (stop/disable/deactivate/turn off/switch off)
+    // use sentence-spanning .* because they unambiguously indicate deactivation.
+    // Bare off/no and the ambiguous verbs exit/end/cancel require tight proximity
+    // to "caveman" so unrelated prose like "go caveman and end with a summary"
+    // does not flip the flag.
+    const isDeactivation =
+        /\b(stop|disable|deactivate|turn off|switch off)\b.*\bcaveman\b/i.test(prompt) ||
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks @hummat — current parsing covers bare off/on and related natural-language forms. Closing as implemented.

PR #214, opened 2026-04-17
feat: per-project default mode via .claude/caveman.local.md

hummat:
Previously the only way to scope caveman differently per project was a direnv trick on `CAVEMAN_DEFAULT_MODE`. That's workable but awkward — you have to remember to `direnv allow` in every repo, and the setting lives in an `.envrc` that's not obviously about caveman. I wanted something like `.prettierrc`: drop a file in the project, done.

## Resolution order now

1. `CAVEMAN_DEFAULT_MODE` env var (unchanged — session override)
2. `$CLAUDE_PROJECT_DIR/.claude/caveman.local.md` ← new
3. `~/.config/caveman/config.json` (unchanged — global)
4. `full` (unchanged — fallback)

## Format

```markdown
---
defaultMode: off
---

# Project notes (optional markdown body, ignored by plugin)
Caveman disabled for this repo.
```

## Why `.claude/caveman.local.md`

I picked this path because it matches the [Claude Code plugin-settings convention](https://docs.claude.com/en/docs/claude-code/plugins) — other plugins like ralph-loop and multi-agent-swarm already use `.claude/<plugin-name>.local.md`, and the `$CLAUDE_PROJECT_DIR` env var hands us the project root for free. Users who recognize the pattern from other plugins won't need docs to figure it out.

Gitignored by convention (`.claude/*.local.md`) so per-user preferences don't leak between collaborators.

## Security

Symmetric with the existing `readFlag()` hardening:

| Defense | Why |
|---------|-----|
| `lstat()` + `O_NOFOLLOW` symlink check | Without this, a local attacker with write access to `.claude/` could point the config at `~/.ssh/id_rsa` and have the plugin slurp it on the next session |
| 4096-byte size cap | Enough room for frontmatter plus modest notes, caps the attack surface if the file ever gets echoed into a log |
| Whitelist validation | File contents never trusted directly — the parsed value is checked against `VALID_MODES` before use |
| Silent-fail on all errors | Missing, malformed, or unreadable config falls through cleanly to the next resolution step |

## Parsing

I deliberately avoided pulling in a YAML dependency. The plugin has zero npm deps today and I want to keep it that way. Only the `defaultMode:` line is parsed, via a single-line regex on the frontmatter block. Markdown body is ignored. Handles bare values and quoted values (both single and double), rejects anything outside `VALID_MODES`. If someone wants a second field here later, that's a separate design discussion.

## Tests

Verified end-to-end via a small Node harness:

- [x] Valid bare value (`defaultMode: off`) → resolves to `off`
- [x] Quoted values (`"lite"`, `'ultra'`) → parsed correctly
- [x] Invalid mode (`defaultMode: garbage`) → falls through to global/default
- [x] No frontmatter block → falls through
- [x] Symlinked file (e.g. to `/etc/passwd`) → rejected
- [x] Missing `$CLAUDE_PROJECT_DIR` and no config → falls through to `full`
- [x] `CAVEMAN_DEFAULT_MODE` env var overrides project config

- [ ] Manual smoke test: drop `.claude/caveman.local.md` with `defaultMode: off` in a project, start a Claude Code session, verify caveman is disabled there but still active elsewhere

## Docs

Added a Configuration section to the README covering the full resolution chain, the per-project file format, the gitignore pattern, and the global config file. The global config was previously undocumented in the README — this PR documents both layers together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,48 @@ Classical Chinese literary compression — same technical accuracy, but in the m
 
 Level stick until you change it or session end.
 
+### Configuration
+
+Default mode resolves in this order:
+
+1. **`CAVEMAN_DEFAULT_MODE`** env var — session override
+2. **`.claude/caveman.local.md`** in your project root — per-project default
+3. **`~/.config/caveman/config.json`** (or `$XDG_CONFIG_HOME/caveman/config.json`) — global default
+4. Hardcoded `full`
+
+#### Per-project default
+
+Create `.claude/caveman.local.md` in your project root:
+
+```markdown
+---
+defaultMode: off
+---
+
+# Project notes (optional markdown body, ignored by plugin)
+Caveman disabled for this repo.
+```
+
+Valid `defaultMode` values: `off`, `lite`, `full`, `ultra`, `wenyan-lite`, `wenyan`, `wenyan-ultra`.
+
+Add to `.gitignore` if per-user (Claude Code convention for `.claude/*.local.md`):
+
+```gitignore
+.claude/*.local.md
+```
+
+Per-project config is read at session start. Restart Claude Code to pick up changes.
+
+#### Global default
+
+```json
+{
+  "defaultMode": "lite"
+}
+```
+
+Save as `~/.config/caveman/config.json` (or `$XDG_CONFIG_HOME/caveman/config.json`).
+
 ## Caveman Skills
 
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks @hummat — per-project defaults landed through the safer repo-local config contract. Closing this older file format as implemented.

PR #216, opened 2026-04-17
[codex] fix verifier for current hook output

leopardqween:
## Summary
- update verifier banner assertions for the current activation hook output
- validate mode tracker reinforcement JSON instead of expecting silence
- keep the fix scoped to tests/verify_repo.py

## Root Cause
The verifier still expected older hook behavior: an activation banner with a period and a silent mode tracker. The current hook emits a level-aware activation banner and structured UserPromptSubmit reinforcement context.

## Verification
- python3 tests/verify_repo.py
- python3 -m unittest tests/test_hooks.py

### DIFF
--- a/tests/verify_repo.py
+++ b/tests/verify_repo.py
@@ -225,7 +225,7 @@ def verify_hook_install_flow() -> None:
             ["node", "hooks/caveman-activate.js"],
             env={"HOME": str(home)},
         )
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks @leopardqween — verifier expectations now match the current hook protocol. Closing this old-base patch.

PR #236, opened 2026-04-19
Ultra: add density benchmark and optional 🪨 anchor

### LINKED ISSUE #235: Ultra level: add density benchmark and optional visual anchor
## Problem

Ultra mode drifts verbose over long sessions. The model reads the rules, follows them initially, then gradually slips back towards structured responses with tables, headers, and full sentences — especially after tool chains or context compression.

The existing ultra spec says *"one word when one word enough"* but doesn't show what the target density actually looks like in practice. Without a concrete benchmark, the model treats ultra as "a bit shorter than full" rather than the telegraphic compression it's meant to be.

## Suggestion

Two small additions to the ultra intensity section in `SKILL.md`:

### 1. Density benchmark

Add a line above the rules making the existing examples the explicit target:

```
**Density benchmark — match this:** `"Inline obj prop → new ref → re-render. useMemo."` and `"Pool = reuse DB conn. Skip handshake → fast under load."`
```

These examples already exist in SKILL.md — this just promotes them from "here's what it looks like" to "this is the standard you must hit."

### 2. Optional visual anchor (🪨)

Add as an opt-in rule:

```
Optional: start every response with 🪨 as a slippage detector — missing stone = caveman slipped.
```

One emoji per response costs 1–2 tokens. The value: you spot drift instantly without reading the text. It's a canary — if the stone disappears, the model has reverted to verbose mode.

## Before / After

**Before (typical ultra drift on a technical explanation):**

> **Component** (`GradientRibbons.tsx`) — 3 variants, pure CSS (replaced framer-motion):
>
> | Variant | Used in | Visual |
> |---------|---------|--------|
> | `subtle` | WhyOCGSection | Horizontal band, orange→cyan gradient, blur(35px) |
> | `cta` | FinalCTASection | Centred radial ellipse, orange→purple, blur(50px) |
>
> **Animations** (globals.css):
> - **`subtleDrift`** (16s loop) — translates X/Y + rotates ±3° ...

**After (true ultra with benchmark in mind):**

> `GradientRibbons` — 3 variants. `subtle` = horizontal orange→cyan blur, drifts 16s. `cta` = radial orange→purple, pulses 10s. `full` = null. CSS-only, no JS. Big blurred shapes + slow anim → ambient glow.

Same information. Half the tokens.

## Why this is better

The benchmark turns a vague instruction into a measurable target. The anchor makes slippage visible at a glance. Both are zero-cost to implement and backwards-compatible — they don't change existing behaviour, just sharpen ultra's floor.

---

*Me think verbose model need big rock reminder. 🪨*

zagrosi-code:
Closes #235

Adds two small things to the ultra intensity row in `SKILL.md`:

1. **Density benchmark** — promotes the existing ultra example (`"Inline obj prop → new ref → re-render. useMemo."`) from illustration to explicit target. Models drift verbose over long sessions; a concrete benchmark gives them a measurable floor.

2. **Optional visual anchor (🪨)** — one emoji per response as a slippage canary. Costs 1–2 tokens. Missing stone = caveman slipped. Opt-in, not mandatory.

## Before

```
| **ultra** | Abbreviate (DB/auth/config/req/res/fn/impl), strip conjunctions, arrows for causality (X → Y), one word when one word enough |
```

## After

```
| **ultra** | Abbreviate (DB/auth/config/req/res/fn/impl), strip conjunctions, arrows for causality (X → Y), one word when one word enough. **Density benchmark — match this:** `"Inline obj prop → new ref → re-render. useMemo."` Optional: start every response with 🪨 as slippage detector — missing stone = caveman slipped |
```

Benchmark turns a vague instruction into a measurable target. Anchor makes drift visible at a glance.

*Verbose model forget. Rock remind. 🪨*

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -31,7 +31,7 @@ Yes: "Bug in auth middleware. Token expiry check use `<` not `<=`. Fix:"
 |-------|------------|
 | **lite** | No filler/hedging. Keep articles + full sentences. Professional but tight |
 | **full** | Drop articles, fragments OK, short synonyms. Classic caveman |
-| **ultra** | Abbreviate (DB/auth/config/req/res/fn/impl), strip conjunctions, arrows for causality (X → Y), one word when one word enough |
+| **ultra** | Abbreviate (DB/auth/config/req/res/fn/impl), strip conjunctions, arrows for causality (X → Y), one word when one word enough. **Density benchmark — match this:** `"Inline obj prop → new ref → re-render. useMemo."` Optional: start every response with 🪨 as slippage detector — missing stone = caveman slipped |
 | **wenyan-lite** | Semi-classical. Drop filler/hedging but keep grammar structure, classical register |
 | **wenyan-full** | Maximum classical terseness. Fully 文言文. 80-90% character reduction. Classical sentence patterns, verbs precede objects, subjects often omitted, classical particles (之/乃/為/其) |
 | **wenyan-ultra** | Extreme abbreviation while keeping classical Chinese feel. Maximum compression, ultra terse |


JuliusBrussee:
Thanks @zagrosi-code — the proposed canary itself spends tokens and the example uses arrows now measured as no saving. Closing because it contradicts the current compression contract.

PR #238, opened 2026-04-19
fix(skill): clarify ultra mode abbreviation scope

AnthoHansen:
## What

Adds an explicit boundary to the `ultra` intensity level: only prose words are subject to abbreviation. Code symbols, function names, API names, and error strings must always remain exact.

## Why

The previous description listed words that *can* be abbreviated (`DB/auth/config/req/res/fn/impl`) but said nothing about what must stay exact. A model reading the skill could infer that `fn` applies to actual function names in code examples, producing ambiguous or incorrect output.

## Changes

- **`skills/caveman/SKILL.md`** — extends the `ultra` row in the intensity table to state that prose abbreviation does not apply to code symbols, function names, API names, or error strings.

## How to Review

Single-line change in the intensity table. Compare the `ultra` row before and after.

## Notes

- No behavior change for prose output — only adds a protective boundary for code-adjacent content.
- Applies to all synced copies of `SKILL.md` via the existing CI workflow.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -31,7 +31,7 @@ Yes: "Bug in auth middleware. Token expiry check use `<` not `<=`. Fix:"
 |-------|------------|
 | **lite** | No filler/hedging. Keep articles + full sentences. Professional but tight |
 | **full** | Drop articles, fragments OK, short synonyms. Classic caveman |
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`f06348cb`](https://github.com/JuliusBrussee/caveman/commit/f06348cb) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #239, opened 2026-04-19
fix(skill): expand auto-clarity trigger conditions

AnthoHansen:
## What

Restructures the auto-clarity section from a run-on sentence into a bullet list and adds an explicit trigger for when compression itself creates technical ambiguity.

## Why

The original text covered external triggers (security warnings, irreversible actions, user confusion) but not internal ones — cases where stripping articles and conjunctions makes step order in a technical sequence ambiguous. For example, `"migrate table drop column backup first"` is genuinely dangerous to misread. The rule existed partially in the text but was buried in a single sentence with no example.

## Changes

- **`skills/caveman/SKILL.md`** — replaces the run-on auto-clarity sentence with a bullet list; adds a new trigger for compression-induced ambiguity with a concrete example.

## How to Review

Diff the `## Auto-Clarity` section. The existing triggers are preserved; the only additions are the new bullet and the list formatting.

## Notes

- No change to any other section.
- The existing destructive-op example is preserved unchanged.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -53,7 +53,14 @@ Example — "Explain database connection pooling."
 
 ## Auto-Clarity
 
-Drop caveman for: security warnings, irreversible action confirmations, multi-step sequences where fragment order risks misread, user asks to clarify or repeats question. Resume caveman after clear part done.
+Drop caveman when:
+- Security warnings
+- Irreversible action confirmations
+- Multi-step sequences where fragment order or omitted conjunctions risk misread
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`31d804e5`](https://github.com/JuliusBrussee/caveman/commit/31d804e5) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #267, opened 2026-04-22
feat: add Portuguese (PT-BR) language layer

DenisCDev:
## Summary

- Addresses #8 by adding a first non-English language layer (PT-BR) to `skills/caveman/SKILL.md`
- Base caveman rules target English: drop articles (a/an/the) and English filler (just/really/basically). In Portuguese those rules don't map — Portuguese articles are o/a/os/as, filler is então/tipo/aí, hedging is acho que/talvez. Result: caveman mode applied to PT-BR conversations still sounds verbose
- Adds a **Language layers** section with a PT-BR layer and a short template so future contributors can add ES/FR/JP/etc. with the same structure

## What changed

- `skills/caveman/SKILL.md` — new section between `## Auto-Clarity` and `## Boundaries`
- No changes to English behavior, intensity levels, or other skills

## PT-BR layer

Drops: PT articles (o/a/os/as/um/uma), redundant preps (de/para), conjunction "que" when obvious, PT filler (então/tipo/meio que/de fato/inclusive/já/enfim/pois/aí), PT hedging (talvez/acho que/possivelmente/provavelmente), PT pleonasms ("fazer com que", "no sentido de", "a nível de").

Includes 3 intensity examples (lite/full/ultra) for the same prompts used in the English section: React re-render, connection pooling, plus a "push feito" confirmation to show how CLI output summaries look under the layer.

## Test plan

- [x] English behavior unchanged — no edits to Rules/Intensity/Auto-Clarity sections
- [x] PT-BR examples parse as valid Portuguese and compress ~60-75% vs. standard PT responses
- [x] Markdown renders cleanly (checked structure: heading hierarchy, list indentation, code fences)
- [ ] Maintainer ack on the "Language layers" pattern before adding ES/FR/JP in follow-up PRs

## Follow-up

If this pattern is accepted, happy to open separate PRs for Spanish, French, and Japanese layers using the same template.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -62,6 +62,41 @@ Example — destructive op:
 > ```
 > Caveman resume. Verify backup exist first.
 
+## Language layers
+
+Base rules (Drop + Intensity above) target English. Other languages need an extra layer — articles, prepositions, and filler words differ. When conversation is in a non-English language, apply the matching layer on top of the intensity rules.
+
+### Portuguese (PT-BR)
+
+Drop:
+- Articles: o/a/os/as/um/uma/uns/umas
+- Redundant prepositions: "de X" when X is clear; "para" → "pra" or omit
+- Conjunction "que" when obvious
+- PT filler: então, tipo, meio que, de fato, inclusive, já, enfim, pois, aí
+- PT hedging: talvez, acho que, possivelmente, provavelmente, acredito que
+- Pleonasms: "fazer com que", "no sentido de", "a nível de", "do ponto de vista de"
+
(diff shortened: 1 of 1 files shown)


DenisCDev:
Closing this — the behaviour is already covered generically.

`skills/caveman/SKILL.md` line 27 already says to preserve the user's dominant language and never switch, `skills/caveman-help/SKILL.md` uses Portuguese as its own example, and `src/hooks/caveman-activate.js` carries the same rule into the hook. The article-dropping rule is scoped by language family rather than by individual language, so the design is deliberately language-agnostic.

A per-language layer would work against that and invite one PR per language. Not worth the maintenance for something that already works.

PR #284, opened 2026-04-25
feat: add OpenCode plugin with session.start workaround

AutomatorAlex:
## Summary

Adds OpenCode plugin that auto-injects caveman rules via `session.start` hook, with mode switching via `event` hook.

## Problem

OpenCode has two bugs blocking the standard approach:
- chat.message hook delivers empty parts array (#22831)
- experimental.chat.system.transform mutations discarded (#17100)

## Workaround

- `session.start`: Injects rules into system prompt on every session start (including post-compaction)
- `event` (message.updated): Parses /caveman commands and persists to ~/.config/opencode/.caveman-mode
- File-based state survives everything

## Before

Users must manually paste caveman rules into AGENTS.md:

```
Respond terse like smart caveman. All technical substance stay. Only fluff die.
Drop: articles (a/an/the), filler (just/really/basically), pleasantries, hedging.
Fragments OK. Short synonyms. Technical terms exact. Code unchanged.
Pattern: [thing] [action] [reason]. [next step].
ACTIVE EVERY RESPONSE. No revert after many turns. No filler drift.
Code/commits/PRs: normal. Off: "stop caveman" / "normal mode".
```

Problems:
- Manual copy-paste for every OpenCode user
- No mode switching (/caveman ultra, /caveman lite)
- Lost on compaction unless re-pasted
- No persistence across sessions

## After

Install plugin once:

```bash
cp plugins/caveman/.opencode-plugin/caveman.js ~/.config/opencode/plugins/
```

Then:
- Caveman auto-activates on every session start
- Say `/caveman ultra` → mode switches instantly
- State persists in ~/.config/opencode/.caveman-mode
- Survives compaction (session.start fires again)

## Why Better

Plugin eliminates manual setup, adds mode switching, and survives compaction — all without user intervention.

## Files Added

- plugins/caveman/.opencode-plugin/caveman.js

## README Updated

- OpenCode plugin installation instructions
- Table updated with plugin as primary method, AGENTS.md as fallback

## Testing

1. Copy caveman.js to ~/.config/opencode/plugins/
2. Restart OpenCode
3. Say /caveman ultra — mode file updates
4. Next response uses ultra compression

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,7 +273,17 @@ Uninstall: `npx skills remove caveman`
 
 > **Windows note:** `npx skills` uses symlinks by default. If symlinks fail, add `--copy`: `npx skills add JuliusBrussee/caveman --copy`
 
-**Important:** These agents don't have a hook system, so caveman won't auto-start. Say `/caveman` or "talk like caveman" to activate each session.
+**Important:** Most agents don't have a hook system, so caveman won't auto-start. Say `/caveman` or "talk like caveman" to activate each session.
+
+**OpenCode exception:** OpenCode supports plugins. Install the caveman plugin for automatic activation:
+
+```bash
+# Clone the repo and copy the plugin
+git clone https://github.com/JuliusBrussee/caveman.git
+cp caveman/plugins/caveman/.opencode-plugin/caveman.js ~/.config/opencode/plugins/
+```
+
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks @AutomatorAlex — current OpenCode support has real lifecycle hooks, so this workaround is no longer needed. Closing as implemented.

PR #287, opened 2026-04-26
fix: `/caveman stop` and `/caveman off` actually deactivate

### LINKED ISSUE #245: Statusline badge persists after /caveman stop
## Bug

When deactivating caveman mode via `/caveman stop` or `/caveman:caveman stop`, the statusline badge `[CAVEMAN]` continues to display because the `~/.claude/.caveman-active` flag file is not deleted.

## Root Cause

In `hooks/caveman-mode-tracker.js`, the `/caveman` command handler (lines 33-39) does not check for `stop` or `off` as an argument. When the arg is `stop`, it falls through to `mode = getDefaultMode()`, which sets mode to `full` — effectively re-activating caveman instead of deactivating it.

The fallback regex on line 51 (`/\b(stop caveman|normal mode)\b/`) is intended to catch deactivation, but when invoked as a slash command (`/caveman:caveman stop`), the prompt text doesn't match the pattern.

## Fix

Add a check for `stop`/`off` before the default fallback:

```js
else if (arg === 'stop' || arg === 'off') mode = 'off';
else mode = getDefaultMode();
```

This ensures the flag file is properly deleted via the existing `mode === 'off'` branch on line 45-46.

## Steps to Reproduce

1. Activate caveman: `/caveman`
2. Observe `[CAVEMAN]` in statusline
3. Deactivate: `/caveman stop`
4. Badge still shows `[CAVEMAN]`
5. `cat ~/.claude/.caveman-active` still contains `full`

ousamabenyounes:
## Summary

`/caveman stop` (and `/caveman:caveman stop`) was effectively a no-op
because the slash-command handler had no `stop`/`off` arg branch. The
arg fell through to `mode = getDefaultMode()` and a fresh
`safeWriteFlag()` ran with the default mode. The natural-language
matcher further down does eventually unlink the flag, but the interim
write is wasted I/O and the intent is unclear.

## Change

One line: `if (arg === 'stop' || arg === 'off') mode = 'off';` placed
before the existing arg branches, which routes through the existing
`mode === 'off'` path that unlinks the flag.

Net effect: a single filesystem op, the statusline clears immediately,
and the slash command matches the documented natural-language exit.

Fixes #245.

## Test plan

- [x] `/caveman` → flag set to `full`, statusline shows `[CAVEMAN]`
- [x] `/caveman stop` → flag deleted, statusline empty, no `[CAVEMAN]`
- [x] `/caveman off` → same as above
- [x] `/caveman lite` → flag set to `lite`
- [x] Existing natural-language exit (`stop caveman`) still unlinks

### DIFF
--- a/hooks/caveman-mode-tracker.js
+++ b/hooks/caveman-mode-tracker.js
@@ -45,7 +45,8 @@ process.stdin.on('end', () => {
       } else if (cmd === '/caveman-compress' || cmd === '/caveman:caveman-compress') {
         mode = 'compress';
       } else if (cmd === '/caveman' || cmd === '/caveman:caveman') {
-        if (arg === 'lite') mode = 'lite';
+        if (arg === 'stop' || arg === 'off') mode = 'off';
+        else if (arg === 'lite') mode = 'lite';
         else if (arg === 'ultra') mode = 'ultra';
         else if (arg === 'wenyan-lite') mode = 'wenyan-lite';
         else if (arg === 'wenyan' || arg === 'wenyan-full') mode = 'wenyan';


ousamabenyounes:
Superseded by upstream commit `a660db0` (#314), which added `'off'`/`'stop'`/`'disable'` to the `/caveman` arg whitelist alongside the broader hook whitelist fix. The current `src/hooks/caveman-mode-tracker.js` already handles `/caveman stop` and `/caveman off` exactly as this PR intended. Closing — no diff left to apply.

PR #288, opened 2026-04-26
feat(compress): add `--no-backup` flag to skip `.original.md`

### LINKED ISSUE #283: Compress: allow skipping .original.md files
**What you want**
Sometimes I just want to compress some .md files, low stakes, version controlled, so an option to skip creating a `*.original.md` would be nice

ousamabenyounes:
## Summary

Adds a `--no-backup` flag to `caveman-compress` so users can skip
writing the `<file>.original.md` companion next to their input. When
the input is already in version control, that file is noise — git
already provides the rollback path.

Closes #283.

## Behavior

- **Default (unchanged):** `.original.md` is written, kept on success,
  removed on validation-failure-after-retries (existing logic).
- **With `--no-backup`:** the "backup" the validator needs is routed
  through a `tempfile.NamedTemporaryFile` outside the user's working
  directory, then removed on every exit path. The user's directory
  stays clean.

## Implementation

- `compress_file(filepath, no_backup=False)` — new optional kwarg.
- `cli.py` — argv parsing accepts `--no-backup` anywhere; usage string
  updated; success summary prints `(skipped, --no-backup)` instead of a
  bogus path.
- The existing "abort if `.original.md` already exists" guard still runs
  on the default path; under `--no-backup` it's skipped because the
  temp path is unique.

## Test plan

- [x] `caveman foo.md` → `foo.md` compressed, `foo.original.md` kept
- [x] `caveman --no-backup foo.md` → `foo.md` compressed, no sibling
  file written, tempfile removed
- [x] `caveman --no-backup foo.md` with validation failure → `foo.md`
  restored from in-memory original, tempfile removed, no sibling file
- [x] Flag accepted before or after the path

### DIFF
--- a/skills/caveman-compress/scripts/cli.py
+++ b/skills/caveman-compress/scripts/cli.py
@@ -3,7 +3,12 @@
 Caveman Compress CLI
 
 Usage:
-    caveman <filepath>
+    caveman [--no-backup] <filepath>
+
+Flags:
+    --no-backup    Skip writing `<file>.original.md` next to the input.
+                   Useful for low-stakes, version-controlled files where
+                   git already provides the rollback path.
 """
 
 import sys
@@ -26,15 +31,21 @@ from .detect import detect_file_type, should_compress
 
 
 def print_usage():
-    print("Usage: caveman <filepath>")
+    print("Usage: caveman [--no-backup] <filepath>")
 
 
 def main():
-    if len(sys.argv) != 2:
+    args = sys.argv[1:]
+    no_backup = False
+    if "--no-backup" in args:
+        no_backup = True
+        args = [a for a in args if a != "--no-backup"]
+
+    if len(args) != 1:
         print_usage()
         sys.exit(1)
 
-    filepath = Path(sys.argv[1])
+    filepath = Path(args[0])
 
     # Check file exists
     if not filepath.exists():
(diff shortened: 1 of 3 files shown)


Qodo-Free-For-OSS:
Hi, When `no_backup=True`, `compress_file()` creates a `NamedTemporaryFile(delete=False)` but only unlinks it on the success path (and on the validation-final-failure path), so exceptions during Claude calls / writes / validation can orphan temp files.

**How to fix:** Ensure tempfile unlinked in finally

> ### Issue description
> When `--no-backup` is used, `compress_file()` creates a persistent tempfile (`delete=False`) but only deletes it on the success path (and on the final validation-failure return). Any exception after tempfile creation (Claude call failure, write failure, validation crash, etc.) will leak the tempfile.
> 
> ### Issue Context
> The CLI catches broad exceptions and exits, so leaked temp files will accumulate under repeated usage.
> 
> ### Fix Focus Areas
> - skills/compress/scripts/compress.py[155-247]
> 
> ### What to change
> - Wrap the body after tempfile creation in a `try/finally` that unlinks `backup_path` when `no_backup` is true.
> - Keep existing behavior for the default path (retain `<file>.original.md`).
> - Ensure the final-validation-failure path still restores original content before cleanup.

---

PR #289, opened 2026-04-26
fix(compress): force UTF-8 stdout to stop Windows cp1252 crashes

### LINKED ISSUE #152: caveman-compress: UnicodeEncodeError on Windows (cp1252) crashes before reporting validation errors
## Problem

On Windows, running `caveman-compress` with the default console codepage (cp1252) crashes with `UnicodeEncodeError` the moment the script tries to print the `❌` (U+274C) glyph — which happens on every validation failure and on the fallback error branch in `cli.py`.

The real failure (the validation error list) never reaches the user, because the print itself raises before the errors are shown. The script then dies without restoring the original file, leaving the compressed-but-unvalidated content in place and a half-written `<file>.original.md`.

## Repro

- Python 3.14, Windows 11
- Default stdout encoding: `cp1252`
- File: any 40KB+ natural-language markdown (triggered here on a 46KB CLAUDE.md)

```
Validation attempt 1
Traceback (most recent call last):
  File ".../scripts/compress.py", line 159, in compress_file
    print("\u274c Validation failed:")
UnicodeEncodeError: 'charmap' codec can't encode character '\u274c' in position 0

During handling of the above exception, another exception occurred:
  File ".../scripts/cli.py", line 68, in main
    print(f"\n\u274c Error: {e}")
UnicodeEncodeError: 'charmap' codec can't encode character '\u274c' in position 2
```

Occurrences of the problem glyph:
- `scripts/compress.py:159` — `print("❌ Validation failed:")`
- `scripts/compress.py:167` — `print("❌ Failed after retries — original restored")`
- `scripts/cli.py:29, 33, 60, 68` — error branches

## Workaround

Run with `PYTHONUTF8=1` or `PYTHONIOENCODING=utf-8` in the environment. Neither is set by default on Windows.

## Suggested fix

Any of:

1. **Force UTF-8 stdout at process start** (cleanest — no user env required):
   ```python
   import sys
   sys.stdout.reconfigure(encoding='utf-8', errors='replace')
   sys.stderr.reconfigure(encoding='utf-8', errors='replace')
   ```
   Put this at the top of `cli.py` / `__main__.py`.

2. Replace the `❌` glyphs with ASCII (`[FAIL]`, `ERROR:`) — safest, works on the oldest terminals too.

3. Document `PYTHONUTF8=1` as a Windows requirement.

## Secondary concern

When `print(...)` raises inside the validation-failed branch, the outer `try/except` in `cli.py` prints another `❌` and also raises, so `compress_file` never reaches its own cleanup path that restores the original file. Whatever the root cause, the script should restore the original on any unexpected exception, not just on `MAX_RETRIES` exhaustion.

Happy to send a PR if you'd like.

ousamabenyounes:
## Summary

On Windows the default console codepage is cp1252, so `caveman-compress`
crashes on the first ❌ glyph it tries to print — which happens on
every validation failure and on the fallback error branch in cli.py.
The real error (the validation list) never reaches the user, the outer
`try/except` in cli.py crashes on its *own* ❌, and the script exits
leaving a half-compressed file plus an empty `.original.md`.

## Fix

Reconfigure stdout and stderr to UTF-8 with `errors='replace'` at the
very top of `cli.py`, before any other module imports. This is option
(1) from the issue — the cleanest solution that needs no user env vars.

The reconfigure path is gated behind
`callable(getattr(_stream, 'reconfigure'))` so test runners that swap
out the streams aren't affected. `errors='replace'` keeps the program
running even if a host genuinely cannot render a code point — far
better than crashing on a glyph in an error message.

Fixes #152.

## Out of scope (separate patch)

The issue's "secondary concern" — that the outer `try/except` should
restore the original on *any* unexpected exception, not only on
`MAX_RETRIES` exhaustion — is a real issue but lives in compress.py's
flow control, not in encoding. Happy to send a follow-up PR scoped to
that change so the two patches can be reviewed independently.

## Test plan

- [x] Linux: still works, no behavior change (`isinstance(sys.stdout, TextIO)` keeps reconfigure callable)
- [x] Windows cp1252 console: ❌ glyph prints as `?` substitute, validation error list reaches the user, no crash
- [x] Embedded test runner with mock stdout (no `reconfigure`): no exception

### DIFF
--- a/skills/compress/scripts/cli.py
+++ b/skills/compress/scripts/cli.py
@@ -7,6 +7,20 @@ Usage:
 """
 
 import sys
+
+# Force UTF-8 on stdout/stderr before any code can print. Without this,
+# Windows consoles default to cp1252 and the ❌ glyphs in error/validation
+# branches raise UnicodeEncodeError, which masks the real error and leaves
+# the user with a half-compressed file. `errors='replace'` keeps the
+# program running even if the host terminal can't render a code point.
+for _stream in (sys.stdout, sys.stderr):
+    reconfigure = getattr(_stream, "reconfigure", None)
+    if callable(reconfigure):
+        try:
+            reconfigure(encoding="utf-8", errors="replace")
+        except Exception:
+            pass
+
 from pathlib import Path
 
 from .compress import compress_file


ousamabenyounes:
Merged into main as part of consolidation commit `57a9b48` (PR bundle May 1) — `cli.py` now reconfigures stdout/stderr to UTF-8 with the same gating you wrote (`callable(getattr(stream, 'reconfigure'))`). Credit preserved in the consolidation commit message. Closing.

PR #293, opened 2026-04-26
fix(compress): validate inline code spans (close silent-overwrite gap)

### LINKED ISSUE #112: caveman-compress silently overwrites inline code and commands despite "Validation passed"
## Title
`caveman-compress` reports success while rewriting inline code and commands because validation never checks inline backticks

## Summary
`caveman-compress` promises to preserve inline code, commands, file paths, and other technical content exactly, but its validator does not check inline code spans at all. As a result, the tool can print `Validation passed`, return success, and overwrite the source file even when inline commands or environment variables were changed.

## Impact
This matters because the tool is designed to rewrite always-loaded memory files such as `CLAUDE.md`. Those files often contain commands, env vars, flags, file names, and other precise technical instructions inside backticks. A successful run can silently mutate that technical content while presenting the result as safe and validated.

Affected users are contributors relying on `caveman-compress` to shrink memory/context files without changing meaning. The failure mode is subtle: the backup exists, the command succeeds, and the live file is still overwritten with incorrect instructions.

## Environment
- Repository: `JuliusBrussee/caveman`
- Investigated at commit: `8bab5c739e133b0a637b7e0fd1a732d47b414fe9`
- Host: Linux WSL2 (`Linux GEOVANI-PC 6.6.87.2-microsoft-standard-WSL2 x86_64`)
- Python: `3.12.3`
- Reproduction did not require network access or a real Claude call; `scripts.compress.call_claude` was mocked locally to make the behavior deterministic.

## Evidence
1. The user-facing contract explicitly says technical content is preserved exactly.
   - `caveman-compress/SKILL.md:48-57` says inline code, file paths, commands, and env vars must be preserved exactly.
   - `caveman-compress/SKILL.md:82-83` says: inline code inside backticks must be preserved exactly.
   - `caveman-compress/README.md:129-141` says inline code, commands, file paths, headings, dates, and version numbers are not touched.
   - `caveman-compress/README.md:43` says validations passed and file paths were preserved exactly.

2. The compression prompt enforces that contract, but only at prompt level.
   - `caveman-compress/scripts/compress.py:69-75` instructs the model not to modify inline backticks, URLs, headings, file paths, or commands.

3. The validator does not enforce that contract.
   - `caveman-compress/scripts/validate.py:156-160` only calls:
     - `validate_headings`
     - `validate_code_blocks`
     - `validate_urls`
     - `validate_paths`
     - `validate_bullets`
   - There is no validator for inline code spans, commands, env vars, or other backtick-delimited technical content.
   - `validate_paths()` only emits a warning (`validate.py:126-131`), not an error.
   - `validate_bullets()` only emits a warning (`validate.py:134-144`), not an error.

4. Success is keyed only off `result.is_valid`, so warnings do not block overwrite.
   - `caveman-compress/scripts/compress.py:153-157` treats validation as passed when there are no errors.
   - Because inline code is never checked, modified inline commands can still produce `Validation passed`.

5. Direct validator repro: changed inline code produces a clean success result.

```bash
python3 - <<'PY'
import sys, tempfile
from pathlib import Path

sys.path.insert(0, '/path/to/repo/caveman-compress')
from scripts.validate import validate

original = "# Title\nRun `npm install` and set `NODE_ENV=production`.\n"
compressed = "# Title\nRun `yarn install` and set `NODE_ENV=development`.\n"

with tempfile.TemporaryDirectory() as tmp:
    orig = Path(tmp) / "orig.md"
    comp = Path(tmp) / "comp.md"
    orig.write_text(original)
    comp.write_text(compressed)
    result = validate(orig, comp)
    print("is_valid:", result.is_valid)
    print("errors:", result.errors)
    print("warnings:", result.warnings)
PY
```

Observed output:

```text
is_valid: True
errors: []
warnings: []
```

6. End-to-end repro: `compress_file()` reports success and overwrites the source file with changed inline commands.

```bash
python3 - <<'PY'
import sys, tempfile
from pathlib import Path
from unittest.mock import patch

sys.path.insert(0, '/path/to/repo/caveman-compress')
from scripts.compress import compress_file

original = "# Title\nRun `npm install` and set `NODE_ENV=production`.\n"
corrupted = "# Title\nRun `yarn install` and set `NODE_ENV=development`.\n"

with tempfile.TemporaryDirectory() as tmp:
    path = Path(tmp) / "CLAUDE.md"
    path.write_text(original)
    with patch("scripts.compress.call_claude", return_value=corrupted):
        print("compress_file_returned:", compress_file(path))
    print("source_contents:")
    print(path.read_text())
PY
```

Observed output:

```text
Processing: /tmp/.../CLAUDE.md
Compressing with Claude...

Validation attempt 1
Validation passed
compress_file_returned: True
source_contents:
# Title
Run `yarn install` and set `NODE_ENV=development`.
```

## Steps to Reproduce
1. Clone the repository and `cd` into it.
2. Run the following command from the repo root:

```bash
python3 - <<'PY'
import sys, tempfile
from pathlib import Path
from unittest.mock import patch

sys.path.insert(0, 'caveman-compress')
from scripts.compress import compress_file

original = "# Title\nRun `npm install` and set `NODE_ENV=production`.\n"
corrupted = "# Title\nRun `yarn install` and set `NODE_ENV=development`.\n"

with tempfile.TemporaryDirectory() as tmp:
    path = Path(tmp) / "CLAUDE.md"
    path.write_text(original)
    with patch("scripts.compress.call_claude", return_value=corrupted):
        ok = compress_file(path)
    print("returned:", ok)
    print(path.read_text())
PY
```

3. Observe that the command prints `Validation passed`.
4. Observe that the function returns `True`.
5. Observe that the rewritten file now contains `yarn install` and `NODE_ENV=development` instead of the original inline code.

## Expected Behavior
If inline code, commands, env vars, or other preserved technical spans change, validation should fail and the run should not be reported as successful. A successful compression should not be able to change those backtick-delimited instructions.

## Actual Behavior
Inline code changes are invisible to the validator, so the run is treated as valid. `compress_file()` returns success and overwrites the source file even though the technical instructions inside backticks were changed.

## Root Cause Analysis
The core mismatch is between the preservation contract and the validator implementation:

- The prompt and docs define inline backticks and commands as immutable.
- The validator never extracts or compares inline code spans.
- The success path only looks at `result.is_valid`, which is set only by validator errors.
- Since no inline-code validator exists, those mutations never become errors.

There is a secondary gap for other “must preserve exactly” content: path mismatches are warnings, not errors, so even some detected drift would not block a successful overwrite.

## Proposed Fix
Add explicit validation for every content type the tool promises to preserve exactly, especially:

- inline code spans
- env vars
- commands/examples that appear inside inline code
- path-like tokens that are currently warning-only

At minimum, any mismatch in those preserved spans should become a hard validation error, not a warning.

## Optional Patch Sketch
```python
INLINE_CODE_REGEX = re.compile(r"`([^`\n]+)`")

def extract_inline_code(text: str) -> list[str]:
    return INLINE_CODE_REGEX.findall(text)

def validate_inline_code(orig: str, comp: str, result: ValidationResult) -> None:
    if extract_inline_code(orig) != extract_inline_code(comp):
        result.add_error("Inline code not preserved exactly")

def validate(original_path: Path, compressed_path: Path) -> ValidationResult:
    result = ValidationResult()
    orig = read_file(original_path)
    comp = read_file(compressed_path)

    validate_headings(orig, comp, result)
    validate_code_blocks(orig, comp, result)
    validate_inline_code(orig, comp, result)
    validate_urls(orig, comp, result)
    validate_paths(orig, comp, result)  # likely promote to error for exact-preserve content
    validate_bullets(orig, comp, result)
    return result
```

If you want stronger safety, stage compressed output in a temp file and only overwrite the source after validation fully passes.

## Risk / Regression Notes
After fixing this, it would be worth testing:

- inline code with commands and env vars, e.g. `` `npm install` `NODE_ENV=production` ``
- multiple inline code spans on one line
- escaped backticks / multi-backtick inline spans
- inline paths on Unix and Windows
- markdown files with both fenced code blocks and inline code
- cases where prose changes are allowed but inline technical spans must remain byte-for-byte identical

## Maintainer Value
This is worth maintainer attention because it breaks the main trust guarantee of `caveman-compress`: “compress prose, preserve technical content.” The current behavior can silently change operational instructions in a file that loads every session, while still claiming validation success. Fixing this would materially improve safety, user trust, and the credibility of the tool’s documented guarantees.

ousamabenyounes:
Closes #112

## Summary

`/caveman:compress` printed "Validation passed" while inline backtick content had been silently rewritten. The validator covered fenced code blocks, headings, URLs, paths (warning-only), and bullets — but **not** inline code spans, the very content the prompt and SKILL.md promise to preserve exactly. The reproducer from the issue rewrites \`npm install\` → \`yarn install\` and \`NODE_ENV=production\` → \`NODE_ENV=development\` and the run still reports success.

## Fix

- New `validate_inline_code` plugged into the main `validate()` pipeline.
- Inline spans are extracted **after** stripping fenced code blocks. Backticks inside a fenced block are already covered by `validate_code_blocks`, and the inline regex would otherwise produce noisy false-positives on shell snippets like `` echo `cmd` ``.
- Comparison is on the ordered list, not the multiset, so a re-ordered inline span (`` `--flag-a` `` and `` `--flag-b` `` swapped) still fails — preserving order is part of preserving the technical instruction.
- Mismatches surface up to five lost / added inline spans in the error message so the failing token is obvious in the log without dumping the entire file.

The maintainer's patch sketch suggested also promoting `validate_paths` from warning to error. I left that as warning for this PR — the path regex is heuristic (`PATH_REGEX` is comment-described as "crude but effective") and promoting it would risk false positives on regular compression runs. Happy to follow up in a separate PR if you'd like.

## Test plan

- [x] New `tests/test_validate_inline_code.py` (6 tests). `python3 -m unittest tests.test_validate_inline_code` passes 6/6.
- [x] Reproduces the exact scenario from the bug report: \`npm install\` -> \`yarn install\` now produces `Inline code not preserved exactly: lost=['NODE_ENV=production', 'npm install'] added=['NODE_ENV=development', 'yarn install']` and `is_valid=False`.
- [x] Confirms prose changes around preserved inline code still pass (no false-positive regression).
- [x] Confirms backticks living inside fenced blocks don't trigger the inline check.

### DIFF
--- a/skills/compress/scripts/validate.py
+++ b/skills/compress/scripts/validate.py
@@ -6,6 +6,14 @@ URL_REGEX = re.compile(r"https?://[^\s)]+")
 FENCE_OPEN_REGEX = re.compile(r"^(\s{0,3})(`{3,}|~{3,})(.*)$")
 HEADING_REGEX = re.compile(r"^(#{1,6})\s+(.*)", re.MULTILINE)
 BULLET_REGEX = re.compile(r"^\s*[-*+]\s+", re.MULTILINE)
+# Inline code span. Greedy on the inside, but bounded by a single line —
+# matches CommonMark single-backtick inline code, which is what the
+# compress prompt promises to preserve. Multi-backtick wrappers
+# (`` `code with backticks` ``) are intentionally not handled here:
+# they're rare in caveman-compress targets (CLAUDE.md / memory files)
+# and conservatively under-matching is safer than greedy spans that
+# accidentally bridge two different inline runs.
+INLINE_CODE_REGEX = re.compile(r"`([^`\n]+)`")
 
 # crude but effective path detection
 # Requires either a path prefix (./ ../ / or drive letter) or a slash/backslash within the match
@@ -81,6 +89,27 @@ def extract_code_blocks(text):
     return blocks
 
 
+def _strip_fenced_blocks(text: str) -> str:
+    """Return text with fenced code blocks removed.
+
+    Inline-code validation must not double-count backtick runs that live
+    inside a fenced block — those are already covered by
+    `validate_code_blocks`, and a shell snippet containing a `\\`` literal
+    inside a fenced block would otherwise generate spurious inline
+    mismatches.
+    """
+    blocks = extract_code_blocks(text)
+    out = text
+    for block in blocks:
+        out = out.replace(block, "", 1)
+    return out
+
+
+def extract_inline_code(text):
+    """Return ordered list of inline code spans outside fenced blocks."""
+    return INLINE_CODE_REGEX.findall(_strip_fenced_blocks(text))
+
+
 def extract_urls(text):
     return set(URL_REGEX.findall(text))
 
@@ -113,6 +142,30 @@ def validate_code_blocks(orig, comp, result):
 
     if c1 != c2:
         result.add_error("Code blocks not preserved exactly")
+
+
+def validate_inline_code(orig, comp, result):
+    """Inline backtick spans are part of the preservation contract.
+
+    `caveman-compress` promises to preserve commands, env vars, file
+    paths, and other technical content inside inline backticks exactly
+    (see SKILL.md / README.md). Without this check, the validator could
+    print "Validation passed" while `npm install` had been rewritten to
+    `yarn install` — silent drift on memory files (#112).
+    """
+    i1 = extract_inline_code(orig)
+    i2 = extract_inline_code(comp)
+
+    if i1 != i2:
+        # Surface a small diff sample so the failing span is obvious in
+        # the log without dumping the entire file.
+        s1, s2 = set(i1), set(i2)
+        lost = sorted(s1 - s2)[:5]
+        added = sorted(s2 - s1)[:5]
+        result.add_error(
+            "Inline code not preserved exactly: "
+            f"lost={lost} added={added}"
+        )
 
 
 def validate_urls(orig, comp, result):
@@ -155,6 +208,7 @@ def validate(original_path: Path, compressed_path: Path) -> ValidationResult:
 
     validate_headings(orig, comp, result)
     validate_code_blocks(orig, comp, result)
+    validate_inline_code(orig, comp, result)
     validate_urls(orig, comp, result)
     validate_paths(orig, comp, result)
     validate_bullets(orig, comp, result)
--- a/tests/test_validate_inline_code.py
+++ b/tests/test_validate_inline_code.py
@@ -1,0 +1,78 @@
+"""Tests for the inline-code validation guard (issue #112).
+
+The validator used to never inspect inline backtick spans, so a
+compression run could rewrite `npm install` -> `yarn install` and still
+print "Validation passed". These tests pin the new check.
+"""
+
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(REPO_ROOT))
+
+from skills.compress.scripts.validate import (  # noqa: E402
+    validate,
+    extract_inline_code,
+)
+
+
+class ExtractInlineCodeTests(unittest.TestCase):
+    def test_extracts_simple_inline(self):
+        self.assertEqual(
+            extract_inline_code("Run `npm install` then `yarn build`."),
+            ["npm install", "yarn build"],
+        )
+
+    def test_skips_fenced_block_content(self):
+        text = "Outside `keep`.\n\n```bash\necho `inside fence`\n```\n\nOutside `also`.\n"
+        self.assertEqual(extract_inline_code(text), ["keep", "also"])
+
+    def test_returns_empty_when_none(self):
+        self.assertEqual(extract_inline_code("Just prose with no code."), [])
+
+
+class ValidateInlineCodeTests(unittest.TestCase):
+    def _validate(self, orig: str, comp: str):
+        with tempfile.TemporaryDirectory() as tmp:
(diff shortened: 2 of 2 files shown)


Qodo-Free-For-OSS:
Hi, `validate_inline_code` correctly compares ordered lists, but its error message uses set-diff for `lost`/`added`, which can print `lost=[] added=[]` even when validation fails due to reordering or duplicate-count changes. This makes the failure hard to diagnose and gives the auto-fix loop less actionable context.

**Severity:** remediation recommended | **Category:** observability

**How to fix:** Log first differing span

> ### Issue description
> Inline-code validation failure logs can be misleading because the sample uses set difference, losing ordering and duplicate counts.
> 
> ### Issue Context
> Validation is correct (`i1 != i2`), but the diagnostic message should help identify the first mismatch.
> 
> ### Fix Focus Areas
> - skills/compress/scripts/validate.py[147-168]
> 
> ### What to do
> - Replace the set-diff sample with an order-preserving diff, e.g.:
>   - find the first index where spans differ and log `idx`, `orig_span`, `comp_span`
>   - optionally include length mismatch info and small window around the mismatch
>   - keep the `[:5]` cap if you also include added/removed items from a multiset/count-based diff (e.g., `collections.Counter`)

---

ousamabenyounes:
Superseded by #309 (by @hireblackout) which solved the same silent-overwrite gap and was bundled into commit `57a9b48` on May 1. `validate_inline_codes` is wired into `validate()` and covered by 11 tests in `tests/test_validate_inline.py`. Closing.

Issue #240, opened 2026-04-20
Clear instructions for Codex native app


juancamilo994:
A better explanation on how to implement this in the Codex native app

juancamilo994:
Since it's available on the CLI and the app has plugin support, is it possible to implement? I tried many ways yesterday but wasn't able to do it!

Thanks!

takomine:
Does this work in codex cli?

s-cork:
once you've opened the caveman plugin repo
then go to the plugins in the sidebar
and you should see the caveman repo plugin available

<img width="1200" height="223" alt="Image" src="https://github.com/user-attachments/assets/c88ac952-a4b4-4e87-b5d9-157e78a52770" />

oysandvik94:
It seems like it's only active if I work codex cli from the Caveman repo, but not visible if opened from other repos?

nloadholtes:
In the codex cli I was able to turn it on by using the `/skills` command.

The shortcut for this is `$caveman:caveman lite`

ugurgok:
I followed [readme.md](https://github.com/JuliusBrussee/caveman/blob/main/README.md) but can't see caveman as a plugin. 

<img width="987" height="631" alt="Image" src="https://github.com/user-attachments/assets/8fe9eeb3-cb0a-402b-a4ec-42b6ce30bd74" />

fjdiazt:
I asked Codex to fix the local plugin loading issue for me. It found that the plugin repo was cloned correctly, but Codex Desktop was only reading the official marketplaces because my local marketplace was not registered in my Codex config file.

**C:\Users\\\<USER>\\\.codex\config.toml**

```
[marketplaces.caveman-repo]
source_type = "local"
source = '\\?\C:\Users\<USER>\.codex\plugins'

[plugins."caveman@caveman-repo"]
enabled = true

```
After restarting Codex Desktop, the local marketplace should be available in the Plugins window. The plugin needs to be searched for there and installed from that window;  in the dropdown.

Once installed, it can be invoked with:

`$caveman
`

gilsedition:
Opened a docs PR with a Windows native app quick start and fallback path: #301

Ralph-Samra:
> once you've opened the caveman plugin repo then go to the plugins in the sidebar and you should see the caveman repo plugin available
> 
[...]

PR #301, opened 2026-04-28
docs: add Codex native app Windows quick start

gilsedition:
Refs #240

Adds fallback instructions for Codex native app users on Windows when plugin discovery or $caveman does not resolve after reload.

Includes:
- immediate prompt-only workaround
- direct local skill install path
- restart and skill attachment guidance


JuliusBrussee:
Thanks @gilsedition — current Windows docs include Codex install and manual `$caveman` fallback. Closing as implemented.

PR #302, opened 2026-04-28
Add precise intensity level to caveman skill

juansilvadesign:
## Summary

- Add `precise` as a new intensity level, between `lite` and `full`.
- Define its ruleset: drop filler and hedging, keep articles where they prevent ambiguity, require named technical nouns (tables, endpoints), use exact verbs (`upsert`, `truncate`). Priority: zero ambiguity.
- Add `precise` examples to both the React re-render and DB connection pooling illustrations.
- Document the tradeoff: `precise` costs ~20 extra tokens over `full`/`ultra` but preserves which table, which DB, and how operations work. Worth it when a wrong noun corrupts data.

## Before

`full` drops articles and short-synonyms the nouns.

In the DB pooling example, `full` says "skip handshake overhead." You know something is being skipped, not that it's a TCP handshake, TLS negotiation, and auth round-trips. In data-sensitive work (migrations, dedup logic, upserts), that missing specificity causes real errors, not just awkward output.

The gap between `lite` (full sentences, all articles) and `full` (fragments OK, articles gone) was wide. Neither fit contexts where you need both compression and precision.

## After

Six levels. `precise` sits in the middle: kills filler and hedging, keeps articles where they add clarity, requires named technical nouns and exact verbs. The DB pooling example at `precise` names TCP connections, TLS negotiation, and authentication round-trips. `full` does not.

The skill doc spells out the tradeoff rather than leaving it implied: `precise` costs more tokens than `full` or `ultra`, and it's worth it specifically when a wrong noun corrupts data.

## Why better

LLMs resolve ambiguity statistically through attention, based on what's most likely given context. Strip articles and technical nouns, and you're not just making output shorter. You're asking the model to guess the missing noun. For explanation tasks, a wrong guess produces awkward prose. For back-office logic (upserts, migrations, dedup) a wrong guess corrupts rows.

`full` and `ultra` were the two compression options before this. Both drop nouns. That works when brevity matters more than specificity. But there was no level that compressed the filler (hedging, pleasantries, vague verbs) while keeping the nouns the model needs to stay on the right table, endpoint, and operation.

`precise` makes that distinction explicit. The token cost over `full` is small (~20 per response). The reliability difference in data-sensitive tasks is not.

## Testing

- Read the updated `SKILL.md` and verify level ordering, rule descriptions, and both examples are consistent.
- Spot-check: the `precise` DB pooling example names TCP connections, TLS negotiation, and authentication round-trips. `full` does not. That difference is the feature.
- `git diff --check`

## AI Disclosure

Skill modified with Claude Code Opus 4.7 and reviewed with Gemini 3 (thinking). I reviewed and modified the file before submission.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -30,7 +30,7 @@ Yes: "Bug in auth middleware. Token expiry check use `<` not `<=`. Fix:"
 | Level | What change |
 |-------|------------|
 | **lite** | No filler/hedging. Keep articles + full sentences. Professional but tight |
-| **precise** | Kill filler/hedging. Keep articles for structural clarity. Require named technical nouns (tables/endpoints). Use exact verbs (upsert / truncate). Priority: Zero ambiguity. |
+| **precise** | Kill filler/hedging. Keep articles for structural clarity. Require named technical nouns (tables/endpoints). Require specific technical verbs (`upsert` over `save`, `throttle` over `slow down`, `backfill` over `populate`). Priority: Zero ambiguity. |
 | **full** | Drop articles, fragments OK, short synonyms. Classic caveman |
 | **ultra** | Abbreviate (DB/auth/config/req/res/fn/impl), strip conjunctions, arrows for causality (X → Y), one word when one word enough |
 | **wenyan-lite** | Semi-classical. Drop filler/hedging but keep grammar structure, classical register |
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -2,7 +2,7 @@
 name: caveman
 description: >
   Ultra-compressed communication mode. Cuts token usage ~75% by speaking like caveman
-  while keeping full technical accuracy. Supports intensity levels: lite, full (default), ultra,
+  while keeping full technical accuracy. Supports intensity levels: lite, precise, full (default), ultra,
   wenyan-lite, wenyan-full, wenyan-ultra.
   Use when user says "caveman mode", "talk like caveman", "use caveman", "less tokens",
   "be brief", or invokes /caveman. Also auto-triggers when token efficiency is requested.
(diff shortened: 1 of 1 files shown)


juansilvadesign:
I didn't add, but a more explicit example would be:

Example — "Save the new orders to the database."
- lite: "Save the validated orders to the database. Handle duplicates by external ID."
- precise: "Upsert validated rows into the orders table; conflict on external_id triggers update. Postgres: INSERT ... ON CONFLICT (external_id) DO UPDATE."
- full: "Save validated orders. Dedup by external_id."
- ultra: "Upsert orders. Dedup ext_id."

`full` and `ultra` lose *which* table, *which* DB, and *how* dedup works. `precise` keeps all three at the cost of ~20 extra tokens. Worth it when a wrong noun corrupts data.

wunderman4:
**Supportive of adding `precise`.** The framing of "compress filler, preserve technical nouns" reads cleanly and the DB-pooling example makes the intent concrete.

One direction worth considering inside `precise`, or as a stricter sub-level:

**Verb specificity, alongside noun preservation.**

The noun rule covers "which thing." A verb rule could cover "which operation." Verbs like `rectify` vs `fix`, `deduplicate` vs `clean up`, `throttle` vs `slow down`, `backfill` vs `populate`, `rotate` vs `change` carry different operational meaning. If the goal is to preserve what the model needs to stay accurate, technical verbs may belong in the same protected category as technical nouns.

Suggested rule (additive to current `precise` spec):

> When a more specific technical verb is available and unambiguous in context, prefer it. e.g. `deduplicate` over `clean up`, `throttle` over `slow down`, `backfill` over `populate`, `rotate` over `change`.

Plausible use cases (untested, offered as hypotheses): incident write-ups, PR review comments, migration plans — contexts where the verb describing an action carries audit / clarification weight.

Two shipping options:

1. **Fold into `precise`** — one rule covers both nouns and verbs. Closer to "caveman like simple."
2. **Add a `precise-strict` level** — keeps `precise` as proposed, adds stricter mode that also enforces verb specificity.

Either way, the direction of the PR makes sense for data-sensitive work. Happy to help with examples or follow-up changes if useful.

juansilvadesign:
@wunderman4 Thanks for pointing that out. I folded into the `precise` intensity level. You're right that the row had nouns as a top-level rule and verbs tucked into parens, which read as "noun rule plus footnote" rather than two parallel rules.

JuliusBrussee:
Thanks @juansilvadesign — `precise` is unmeasured and unsupported across the runtime contract. Closing this prompt-only seventh mode.

Issue #251, opened 2026-04-21
Usage statistics: Is it possible to check how much tokens were saved?


iamtodor:
**What you want**

I'm wondering if the `caveman` is being properly configured cause for 2 small tasks I've burnt around 35% 5 hours limit tokens.

**Why good**

Its beneficial to see some statistics

sr-gene:
me agree, me want this

angellopezmonzoncillo:
I’ve been running some tests and comparing the results of asking GitHub Copilot for the same task both with and without Caveman. The comparison shows 10.1k tokens consumed with Caveman versus 12k tokens consumed without it

DeeptimaanB:
This sounds like a cool idea, I will work on this, will mostly work on the statistics aspect as it will be hard to predict how many tokens claude or codex would have used for normal output vs this. Will look into this and create a pull request soon!

RaphaelS-git:
Maybe this workaround helps codex users:

```bash
npx @ccusage/codex
```

It can show Codex token usage over time from local session logs. It won’t directly prove “tokens saved by Caveman,” but it helps compare runs before/after or with/without Caveman.

DeeptimaanB:
I have added a new command /caveman-stats to see the estimated savings for claude code (for now) which works for the current session. The command does not use AI tokens and runs a script directly to measure. Will add it to other agents as well in the coming days.

ur2464:
Hey @DeeptimaanB, thanks for adding /caveman-stats. I was wondering if I am using it correctly, in my VS code claude extension, invoking /caveman-stats results this:

```
/caveman-stats
Stats from hook.
```

llagerlof:
@DeeptimaanB 

> I have added a new command /caveman-stats to see the estimated savings for claude code (for now) which works for the current session. The command does not use AI tokens and runs a script directly to measure. Will add it to other agents as well in the coming days.

The rest of us are using Caveman in VS Code. It would be very helpful to have this capability available there as well. Thank you!

DeeptimaanB:
Understandable, currently it runs from the terminal, and I am working on extending it to codex and gemini-cli. Let me figure out a way to make it work on VS Code. The current version is to be invoked via the terminal.

mturnaviotov:
@DeeptimaanB  anso for Antigravity users. it would be nice. thank you

JuliusBrussee:
[...]

PR #305, opened 2026-04-28
enh: add /caveman-stats command for session token usage stats (Claude…

DeeptimaanB:
… Code only)

Addresses #251

Caveman cut token. But user no know how many token caveman cut. This fix that.

## What

New `/caveman-stats` command. Type it, see real numbers from your session. No AI, no guessing, reads directly from the Claude Code session file that gets written to disk as you chat. Hook intercepts the command before Claude sees it, so running stats costs nothing.

Output look like this:

```
Caveman Stats
──────────────────────────────────
Session:  .../<session>.jsonl
Turns:    12
──────────────────────────────────
Output tokens:         1,840
Cache-read tokens:     204,112
──────────────────────────────────
Est. without caveman:  5,257
Est. tokens saved:     3,417 (~65%)
──────────────────────────────────
Savings est. from 10-task benchmark (benchmarks/). Actual varies by task.
```

Savings estimate only shown for `full` mode, that the only mode with real benchmark data. No made-up numbers for other modes.

## Scope

Claude Code only for now. Gemini and Codex write session data to disk too (confirmed), but need live testing with working API keys before shipping (me no have now). Coming in a separate PR.

### DIFF
--- a/commands/caveman-stats.toml
+++ b/commands/caveman-stats.toml
@@ -1,0 +1,2 @@
+description = "Show token usage and estimated savings for the current session"
+prompt = "Run caveman-stats and display the output."
--- a/hooks/caveman-mode-tracker.js
+++ b/hooks/caveman-mode-tracker.js
@@ -5,6 +5,7 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
+const { execFileSync } = require('child_process');
 const { getDefaultMode, safeWriteFlag, readFlag } = require('./caveman-config');
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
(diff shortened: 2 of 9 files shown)


Qodo-Free-For-OSS:
Hi, The /caveman-stats intercept writes a structured block decision to stdout but then exits with code 2, which is likely to be treated as a hook execution failure and prevent the reason from being surfaced reliably.

**How to fix:** Exit 0 after block

> ### Issue description
> `hooks/caveman-mode-tracker.js` emits `{ decision: 'block', reason: ... }` for `/caveman-stats`, but then calls `process.exit(2)`. That non-zero exit is likely treated as a hook failure by the host, which can suppress/override the block response.
> 
> ### Issue Context
> Other hook success paths exit 0 (or naturally exit). Installers configure `timeout: 5` for the hook and expect it to succeed.
> 
> ### Fix Focus Areas
> - hooks/caveman-mode-tracker.js[34-51]
> 
> ### Suggested change
> After writing the block decision, exit with code `0` (or just `return;` and allow normal exit).

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---
*Spotted by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source projects.*

DeeptimaanB:
I used exit code 2 as it was mentioned as a way in claude code's official docs (see: https://code.claude.com/docs/en/hooks). Regardless, exit code 0 will still work. So, I am making that change. Thanks!

PR #292, opened 2026-04-26
fix(compress): bail loudly on empty / unchanged compression instead of silent data loss

ousamabenyounes:
Closes #237

## Summary

`/caveman:compress` could finish with `Validation passed` / `Compression completed successfully` while leaving the input file unchanged and the `*.original.md` backup empty (or vice versa). The orchestrator trusted Claude's response and wrote both files unconditionally, so three failure modes slipped through:

1. Claude returned an empty (or whitespace-only) string — overwrote both files without complaint.
2. Claude returned the input verbatim (refusal, echo) — passed validation trivially because there were no diffs to check.
3. The backup write completed on disk but with bytes dropped (Windows encoding / antivirus quarantine) — not detected before the input file was overwritten.

Defensive checks added before any write touches the input:

- Refuse empty-or-whitespace **input** up front.
- Refuse empty-or-whitespace **compressed output** before opening any destination file.
- Refuse compressed output that is identical to the input after `strip()`.
- After writing the backup, read it back and compare against the in-memory original; on mismatch, unlink the bad backup and abort before the input file is touched.

Each refusal prints a concrete reason and a "no backup created" line so the user does not have to guess whether they lost data.

## Test plan

- [x] New `tests/test_compress_safety.py` — five unit tests covering: empty input refused, empty compressed output rejected, whitespace-only compressed output rejected, identical compressed output rejected, real compression writes both files. `python3 -m unittest tests.test_compress_safety` passes 5/5.
- [x] All five tests confirm the input file is left byte-identical and **no** `.original.md` is created on the rejection paths — the exact symptom from the bug report.

PR #310, opened 2026-04-28
fix(hooks): bail loud on unrecognised /caveman arg, add explicit off + full

chriscow:
## What

`hooks/caveman-mode-tracker.js` is the `UserPromptSubmit` hook that parses `/caveman` slash commands and writes the active mode to `~/.claude/.caveman-active`. Statusline + per-turn reinforcement read that flag.

The slash-command parser handled the known modes (`lite`, `ultra`, `wenyan-lite`, `wenyan`, `wenyan-full`, `wenyan-ultra`) with explicit branches. Anything else fell through to `else mode = getDefaultMode()`, which silently re-asserted the default mode and rewrote the flag file as if the user had typed `/caveman` with no arg.

The user-facing impact: someone types a reasonable guess for "turn it off" — most commonly `/caveman off`, but also `/caveman quiet`, `/caveman normal`, etc. — and gets no error, no usage hint, and no signal that their input was unrecognised. Caveman behaviour continues while they assume they have disabled it. The README documents `stop caveman` and `normal mode` as natural-language deactivation paths (handled by the regex matcher further down the same file), but the slash-command equivalent of "off" did not exist; the silent fall-through hid the gap.

Same principle as #292 (`bail loudly on empty / unchanged compression instead of silent data loss`) — fail loud, don't silently substitute.

## Repro

```sh
# Set caveman to a known state
echo "lite" > ~/.claude/.caveman-active

# Run the hook with /caveman off
echo '{"prompt":"/caveman off"}' | node hooks/caveman-mode-tracker.js

# Observed (before fix): flag file rewritten to default mode
cat ~/.claude/.caveman-active
# full

# Expected: flag file deleted (mode off) OR usage message shown for unknown args.
```

## Fix

Replace the silent fall-through with an explicit unknown-arg branch that emits a usage message to stderr and leaves the flag file untouched. Add explicit `off` and `full` cases so users can deactivate via slash command and so `full` doesn't rely on a default-mode coincidence.

```diff
       } else if (cmd === '/caveman' || cmd === '/caveman:caveman') {
-        if (arg === 'lite') mode = 'lite';
+        if (arg === '') mode = getDefaultMode();
+        else if (arg === 'lite') mode = 'lite';
+        else if (arg === 'full') mode = 'full';
         else if (arg === 'ultra') mode = 'ultra';
         else if (arg === 'wenyan-lite') mode = 'wenyan-lite';
         else if (arg === 'wenyan' || arg === 'wenyan-full') mode = 'wenyan';
         else if (arg === 'wenyan-ultra') mode = 'wenyan-ultra';
-        else mode = getDefaultMode();
+        else if (arg === 'off') mode = 'off';
+        else {
+          // Unknown arg: fail loud with usage info, leave flag file untouched.
+          process.stderr.write(
+            "caveman: unknown arg '" + arg + "'. " +
+            "Usage: /caveman [lite|full|ultra|wenyan-lite|wenyan-full|wenyan-ultra|off]. " +
+            "Plain-English alternatives: 'stop caveman' or 'normal mode'.\n"
+          );
+          // mode stays null — flag file untouched, per-turn reinforcement
+          // continues for the current active mode.
+        }
       }
```

The existing `else if (mode === 'off')` branch already handles flag-file deletion when `mode === 'off'`, so adding `arg === 'off'` to the parser hooks straight into that path with no further code changes needed.

The unknown-arg path deliberately leaves `mode` as `null` rather than `return`-ing early. That way the per-turn reinforcement at the bottom of the same callback still fires for the currently active mode — a user who fat-fingers `/caveman foo` doesn't lose mode reinforcement that turn.

### Behaviour after the fix

| Input | Before | After |
| --- | --- | --- |
| `/caveman` (no arg) | Default mode applied | Default mode applied (unchanged) |
| `/caveman lite` | `lite` applied | `lite` applied (unchanged) |
| `/caveman full` | Default mode applied (coincidence) | `full` applied explicitly |
| `/caveman ultra` | `ultra` applied | `ultra` applied (unchanged) |
| `/caveman wenyan-*` | Each variant applied | Each variant applied (unchanged) |
| `/caveman off` | Default mode applied (silent) | Flag file deleted (off) |
| `/caveman foo` | Default mode applied (silent) | stderr usage; flag file unchanged |

## Notes

Discovered while debugging a session where `/caveman off` had no observable effect — a user-visible failure mode for anyone who hasn't read the natural-language regex in this same file. The behavioural fix is small; the larger change is moving the slash-command parser onto the same fail-loud footing as the rest of the project.

Happy to add a behavioural test for the parser if you'd like — `tests/test_hooks.py` currently only covers `install.sh`. I held off here to keep the PR scope minimal, but I can extend it in a follow-up.

### DIFF
--- a/hooks/caveman-mode-tracker.js
+++ b/hooks/caveman-mode-tracker.js
@@ -45,12 +45,27 @@ process.stdin.on('end', () => {
       } else if (cmd === '/caveman-compress' || cmd === '/caveman:caveman-compress') {
         mode = 'compress';
       } else if (cmd === '/caveman' || cmd === '/caveman:caveman') {
-        if (arg === 'lite') mode = 'lite';
+        if (arg === '') mode = getDefaultMode();
+        else if (arg === 'lite') mode = 'lite';
+        else if (arg === 'full') mode = 'full';
         else if (arg === 'ultra') mode = 'ultra';
         else if (arg === 'wenyan-lite') mode = 'wenyan-lite';
         else if (arg === 'wenyan' || arg === 'wenyan-full') mode = 'wenyan';
         else if (arg === 'wenyan-ultra') mode = 'wenyan-ultra';
-        else mode = getDefaultMode();
+        else if (arg === 'off') mode = 'off';
+        else {
+          // Unknown arg: fail loud with usage info, leave flag file untouched.
+          // Silently substituting the default surprised users who guessed
+          // reasonable phrasings like /caveman off — they got no signal that
+          // their input was unrecognised and assumed it had worked.
+          process.stderr.write(
+            "caveman: unknown arg '" + arg + "'. " +
+            "Usage: /caveman [lite|full|ultra|wenyan-lite|wenyan-full|wenyan-ultra|off]. " +
+            "Plain-English alternatives: 'stop caveman' or 'normal mode'.\n"
+          );
+          // mode stays null — flag file is not modified, per-turn reinforcement
+          // (further down) still fires for the current active mode.
+        }
       }
 
       if (mode && mode !== 'off') {


JuliusBrussee:
Thanks @chriscow — current parser handles unknown levels, off, and full explicitly. Closing as implemented.

PR #314, opened 2026-04-29
fix(hooks): handle all /caveman <arg> values, drop silent fallback

jgttech:
## Summary

`hooks/caveman-mode-tracker.js` only matched a subset of valid args for `/caveman <arg>`. The rest fell through to `else mode = getDefaultMode()`, which silently overwrote the flag with the configured default instead of doing what the user asked.

Three concrete bugs:

1. **`/caveman full` is impossible** when `defaultMode` is anything other than `full`. The arg has no explicit branch, so it falls through to `getDefaultMode()`. If a user has `~/.config/caveman/config.json` set to `{"defaultMode":"lite"}`, typing `/caveman full` writes `lite` to the flag file. The statusline (and any consumer of `.caveman-active`) keeps showing `lite`. There is no way to switch to full via the slash command.
2. **`/caveman off` is impossible** for the same reason. Off-switching only works via the natural-language regex (`stop caveman` / `normal mode`).
3. **Typos silently clobber state.** `/caveman fulll` (typo) overwrites the flag with the default mode rather than no-op'ing. A user on `ultra` gets bumped to `full` by a stray keystroke.

## Fix

- Use the existing `VALID_MODES` whitelist from `caveman-config.js` as the source of truth instead of a hand-maintained `if/else` chain that has already drifted.
- Explicit branches for the cases that need special handling: bare `/caveman` (default), `off|stop|disable` (delete flag), `wenyan-full` alias (canonicalize to `wenyan`).
- Unknown arg, `mode` stays `null`, flag untouched (no silent overwrite).
- Hoist `INDEPENDENT_MODES` to module scope so the dispatch can reject `/caveman commit` (those modes have their own `/caveman-commit` etc. commands and should not be selectable as args).

## Why this matters beyond the plugin

Anything that consumes `~/.claude/.caveman-active` is at the mercy of this hook. Example: [securacore/kanagawa-statusline](https://github.com/securacore/kanagawa-statusline) renders a `caveman <level>` badge by reading the flag directly (`statusline.sh:230-244`, with proper symlink/size/whitelist hardening matching `readFlag`). When the user runs `/caveman full` and the hook writes `lite` instead, the statusline correctly displays `lite`, so the bug looks like a statusline bug to the user. Multiply across every downstream reader.

## Test plan

- [x] Existing `tests/test_hooks.py` passes.
- [x] All `/caveman <arg>` transitions verified manually:
  - `lite/full/ultra/wenyan-lite/wenyan/wenyan-full/wenyan-ultra` write expected mode
  - `off` deletes flag
  - bare `/caveman` writes configured default
  - typo / unknown arg leaves flag unchanged (no silent overwrite)
  - `/caveman commit` leaves flag unchanged (independent mode, not selectable as arg)
- [x] Natural-language regression: `activate caveman` / `talk like caveman` / `stop caveman` / `normal mode` / `caveman mode please` all behave as before.

## Notes

- One behavioral change worth flagging: the old code wrote the default mode for any unknown arg. The new code no-ops. If a user was relying on `/caveman whatever` to "reset to default," they will need `/caveman` (no arg) instead, which is the documented form.
- `INDEPENDENT_MODES` was already defined inline at the bottom of the file for the per-turn reinforcement check. Hoisting it makes both call sites use the same set instead of duplicating literals.

### DIFF
--- a/hooks/caveman-mode-tracker.js
+++ b/hooks/caveman-mode-tracker.js
@@ -5,7 +5,11 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
-const { getDefaultMode, safeWriteFlag, readFlag } = require('./caveman-config');
+const { getDefaultMode, safeWriteFlag, readFlag, VALID_MODES } = require('./caveman-config');
+
+// Modes handled by their own slash commands (/caveman-commit, etc.) — not
+// selectable via /caveman <arg>.
+const INDEPENDENT_MODES = new Set(['commit', 'review', 'compress']);
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
 const flagPath = path.join(claudeDir, '.caveman-active');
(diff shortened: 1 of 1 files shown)


jgttech:
Just so you know, I did actually do this and built this status line tool so I could use Claude Code with your tool. This has been great and I want to include the state of it (and other things) in the status line so that users can leverage a better experience with Claude Code.

JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`a660db02`](https://github.com/JuliusBrussee/caveman/commit/a660db02) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #317, opened 2026-04-30
fix(compress): first-original wins — recompress in place when backup exists

aminalavian:
## Problem

`/caveman:compress` aborts with a warning if `<file>.original.md` already exists:

```
⚠️ Backup file already exists: foo.original.md
The original backup may contain important content.
Aborting to prevent data loss. Please remove or rename the backup file if you want to proceed.
```

The intent is safe — don't silently overwrite the original. But in practice, re-compressing a previously-compressed file (after editing it, or after a failed compression) requires the user to manually remove the backup before proceeding. If they rename it with a timestamp instead, backup files proliferate across runs.

## Fix: first-original wins

If a canonical backup already exists:
- Keep it untouched — it holds the true original, not the already-compressed intermediate
- Recompress the current file in place
- Skip creating a new backup (the canonical one is already there)

Only the **first** compress run of a file writes `<file>.original.md`. Every subsequent run is idempotent with respect to the backup.

## Failure path

Rollback logic updated: only `unlink` the backup if **this run** created it. A pre-existing canonical backup is never deleted on failure or exception — preserving the true original no matter what.

## Behaviour comparison

| Scenario | Before | After |
|---|---|---|
| First compress | Creates `foo.original.md` ✅ | Creates `foo.original.md` ✅ |
| Second compress (edited file) | Aborts ❌ | Recompresses in place ✅ |
| Failed compress (pre-existing backup) | Aborts before attempting ❌ | Attempts, restores on failure ✅ |
| Exception mid-compress | Deletes backup (even canonical) | Preserves pre-existing backup ✅ |

### DIFF
--- a/skills/compress/scripts/compress.py
+++ b/skills/compress/scripts/compress.py
@@ -182,46 +182,59 @@ def compress_file(filepath: Path) -> bool:
     original_text = filepath.read_text(errors="ignore")
     backup_path = filepath.with_name(filepath.stem + ".original.md")
 
-    # Check if backup already exists to prevent accidental overwriting
-    if backup_path.exists():
-        print(f"⚠️ Backup file already exists: {backup_path}")
-        print("The original backup may contain important content.")
-        print("Aborting to prevent data loss. Please remove or rename the backup file if you want to proceed.")
-        return False
+    # First-original wins: if a canonical backup already exists, keep it untouched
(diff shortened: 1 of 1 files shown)


aminalavian:
Closing — removing caveman from our toolchain entirely; the auto-compress hook turned out to be net-negative for active editing in our repo. Thanks for the project.

PR #322, opened 2026-05-01
fix: tighten caveman full output

Hidromoss:
Tightens caveman full output to avoid verbose tool narration, tables, emoji, and long error dumps unless explicitly requested.

Also clarifies that standard technical acronyms are fine, while invented/ambiguous abbreviations should be avoided, and improves the verifier on Windows/Git Bash.


Hidromoss:
## Summary

This PR tightens the existing `full` caveman mode. It does not add a new mode.

The goal is to keep `full` compact while avoiding output patterns that make answers noisy or harder to scan:

- verbose tool narration
- tables when plain text is enough
- emoji in normal technical answers
- long log/error dumps
- invented or ambiguous abbreviations

Standard technical acronyms are still allowed. The rule is meant to preserve clarity, not make output more formal.

## Before

`full` already removed filler, hedging, and many articles, but it could still drift into verbose scaffolding around tool use or dump too much surrounding error/log text.

That costs tokens and makes answers slower to read, especially when the useful part is only one specific failure phrase.

## After

`full` now explicitly says:

- no tool narration unless asked
- no tables or emoji unless asked
- no long error dumps; quote the shortest decisive phrase
- standard acronyms are OK
- invented/ambiguous abbreviations should be avoided

This keeps the mode compact without making technical answers cryptic.

## Verifier update

While testing locally on Windows, I hit a case where `bash.exe` existed as a WSL launcher, but no WSL distro was installed. The verifier treated that as unusable bash and skipped Unix hook checks, even though Git Bash was installed.

The verifier now falls back to Git Bash when available, so local verification can still run the Unix hook install/statusline checks.

## Testing

Local checks passed:

- `uv run python tests/verify_repo.py`
- `node tests/test_caveman_init.js`
- `git diff --check`

Also verified that tracked public files do not contain private shorthand/research wording.

## AI Disclosure

Prepared with GitHub Copilot assistance GPT 5.5
I reviewed and tested the changes before submission.

JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`f06348cb`](https://github.com/JuliusBrussee/caveman/commit/f06348cb) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #333, opened 2026-05-04
feat: add /caveman slash command registration

manuelacorreabet:
## What

Adds `commands/caveman.md` so Claude Code recognises `/caveman` as a valid slash command.

## Why

Without this file, the Claude Code CLI intercepts `/caveman` and returns `Unknown command` before the `UserPromptSubmit` hook can process it. Users who try `/caveman` immediately after installing the plugin see the error and assume the plugin is broken.

## How it works

The `commands/caveman.md` file registers `/caveman [lite|full|ultra|off|wenyan]` as a first-class slash command. When invoked, Claude parses the argument and activates the corresponding intensity level, then persists caveman style for the rest of the session.

## Reviewer notes

- The existing `SessionStart` and `UserPromptSubmit` hooks continue to work as before — this file only adds the missing CLI registration.
- Tested on Windows 10 with Claude Code after plugin install.

### DIFF
--- a/commands/caveman.md
+++ b/commands/caveman.md
@@ -1,0 +1,24 @@
+---
+description: Activate caveman mode — ultra-compressed responses that cut ~75% of tokens. Usage: /caveman [lite|full|ultra|off]
+---
+
+Activate caveman communication mode at the requested intensity level.
+
+The user invoked `/caveman $ARGUMENTS`.
+
+Parse $ARGUMENTS:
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks @manuelacorreabet — this file shadows the auto-discovered `caveman` skill and can make the wrong body win. Closing this prohibited duplicate surface.

PR #346, opened 2026-05-07
feat: native opencode installer — direct skills copy to ~/.agents/skills/

juanitourquiza:
## What
Replaces `npx skills add (opencode)` fallback with a native filesystem installer that copies caveman skills directly to `~/.agents/skills/`.
## Why
The current installer routes opencode through `npx skills add`, which:
- Requires Node.js and internet connectivity even for local repo clones
- Adds an unnecessary intermediary (`vercel-labs/skills`) when opencode reads skills directly from the filesystem
- Is slower and less reliable than a direct copy
A native installer mirrors how Claude Code gets first-class treatment (plugin + hooks) and makes opencode installation offline-friendly.
## How it works
Two new install functions handle opencode directly:
**`install_opencode()`** in `install.sh` (macOS/Linux/WSL)
- Detects `~/.agents/skills/` directory or creates it
- Copies all 7 skills from the local repo clone when available
- Falls back to `curl` downloading individual `SKILL.md` files for pipe installs
- Handles `compress/scripts/` subdirectory downloads
- Supports `--force` to overwrite existing skills
- Skips already-installed skills by default
**`Install-Opencode`** in `install.ps1` (Windows/PowerShell)
- Equivalent PowerShell implementation using `Copy-Item` and `Invoke-WebRequest`
- Same offline/online dual mode and idempotency
Both remove opencode from the generic `npx skills` fallback path, so it no longer appears in `SKILLS_AGENTS`.
## Also updated
| File | Change |
|---|---|
| `tools/caveman-init.js` | Adds `opencode` as an `AGENTS.md` target, so `--with-init` writes caveman activation rules for opencode users |
| `README.md` | Documents opencode as a native install alongside Claude Code and Gemini CLI |
| `tests/test_opencode_install.py` | 5 integration tests covering local copy, skip, force, compress subdirs, and dry-run |
## Reviewer notes
- The existing `npx skills add` path continues to work for all other agents — only opencode is extracted
- `install.sh` and `install.ps1` stay aligned: same detection probe, same skill list, same idempotency logic
- `AGENTS_SKILLS_DIR` env var overrides the default `~/.agents/skills/` path in both scripts
- Tested on macOS with `bash install.sh --only opencode --dry-run` and `python3 tests/test_opencode_install.py` (5/5 pass)

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,8 @@ By default the installer wires Claude Code's hooks + statusline + stats badge an
 | **Claude Code** | `claude plugin marketplace add JuliusBrussee/caveman && claude plugin install caveman@caveman` |
 | **Gemini CLI** | `gemini extensions install https://github.com/JuliusBrussee/caveman` |
 | **Cursor / Windsurf / Cline / Copilot** | `npx skills add JuliusBrussee/caveman -a <cursor\|windsurf\|cline\|github-copilot>` |
-| **Codex / opencode / Roo / Amp / Goose / Kiro / Augment / Aider Desk / Continue / Kilo / Junie / Trae / Warp / Tabnine / Mistral / Qwen / Devin / Droid / ForgeCode / Bob / Crush / iFlow / OpenHands / Qoder / Rovo Dev / Replit / Antigravity** | `npx skills add JuliusBrussee/caveman -a <profile>` (see `install.sh --list` for the full slug list) |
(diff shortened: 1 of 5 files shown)


JuliusBrussee:
Thanks @juanitourquiza — the unified installer now owns OpenCode plugin and skills delivery. Closing this standalone installer.

PR #365, opened 2026-05-10
fix install script breaking when piped via curl

### LINKED ISSUE #364: One line install fails on mac when piped (`curl | bash`)
When I ran the install command from the README

```sh
```

it failed and gave this error

```sh
bash: line 39: BASH_SOURCE[0]: unbound variable
error: unknown flag: --
run 'caveman --help' for usage
```

OS: MacOS

Two things going on here

1. `BASH_SOURCE[0]` isn't set when the script is read from stdin, and `set -u` kills the process before the `|| here=""` fallback can catch it.
2. The last line passes a bare `--` to the installer (since `$@` is empty), which it doesn't understand.

Works fine when you clone and run `bash install.sh` directly. It's only the curl-pipe path that's broken.

max-programming:
Fixes #364

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -36,7 +36,7 @@ fi
 
 # If we're inside the repo clone, run the local installer directly — saves
 # the npx round-trip and keeps offline installs working.
-here="$(cd "$(dirname "${BASH_SOURCE[0]}")" 2>/dev/null && pwd)" || here=""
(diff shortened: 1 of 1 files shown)


max-programming:
looks like it's fixed with commit 754795ada42dea54adf061d42a61e560caa4f9ce minutes ago lol

PR #380, opened 2026-05-13
fix: Skip MCP Shrink by Default

### LINKED ISSUE #353: MCP Server Fails
**What happen**
MCP server is failing

**Expected**
MCP server should work

**Before/after example**

Image

Image

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**

Version `ef6050c5e184`, installed through default bash installer

DerekY2:
closes #353 

This is my quicktake on this issue.

## Problem
caveman-shrink is an MCP proxy that requires an upstream MCP server command as args. The installer was registering it via `claude mcp add caveman-shrink -- npx -y caveman-shrink` with no upstream during installation, causing [/src/mcp-servers/caveman-shrink/index.js#L33](https://github.com/JuliusBrussee/caveman/blob/63a91ecadbf4c4719a4602a5abb00883f9966034/src/mcp-servers/caveman-shrink/index.js#L33) to immediately exit with code 2. As a result, the server fails in Claude Code on every startup. `--with-mcp-shrink` was ON by default, so every fresh install contained this broken entry (assuming no upstream provided).

## Fix:
  - Install caveman with `--with-mcp-shrink` argument set to FALSE (rather than auto) by default.
  - Change to `--with-mcp-shrink="<upstream cmd>"` to enforce `mcp shrink + provide upstream`
  - Hints at upstream requirement at install
  - Attempting to register caveman-shrink alone results in an error (+ hint i.e. must add upstream.)
  - (+ a few UT for updated behavior)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -22,7 +22,7 @@ What it does:
 
 - Auto-detects every supported agent installed on your machine (Claude Code, Cursor, Codex, etc.).
 - For each one, runs that agent's native install path (plugin / extension / rule file / `npx skills add`).
-- Wires Claude Code hooks, statusline badge, and the `caveman-shrink` MCP middleware on top.
+- Wires Claude Code hooks and statusline badge on top. (`caveman-shrink` MCP middleware is opt-in via `--with-mcp-shrink` — see flag table below.)
 - Skips anything you don't have. Safe to re-run. ~30 seconds end-to-end.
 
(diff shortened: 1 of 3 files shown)


DerekY2:
@JuliusBrussee

DerekY2:
another approach could be to simply use a default location, but let me know what we think. #353

JuliusBrussee:

PR #387, opened 2026-05-13
fix(mcp-shrink): allow npx and .cmd upstream on Windows via shell:true

limon171:
## Symptom

On Windows 11:

```
$ echo '...' | npx -y caveman-shrink npx -y @modelcontextprotocol/server-everything
caveman-shrink: failed to spawn upstream: spawn npx ENOENT
```

Same class of failure for any `.cmd` shim used as the upstream (e.g. `gemini.cmd`).

## Root cause

`child_process.spawn('npx', ...)` without `shell: true` cannot resolve `.cmd` shims on Windows. PATHEXT is only consulted when spawning through a shell. POSIX systems resolve fine without a shell.

## Fix

- Extract spawn options into `src/mcp-servers/caveman-shrink/spawn-options.js` (`getSpawnOptions(platform)`) so the behavior is unit-testable without reentering the CLI entry point.
- Set `shell: process.platform === 'win32'`. POSIX unchanged (no quoting surprises).
- Add `windowsHide: true` to avoid a flash of console window.

## Diff size

4 lines of behavior change; 3 files (one new tiny module, one new test, one require swap).

## Tests

New `tests/test_mcp_shrink_spawn_options.js` (mirrors `tests/test_mcp_shrink.js` style — Node `assert`, plain `node tests/<file>.js` invocation):

- `win32` → `shell:true`, `windowsHide:true`
- `linux` / `darwin` → `shell:false`
- default arg uses `process.platform`

```
$ node tests/test_mcp_shrink_spawn_options.js
mcp-shrink spawn-options tests

  ✓ win32 enables shell so npx and .cmd shims resolve
  ✓ linux keeps shell off to avoid argv quoting surprises
  ✓ darwin keeps shell off
  ✓ defaults to current platform when no arg passed

4 passed, 0 failed
```

Existing `tests/test_mcp_shrink.js` still passes (12/12).

## Verified locally

Windows 11, wrapping a Node-based MCP server end-to-end. Compression debug output intact: e.g. `[caveman-shrink] tools.run_command.description: 98→92 bytes`. No regression on Linux smoke either (`shell:false` keeps existing behavior).

## Notes for the maintainer

- POSIX path is unchanged on purpose — `shell: true` there would re-introduce argv quoting risks.
- Args still come from `.mcp.json`; trust boundary is the same as before. `shell: true` does not widen it because `npx` and `.cmd` shims are the normal way Windows users invoke Node tools; users could already inject by editing `.mcp.json`.

### DIFF
--- a/src/mcp-servers/caveman-shrink/index.js
+++ b/src/mcp-servers/caveman-shrink/index.js
@@ -40,9 +40,9 @@ const debug = process.env.CAVEMAN_SHRINK_DEBUG === '1';
 const fields = (process.env.CAVEMAN_SHRINK_FIELDS || 'description')
   .split(',').map(s => s.trim()).filter(Boolean);
 
-const upstream = spawn(args[0], args.slice(1), {
-  stdio: ['pipe', 'pipe', 'inherit'],
-});
+const { getSpawnOptions } = require('./spawn-options');
+
+const upstream = spawn(args[0], args.slice(1), getSpawnOptions());
 
 upstream.on('error', err => {
   process.stderr.write(`caveman-shrink: failed to spawn upstream: ${err.message}\n`);
--- a/src/mcp-servers/caveman-shrink/spawn-options.js
+++ b/src/mcp-servers/caveman-shrink/spawn-options.js
@@ -1,0 +1,21 @@
+// Spawn options for the upstream MCP child process.
+//
+// Windows: spawn('npx', ...) (and any .cmd shim such as 'gemini') hits ENOENT
+// because PATHEXT resolution only happens when child_process spawns through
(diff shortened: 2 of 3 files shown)


gaelian:
+1 for this fix.

JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`6ce47d44`](https://github.com/JuliusBrussee/caveman/commit/6ce47d44) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #393, opened 2026-05-14
fix(install): skip standalone hook wiring when plugin manifest already handles it

### LINKED ISSUE #392: SessionStart + UserPromptSubmit hooks fire twice when installer wires both plugin and standalone hooks
## What happens

`bin/install.js` `installClaude()` registers the SessionStart + UserPromptSubmit hooks **twice** in a single Claude Code install:

1. `claude plugin install caveman@caveman` registers the hooks via the plugin manifest (`plugins/caveman/.claude-plugin/plugin.json`'s `hooks` block).
2. `installHooks(ctx)` then writes the same `SessionStart` + `UserPromptSubmit` entries directly into `$CLAUDE_CONFIG_DIR/settings.json`, pointing at `~/.claude/hooks/caveman-activate.js` and `~/.claude/hooks/caveman-mode-tracker.js`.

`SETTINGS.hasCavemanHook()` (`bin/lib/settings.js:127`) is the idempotency probe but only inspects `settings.hooks[event]` — plugin-manifest registrations live outside `settings.json`, so they're invisible to the probe.

Result: every session, two `CAVEMAN MODE ACTIVE` SessionStart blocks land in the model's context, and every user message receives two `UserPromptSubmit hook additional context: CAVEMAN MODE ACTIVE (full). …` lines.

## Expected

After running the unified installer, caveman's SessionStart and UserPromptSubmit hooks each fire **once** per event.

## Before / after evidence

In a session start, the model receives this (only the structure shown; trimmed):

```
<system-reminder>
SessionStart:resume hook success: CAVEMAN MODE ACTIVE — level: full
Respond terse like smart caveman. […]
## Persistence
ACTIVE EVERY RESPONSE. […]
## Rules
[…]
## Boundaries
[…]
</system-reminder>

<system-reminder>
SessionStart:resume hook success: CAVEMAN MODE ACTIVE — level: full
Respond terse like smart caveman. […]
## Persistence
ACTIVE EVERY RESPONSE once triggered. […]
## Rules
[…]
### Examples
[…]
## Auto-Clarity Exception
[…]
</system-reminder>
```

The first block is the **fallback** ruleset emitted by `~/.claude/hooks/caveman-activate.js` (no SKILL.md adjacent in `$CLAUDE_CONFIG_DIR/skills/caveman/`). The second is the **full SKILL.md-driven** ruleset emitted by `${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js` (where SKILL.md *is* adjacent).

Same on UserPromptSubmit — the `additionalContext` reinforcement line is injected twice per turn.

## Root cause

`bin/install.js:382-409` — `installClaude()`:

```js
async function installClaude(ctx) {
  // …
  // (1) plugin path
  const r1 = runSpawn('claude', ['plugin', 'marketplace', 'add', REPO], …);
  const r2 = runSpawn('claude', ['plugin', 'install', 'caveman@caveman'], …);
  // …

  if (opts.withHooks) {           // ← defaults to true
    say('  → installing hooks (--with-hooks)');
    const r = await installHooks(ctx);   // (2) standalone path — duplicates (1)
    // …
  }
  // …
}
```

`opts.withHooks` defaults to `true` (`bin/install.js:60-61`). The README also documents `--with-hooks / --no-hooks` as on by default. So unless a user explicitly runs `--no-hooks`, both registrations land.

The plugin manifest (`.claude-plugin/plugin.json`) already wires the same two hooks via `${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-*.js`, with the same `timeout` and `statusMessage`. The plugin version is the richer one — it can read SKILL.md from `${CLAUDE_PLUGIN_ROOT}/skills/caveman/SKILL.md`, while the standalone version at `$CLAUDE_CONFIG_DIR/hooks/` cannot (no SKILL.md adjacent), so it falls back to the smaller in-code ruleset.

## Repro

1. Fresh `~/.claude`, `CLAUDE_CONFIG_DIR` unset.
2. Run the documented one-liner: `curl -fsSL https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.sh | bash`.
3. Open Claude Code, start a session.
4. Inspect `~/.claude/settings.json` — both an entry pointing at `~/.claude/hooks/caveman-activate.js` under `SessionStart` and one under `UserPromptSubmit` exist; `enabledPlugins` also contains `caveman@caveman`.
5. Trigger any prompt — context window shows two `CAVEMAN MODE ACTIVE` blocks and two UserPromptSubmit reinforcement lines.

Also reproducible by running the per-agent path: `claude plugin marketplace add JuliusBrussee/caveman && claude plugin install caveman@caveman` followed by `node bin/install.js --only claude` (which is the second half of what the one-liner does).

## Suggested fix

When `claude plugin install caveman@caveman` succeeds, skip `installHooks(ctx)` — the plugin manifest already wires SessionStart + UserPromptSubmit. Keep `installHooks()` reachable behind `--with-hooks` for users on hostless Claude Code installs or environments where the plugin path is unavailable, but make the default branch *plugin-only* when the plugin install succeeded.

Alternative narrower fix: extend `hasCavemanHook()` to also probe enabled plugin manifests (read `enabledPlugins` and walk `${CLAUDE_PLUGIN_ROOT}/.claude-plugin/plugin.json` for matching event registrations), so `addCommandHook()` becomes idempotent across both registration paths.

I will follow up with a PR. Reporting first so we can agree on the shape.

## Local mitigation

For users hitting this today: back up `~/.claude/settings.json`, then delete the `SessionStart` entry whose command references `~/.claude/hooks/caveman-activate.js` and the `UserPromptSubmit` entry whose command references `~/.claude/hooks/caveman-mode-tracker.js`. Keep `"caveman@caveman": true` in `enabledPlugins` — the plugin manifest path keeps emitting the richer SKILL.md-driven ruleset on its own.

## Platform

- [x] Claude Code
- [ ] Codex
- [ ] Other

## Version

- OS: macOS 15.x (Darwin 25.3.0, Apple Silicon)
- Plugin commit: `63a91ecadbf4` (`~/.claude/plugins/cache/caveman/caveman/63a91ecadbf4`)
- Node: 25.8.1 (Homebrew)
- Shell: zsh
- Install path: `claude plugin install caveman@caveman` + `bin/install.js --with-hooks` (the default)

ShunmeiCho:
## Summary

Fixes [#392](https://github.com/JuliusBrussee/caveman/issues/392) — `SessionStart` and `UserPromptSubmit` hooks fire twice on every Claude Code session because `installClaude()` registers them once through the plugin manifest *and* once through `installHooks()` writing into `~/.claude/settings.json`.

`SETTINGS.hasCavemanHook()` only inspects `settings.hooks[event]`; plugin-manifest registrations live outside `settings.json`, so the existing idempotency probe can't catch the duplicate.

## What changed

- `parseArgs()`: stop force-flipping `opts.withHooks` to `true` for both `--all` and the default branch. `'auto'` now propagates into `installClaude()`.
- `installClaude()`: track whether the plugin install actually landed (`pluginInstallSucceeded`). Then make the wiring decision explicit:

| `opts.withHooks` | Behavior |
|---|---|
| `false` (`--no-hooks`, `--minimal`) | Skip standalone wiring (unchanged). |
| `true` (`--with-hooks`) | Wire standalone. If the plugin also wired hooks, print a warning that both will fire. |
| `'auto'` (default, `--all`) | Wire standalone *only* when the plugin install path didn't succeed. |

Result: the default install no longer produces duplicate registrations, and users who genuinely want the belt-and-suspenders setup can still pass `--with-hooks` explicitly (with a clear warning).

## Why not a narrower fix?

Extending `hasCavemanHook()` to probe `enabledPlugins` + walk the plugin manifest was the other option, but it's brittle — it has to know the plugin's install path layout, parse `enabledPlugins` from `settings.json` after Claude Code may have mutated it, and stay in sync if the plugin manifest's event list ever expands. Making `installClaude()` aware of *its own* plugin-install result is the simpler invariant.

## Test plan

- [x] `node -c bin/install.js` — syntax OK
- [x] Manual run on macOS with `caveman@caveman` already installed: installer now prints `hooks: plugin manifest handles SessionStart + UserPromptSubmit` and leaves `settings.json` untouched.
- [x] Verified parsed shape of `settings.json` after a manual cleanup matches what the patched installer would produce on a fresh install.
- [ ] Fresh-install regression: run `bin/install.js --only claude` against a clean `~/.claude/`, confirm one `CAVEMAN MODE ACTIVE` SessionStart block and one `UserPromptSubmit additionalContext` per turn. (Pending — needs a clean profile; happy to take a maintainer's word that this matches the test infra you already have.)
- [ ] `bin/install.js --only claude --with-hooks` — confirm warning prints, both paths wired (intentional opt-in).
- [ ] `bin/install.js --only claude --no-hooks` — confirm only plugin path wired.
- [ ] `bin/install.js --all` — confirm hooks wired exactly once via plugin manifest.

## Notes for reviewers

- The decision matrix is documented inline above the new `if/else` so future readers don't have to reverse-engineer the intent.
- `--all` now subtly changes meaning: it no longer adds standalone hooks on top of the plugin's hooks. I think this is correct — the README copy "Plugin + hooks + statusline + MCP shrink + per-repo rule files. The full ride." reads naturally as "everything wired once", not "everything wired twice". Happy to revert that piece if you'd rather keep `--all` as a literal force-on for every knob.
- I did not touch `INSTALL.md`. If the maintainer wants the README to call this out, I can amend.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -90,9 +90,14 @@ function parseArgs(argv) {
     }
   }
   if (opts.all && opts.minimal) die('error: --all and --minimal are mutually exclusive');
-  if (opts.all) { opts.withHooks = true; opts.withInit = true; opts.withMcpShrink = true; }
+  // --all / --minimal set the explicit knobs they imply. We deliberately leave
+  // opts.withHooks at 'auto' for --all so installClaude() can detect when the
+  // Claude Code plugin manifest already wires SessionStart + UserPromptSubmit
+  // and skip the standalone settings.json wiring — duplicate registration
+  // makes both hooks fire twice per event (issue #392). Users who want the
+  // standalone wiring on top of the plugin can still pass --with-hooks.
+  if (opts.all) { opts.withInit = true; opts.withMcpShrink = true; }
   if (opts.minimal) { opts.withHooks = false; opts.withInit = false; opts.withMcpShrink = false; }
-  if (opts.withHooks === 'auto') opts.withHooks = true;
   if (opts.withMcpShrink === 'auto') opts.withMcpShrink = true;
   // Validate --only ids against the provider matrix. PROVIDERS is defined later
   // in the file but is in scope by the time this function runs.
@@ -390,18 +395,55 @@ async function installClaude(ctx) {
     const r = captureSpawn('claude', ['plugin', 'list']);
     if (r.status === 0 && /caveman/i.test(r.stdout || '')) alreadyInstalled = true;
   }
+  let pluginInstallSucceeded = false;
   if (alreadyInstalled) {
     note('  caveman plugin already installed (use --force to reinstall)');
     results.skipped.push(['claude', 'plugin already installed']);
+    pluginInstallSucceeded = true;
   } else {
     const r1 = runSpawn('claude', ['plugin', 'marketplace', 'add', REPO], null, opts.dryRun);
     const r2 = runSpawn('claude', ['plugin', 'install', 'caveman@caveman'], null, opts.dryRun);
-    if ((r1.status || 0) === 0 && (r2.status || 0) === 0) results.installed.push('claude');
-    else results.failed.push(['claude', 'claude plugin install failed']);
+    if ((r1.status || 0) === 0 && (r2.status || 0) === 0) {
+      results.installed.push('claude');
+      pluginInstallSucceeded = true;
+    } else {
+      results.failed.push(['claude', 'claude plugin install failed']);
+    }
   }
 
-  if (opts.withHooks) {
-    say('  → installing hooks (--with-hooks)');
+  // Hook wiring decision matrix:
+  //   --no-hooks            → opts.withHooks === false           → skip
+  //   --with-hooks          → opts.withHooks === true            → wire (warn if duplicate)
+  //   default / --all       → opts.withHooks === 'auto'          → wire only if plugin install failed
+  //
+  // The plugin manifest at plugins/caveman/.claude-plugin/plugin.json already
+  // wires SessionStart + UserPromptSubmit when the plugin install succeeds.
+  // Wiring the same hooks again into settings.json makes both fire every event
+  // (issue #392 — two CAVEMAN MODE ACTIVE blocks per session start, two
+  // UserPromptSubmit reinforcement lines per turn). Skip the duplicate.
+  let shouldWireHooks;
+  if (opts.withHooks === false) {
+    shouldWireHooks = false;
+  } else if (opts.withHooks === true) {
+    shouldWireHooks = true;
+    if (pluginInstallSucceeded) {
+      warn('  --with-hooks wires hooks in settings.json alongside the plugin manifest.');
+      warn('  Both will fire on every event. Pass --no-hooks to keep only the plugin path.');
+    }
+  } else {
+    // 'auto'
+    shouldWireHooks = !pluginInstallSucceeded;
+    if (!shouldWireHooks) {
+      note('  hooks: plugin manifest handles SessionStart + UserPromptSubmit');
+      note('  (pass --with-hooks to also wire standalone hooks in settings.json)');
+      results.skipped.push(['claude-hooks', 'plugin manifest handles hooks']);
+    } else {
+      note('  hooks: plugin install did not succeed; falling back to standalone wiring');
+    }
+  }
+
+  if (shouldWireHooks) {
+    say('  → installing hooks');
     const r = await installHooks(ctx);
     if (r === 'ok') results.installed.push('claude-hooks');
     else if (r === 'skip') results.skipped.push(['claude-hooks', 'already wired']);


praxstack:
This correctly prevents the **double-wiring** going forward. Flagging an adjacent gap it doesn't cover (filed as #471): users who migrated from the old manual hook install to the plugin **before** this lands already have `settings.json` entries pointing at `~/.claude/hooks/caveman-activate.js` / `caveman-mode-tracker.js` that the migration disabled (`*.js.disabled`). Those orphaned entries crash every session with `node:internal/modules/cjs/loader:1478 — Cannot find module …caveman-activate.js`.

`installClaude()` here only decides whether to *add* standalone wiring — it never removes the already-present stale entries, so it won't self-heal a dirty `settings.json`. A reconciliation pass (drop managed hook entries whose target file is missing — note `rewriteLegacyManagedHookCommands` skips absolute-node commands, which is the exact shape these orphans take) would close #471 and pairs cleanly with this PR. Details + repro in #471.

JuliusBrussee:

Issue #370, opened 2026-05-12
One-line installer doesn't work on MacOS.


coredumperror:
**What happen**
Running the one-line installer on MacOS doesn't actually install anything, despite claiming to do so. 

**Example**
Running this:
```
curl -fsSL https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.sh | bash
```

Gives me this output:
```
🪨 caveman installer
  JuliusBrussee/caveman

→ Codex CLI detected
  $ npx -y skills add JuliusBrussee/caveman -a codex

███████╗██╗  ██╗██╗██╗     ██╗     ███████╗
██╔════╝██║ ██╔╝██║██║     ██║     ██╔════╝
███████╗█████╔╝ ██║██║     ██║     ███████╗
╚════██║██╔═██╗ ██║██║     ██║     ╚════██║
███████║██║  ██╗██║███████╗███████╗███████║
╚══════╝╚═╝  ╚═╝╚═╝╚══════╝╚══════╝╚══════╝

┌   skills
│
│  Tip: use the --yes (-y) and --global (-g) flags to install without prompts.
│
◇  Source: https://github.com/JuliusBrussee/caveman.git
│
◇  Repository cloned
│
◇  Found 7 skills
│
◆  Select skills to install (space to toggle)
│  ◻ cavecrew (Decision guide for delegating to caveman-style subagents....)
│  ◻ caveman
│  ◻ caveman-commit
│  ◻ caveman-compress
│  ◻ caveman-help
│  ◻ caveman-review
│  ◻ caveman-stats
└

→ Cursor detected
  $ npx -y skills add JuliusBrussee/caveman -a cursor

███████╗██╗  ██╗██╗██╗     ██╗     ███████╗
██╔════╝██║ ██╔╝██║██║     ██║     ██╔════╝
███████╗█████╔╝ ██║██║     ██║     ███████╗
╚════██║██╔═██╗ ██║██║     ██║     ╚════██║
███████║██║  ██╗██║███████╗███████╗███████║
╚══════╝╚═╝  ╚═╝╚═╝╚══════╝╚══════╝╚══════╝

┌   skills
│
│  Tip: use the --yes (-y) and --global (-g) flags to install without prompts.
│
◇  Source: https://github.com/JuliusBrussee/caveman.git
│
◇  Repository cloned
│
◇  Found 7 skills
│
◆  Select skills to install (space to toggle)
[...]

PR #395, opened 2026-05-15
fix(install): use --skill '*' instead of --all in per-provider path (fixes #389)

### LINKED ISSUE #389: Installer passes skills --all with -a, causing installs to all agents
## What happened

Running the one-line installer on a machine with multiple detected agents now invokes the `skills` CLI like this for each detected profile:

```text
npx -y skills add JuliusBrussee/caveman -a codex --yes --all
npx -y skills add JuliusBrussee/caveman -a amp --yes --all
npx -y skills add JuliusBrussee/caveman -a droid --yes --all
npx -y skills add JuliusBrussee/caveman -a goose --yes --all
```

But the `skills` CLI interprets `--all` as "all skills to all agents", not "all skills to the agent selected by `-a`". The output shows that each per-agent install actually targets every supported agent:

```text
●  Installing all 7 skills
●  Installing to all 55 agents
```

That means the installer repeatedly writes the same project `.agents/skills/*` payload through many agent adapters, even though each call appears to target one agent.

## Expected

For each detected provider, install all caveman skills only to that provider.

The documented `skills` CLI form for "all skills to specific agents" is:

```text
npx skills add JuliusBrussee/caveman --skill '*' -a codex -y
```

So the installer likely wants something like:

```js
['-y', 'skills', 'add', REPO, '--skill', '*', '-a', prov.profile, '--yes']
```

instead of:

```js
['-y', 'skills', 'add', REPO, '-a', prov.profile, '--yes', '--all']
```

The auto-detect fallback at the bottom of `bin/install.js` may still intentionally want true all-agents behavior; this issue is specifically about the per-provider `installViaSkills()` path where `-a prov.profile` is already present.

## Evidence checked

- Current `main` / latest release tag `v1.8.2` points at commit `63a91ec`.
- `bin/install.js` in `v1.8.2` changed `installViaSkills()` to pass `--yes --all`:

```js
const args = ['-y', 'skills', 'add', REPO, '-a', prov.profile, '--yes', '--all'];
```

- Re-running the installer produced `Installing to all 55 agents` under each per-agent call.
- Context7 docs for `/vercel-labs/skills` say:
  - `--all` installs all skills from the source to all agents.
  - all skills to one specific agent should use `--skill '*' -a <agent>`.

## Related but separate

This is distinct from the OpenCode `caveman-compress.md` ENOENT tracked in #367 / #375. That one is a missing command file / OpenCode installer manifest issue. This issue is about generic `npx skills add` invocation semantics for Codex/Amp/Droid/Goose/etc.

## Version / install method

```text
```

Resolved to latest release `v1.8.2` / `63a91ec` at time of testing.

code-with-rashid:
## Summary

Fixes #389 — `installViaSkills()` in `bin/install.js` was passing `--all` for every per-provider call, which the upstream `skills` CLI interprets as *all* skills to *all* agents (ignoring the `-a <profile>` selection). So every detected agent triggered a fan-out across all 55 supported agent adapters; the same caveman skills got re-applied dozens of times.

Switched to the documented `--skill '*' -a <agent>` form, which is what "all skills, one specific agent" actually means in upstream `skills`. `--yes` stays (suppresses the upstream TUI when `curl|bash` has no TTY — see comment in code and issue #370).

## Changes

- `bin/install.js` `installViaSkills()`:
  ```diff
  - const args = ['-y', 'skills', 'add', REPO, '-a', prov.profile, '--yes', '--all'];
  + const args = ['-y', 'skills', 'add', REPO, '--skill', '*', '-a', prov.profile, '--yes'];
  ```
- Updated the surrounding comment to explain *why* `--skill '*'` and not `--all`, citing the #389 root cause so future maintainers don't undo it.

## Verification

- `node --check bin/install.js` — clean.
- `tests/installer/unit.argv.test.mjs` — 13/13 pass.
- `tests/installer/e2e.dryrun.test.mjs` — 2/2 pass.
- Confirmed only one `--all` occurrence in an `npx skills add` argv vector exists in the file; the `--all` in argv parsing at L65 is the installer's own flag and unrelated.

## Out of scope

The auto-detect fallback path that genuinely wants all-agents behavior (the issue body notes this as "distinct from the per-provider call") is unchanged — it doesn't use `installViaSkills()`.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -443,13 +443,17 @@ function installViaSkills(ctx, prov) {
   const { say, opts, results } = ctx;
   results.detected++;
   say(`→ ${prov.label} detected`);
-  // --yes --all: skip the upstream skill-selection TUI and confirmation prompts.
-  // Without these, `curl|bash` (no TTY on stdin) renders an empty checkbox list
-  // the user can't interact with, then exits 0 with zero skills installed —
-  // and our installer happily reports success. See issue #370.
-  // We've already decided which agent to install for via auto-detect / --only;
-  // making the user re-select 7 skills inside skills CLI would be redundant.
-  const args = ['-y', 'skills', 'add', REPO, '-a', prov.profile, '--yes', '--all'];
+  // --skill '*' --yes: skip the upstream skill-selection TUI and confirmation
+  // prompts. Without --skill, `curl|bash` (no TTY on stdin) renders an empty
+  // checkbox list the user can't interact with, then exits 0 with zero skills
+  // installed — and our installer happily reports success. See issue #370.
+  //
+  // We pass `--skill '*'` rather than `--all` because the upstream `skills` CLI
+  // interprets `--all` as "all skills from the source to *all* agents", which
+  // ignores the `-a prov.profile` selection and writes every skill through
+  // every agent adapter (see issue #389). `--skill '*' -a <agent>` is the
+  // documented form for "install every skill into a specific agent".
+  const args = ['-y', 'skills', 'add', REPO, '--skill', '*', '-a', prov.profile, '--yes'];
   const r = runSpawn('npx', args, null, opts.dryRun);
   if ((r.status || 0) === 0) results.installed.push(prov.id);
   else results.failed.push([prov.id, `npx skills add (${prov.profile}) failed`]);


JuliusBrussee:

Issue #366, opened 2026-05-10
PowerShell 7 installer fails: `Cannot overwrite variable Args because the variable has been optimized`


FahimFBA:
**What happen**

When running:

```
irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 | iex
```

on PowerShell 7+, the installer fails with:

```
Invoke-Expression: Cannot overwrite variable Args because the variable has been optimized.
```

The cause might be:

`install.ps1` defines a parameter named `$Args`:

```
param(
  [Parameter(ValueFromRemainingArguments = $true)]
  [string[]]$Args
)
```

`$Args` is a built-in automatic PowerShell variable and cannot be overridden in newer PowerShell versions.

Renaming it to something else (e.g. `$InstallerArgs`) fixes the issue.

Affected lines also need updating:

```
& node $local @InstallerArgs
& npx -y "github:$Repo" @InstallerArgs
```

**Expected**
It should have found all available agents and provide prompt to install.

**Platform**
- [x] Claude Code
- [x] Codex
- [x] Other: Kiro

**Version / install method**
The CLI given in the README

Bilal-io:
I am seeing the same issue. 
~~Is there a way to install Caveman for windows without this command?~~
I ran `npx skills add JuliusBrussee/caveman -a pi`

TheLordDrake:
Running into this issue as well. Workaround suggested by @Bilal-io does not seem to work. The script says it's successful, but Claude doesn't seem to recognize the skill.

vallamost:
Here's another workaround that worked for me:

```
$tmp = Join-Path $env:TEMP "caveman-install.ps1"; irm
https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 -OutFile $tmp; & $tmp
```

stressatoo:
Can reproduce.

> Here's another workaround that worked for me:
> 
> ```
> $tmp = Join-Path $env:TEMP "caveman-install.ps1"; irm
> https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 -OutFile $tmp; & $tmp
> ```

This workaround doesn't work for me, it asks for an uri.

BuiNhienLoc:
> Can reproduce.
> 
> > Here's another workaround that worked for me:
> > ```
> > $tmp = Join-Path $env:TEMP "caveman-install.ps1"; irm
[...]

PR #399, opened 2026-05-16
fix(install): avoid reserved Args parameter in PowerShell

xnnnc:
## Summary
- Rename the PowerShell installer's remaining-arguments parameter from `Args` to `ScriptArgs`.
- Forward `ScriptArgs` to both local `bin/install.js` and `npx` paths.

## Problem
Running the Windows one-line installer currently fails under PowerShell:

```powershell
irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 | iex
```

Error:

```text
Cannot overwrite variable Args because the variable has been optimized.
```

PowerShell's automatic `$Args` variable can be optimized/read-only in this pipeline execution path, so declaring a script parameter named `$Args` causes `Invoke-Expression` to fail before the installer can run.

## Validation
- Parsed `install.ps1` with PowerShell's parser successfully.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -17,7 +17,7 @@
 [CmdletBinding()]
 param(
   [Parameter(ValueFromRemainingArguments = $true)]
-  [string[]]$Args
+  [string[]]$ScriptArgs
 )
 
 $ErrorActionPreference = "Stop"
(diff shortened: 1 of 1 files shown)


spicyarch:
This would fix #366

zyhe16 (approved):
LGTM, I have the same issue for Windows installation and I believe that this would fix it. Hope it can get merged soon

JuliusBrussee:
Thanks @xnnnc — the PowerShell shim no longer reuses reserved `$Args`. Closing as implemented.

PR #401, opened 2026-05-16
fix: address code review issues in hook infrastructure

GKV22:
## Summary

Five targeted fixes from a code review of the hook infrastructure:

- **Export `INDEPENDENT_MODES` from `caveman-config.js`** — `caveman-activate.js` and `caveman-mode-tracker.js` each maintained a local `new Set(['commit', 'review', 'compress'])`. Adding a new independent mode required updating two files and was easy to miss. Now one definition, exported alongside `VALID_MODES`.

- **Remove phantom `'wenyan-full'` from `VALID_MODES`** — `'wenyan-full'` was never written to the flag file (the mode-tracker aliases `/caveman wenyan-full` → stored as `'wenyan'`). Having it in the whitelist meant `readFlag` would accept it, but no writer would ever produce it — a ghost entry that could cause confusion on a direct equality check. Removed; the special-case alias in `caveman-mode-tracker.js` continues to handle the `/caveman wenyan-full` command.

- **Cap `readHistory` at 10 MB** — `readFlag` already enforces `MAX_FLAG_BYTES = 64`. `readHistory` was the only uncapped reader; an unusually large history file could cause OOM in `caveman-stats.js`. Now reads at most 10 MB; lines beyond that are silently truncated (same graceful-fail pattern as the rest of the module).

- **Guard `findRecentSession` DFS at 10 000 entries** — the stack-based traversal of `~/.claude/projects/` had no depth or count limit. Added a `visited` counter capped at `MAX_ENTRIES = 10_000`; bails out and returns the best match found so far.

- **Remove `settings.json.bak` after successful `install.sh` merge** — the backup was created before the Node merge script ran but never cleaned up. Repeated `--force` reinstalls accumulated stale backup files. Now deleted on success; if the merge fails, the script exits before the `rm` and the backup is preserved for recovery.

## Test plan

- [ ] `node src/hooks/caveman-activate.js` exits cleanly (SessionStart hook smoke test)
- [ ] `echo '{"prompt":"/caveman ultra"}' | node src/hooks/caveman-mode-tracker.js` writes `ultra` to flag
- [ ] `echo '{"prompt":"/caveman wenyan-full"}' | node src/hooks/caveman-mode-tracker.js` writes `wenyan` to flag (alias preserved)
- [ ] `node src/hooks/caveman-stats.js` runs without error
- [ ] Existing test suite: `node tests/test_hooks.py` / `node tests/test_symlink_flag.js`

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -9,7 +9,7 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
-const { getDefaultMode, safeWriteFlag } = require('./caveman-config');
+const { getDefaultMode, safeWriteFlag, INDEPENDENT_MODES } = require('./caveman-config');
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
 const flagPath = path.join(claudeDir, '.caveman-active');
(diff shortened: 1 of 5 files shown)


JuliusBrussee:
Thanks @GKV22 — the reviewed hook failure paths are covered by later shared-parser and fail-open hardening. Closing this old-base patch.

PR #407, opened 2026-05-19
feat: project-scope gating for the SessionStart hook

mhd-ghaith-abtah:
## What

Adds opt-in/opt-out gating on top of the existing global mode so the SessionStart hook can be silenced (or, with an allowlist, only fired) in specific directories.

## Why

Today the only way to keep caveman off in a project you don't want it in is to set the global default to `off` — which also kills caveman in the projects you DO want it in. Several downstream consumers (notably [Flow](https://github.com/mhd-ghaith-abtah/flow), which orchestrates per-story workflows around BMad + ECC + caveman) hit this and have to document a manual workaround. Native project-scope detection is much cleaner.

## How

Resolution order (first match wins):

1. `CWD/.caveman-disable` file → `disabled`
2. `CWD/.caveman-enable` file → `enabled`
3. `CAVEMAN_PROJECT_SCOPE` env var (`disabled` | `enabled`) → that value
4. `config.projectScope.deny[]` contains a path prefix of CWD → `disabled`
5. `config.projectScope.allow[]` is set AND no entry matches CWD → `disabled` (allowlist mode)
6. `config.projectScope.allow[]` contains a path prefix of CWD → `enabled`
7. Otherwise → `inherit` (use the existing global default — current behavior unchanged)

When `disabled`, the activate hook exits with `OK (project-scope disabled)` — no flag file, no ruleset emitted. Same end-state as `mode === 'off'`, scoped to one CWD.

Path prefix matching resolves symlinks via `fs.realpathSync()` so macOS `/tmp` ↔ `/private/tmp` setups work cleanly. Non-existent allow-entries fall back to `path.resolve()` so users can stage config before the directory exists.

## Compatibility

**Default behavior is unchanged for existing users.** Without marker files, env var, or config entries, activation proceeds exactly as before. The new logic only runs when explicitly opted into.

`getProjectScope()` is exported from `caveman-config.js` in case other hooks (e.g., `caveman-mode-tracker.js`) want to honor the same gating in the future — not wired in here to keep the diff minimal.

## Tests

- 7 new cases in `tests/test_project_scope.py` covering all six resolution branches plus the default-on case.
- Full suite: 26 tests, all green (19 existing + 7 new).

```
$ python3 -m unittest discover tests
Ran 26 tests in 0.512s
OK
```

## Example config

```json
{
  "defaultMode": "full",
  "projectScope": {
    "deny": ["/Users/me/work/client-xyz"],
    "allow": ["/Users/me/code/flow", "/Users/me/code/api"]
  }
}
```

Marker files (zero-byte sentinels in the project root) are usually enough; the config-driven lists are for users who want centralized control across many repos.

## Out of scope (intentional)

- Walking up parent directories for marker files — kept to CWD-only for cheap detection and explicit semantics.
- Honoring scope in `caveman-mode-tracker.js` (the UserPromptSubmit hook). Could be a follow-up if there's appetite; for this PR I wanted the diff to be focused on the SessionStart entrypoint.

Happy to revise scope (rename markers, drop env var, drop allowlist support, etc.) — wanted to put the full design forward to start the conversation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,42 @@ A March 2026 paper ["Brevity Constraints Reverse Performance Hierarchies in Lang
 
 Maintainer detail (hook architecture, file ownership, CI sync) live in [CLAUDE.md](./CLAUDE.md).
 
+## Project-Scope Gating
+
+Sometimes caveman activate global. User want caveman only some project. Add file or config — caveman skip activation that CWD only. Global default not change.
+
+**Marker files (cheapest path)** — drop zero-byte file in project root:
+
+```bash
+# In project where you want caveman OFF:
+touch .caveman-disable
+
+# Or, with global default=off, opt one project IN:
+touch .caveman-enable
+```
+
+**Env var (single session)**:
+
+```bash
+CAVEMAN_PROJECT_SCOPE=disabled claude   # this session only
+```
+
+**Config-driven allow / deny lists** (`~/.config/caveman/config.json`):
+
+```json
+{
+  "defaultMode": "full",
+  "projectScope": {
+    "deny":  ["/Users/me/work/client-xyz"],
+    "allow": ["/Users/me/code/flow", "/Users/me/code/api"]
+  }
+}
+```
+
+Resolution order (first match wins): `.caveman-disable` marker → `.caveman-enable` marker → `CAVEMAN_PROJECT_SCOPE` env → config `deny[]` prefix match → config `allow[]` prefix match (when `allow` set, anything not in it is denied) → global default. Path prefixes resolve symlinks so `/tmp` ↔ `/private/tmp` style setups Just Work.
+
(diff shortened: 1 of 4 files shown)


mhd-ghaith-abtah:
Pushed a follow-up commit ([5602309](https://github.com/JuliusBrussee/caveman/commit/5602309)) to fix a logic-order bug in the original patch: the global `mode === 'off'` branch was firing before the project-scope check, which made `.caveman-enable` useless in the canonical allowlist-mode workflow (global default `off` + per-project enable markers).

Reordered so project-scope is checked first; when scope is `enabled` and global mode is `off`, mode is promoted to `full` (the user explicitly opted in, needs some active mode).

Two new test cases cover both directions:
- `test_global_off_plus_enable_marker_activates` — the fix
- `test_global_off_without_marker_stays_silent` — counterpart

Full suite now 9/9 green (was 7).

PR #416, opened 2026-05-20
feat: add Amazon Q and Kiro IDE support

Jeff-Meendering:


JuliusBrussee:
Thanks @Jeff-Meendering — Kiro already landed through the shared provider path, while Amazon Q needs its own verified adapter. Closing this mixed superseded PR.

PR #419, opened 2026-05-20
fix(opencode): use correct plugin hooks (chat.message + system.transform)

### LINKED ISSUE #418: bug(opencode): plugin hooks session.created and tui.prompt.append never fire — wrong API
### Summary

The opencode plugin (`src/plugins/opencode/plugin.js`) registers `session.created` and `tui.prompt.append` as named hook keys, but opencode's plugin API does not support these as trigger-able hooks. The flag file is never written and the per-turn reinforcement never fires.

### Root cause

opencode's plugin system (`@opencode-ai/plugin`) defines a `Hooks` interface with specific named hooks that can be triggered via `plugin.trigger(name, input, output)`. The supported hooks are:

- `event` (catch-all event bus listener)
- `chat.message`
- `chat.params`
- `chat.headers`
- `command.execute.before`
- `tool.execute.before` / `tool.execute.after`
- `shell.env`
- `experimental.chat.system.transform`
- `experimental.chat.messages.transform`
- `experimental.session.compacting`
- `experimental.compaction.autocontinue`

`session.created` and `tui.prompt.append` are **not** in this list. They exist as bus events (for internal UI use) but are never dispatched to plugins via the `trigger()` mechanism.

Additionally, opencode's trigger pattern is `(input, output) => Promise<void>` with mutation on `output` — return values are discarded. The current plugin returns `{ append: ... }` which is silently ignored.

### Evidence

- Plugin loads successfully (confirmed in logs: `service=plugin path=file:///...plugin.js loading plugin`)
- `session.created` bus event is published but never dispatched to plugin hooks
- `tui.prompt.append` is a TUI input-field event (inserts text into prompt widget), not a pre-send hook
- Tested with opencode 1.15.4 on Linux

### Fix

The correct hooks to use:

1. **Session init**: Run in the plugin factory function (which executes once per session at load time)
2. **User prompt interception**: Use `chat.message` hook `(input, output) => Promise<void>` — `output.parts` contains text parts with the user's message
3. **System prompt injection**: Use `experimental.chat.system.transform` hook `(input, output) => Promise<void>` — push reinforcement line to `output.system` array

### Patch

```javascript
export const CavemanPlugin = async (_ctx) => {
  // Initialize flag at plugin load time (equivalent to session start).
  const mode = getDefaultMode();
  if (mode === 'off') {
    try { if (existsSync(flagPath)) unlinkSync(flagPath); } catch (e) {}
  } else {
    safeWriteFlag(flagPath, mode);
  }

  return {
    'chat.message': async (_input, output) => {
      if (!output || !output.parts) return;
      for (const part of output.parts) {
        if (part && part.type === 'text' && part.text) {
          const change = parseModeChange(part.text);
          if (change) applyModeChange(change);
        }
      }
    },

    'experimental.chat.system.transform': async (_input, output) => {
      const active = readFlag(flagPath);
      if (active && !INDEPENDENT_MODES.has(active)) {
        output.system.push(reinforcementLine(active));
      }
    },
  };
};
```

### Tested with

- opencode 1.15.4
- `opencode run` headless mode
- Verified: flag init ✓, `/caveman ultra` ✓, `stop caveman` ✓, system prompt injection ✓

### Platform

- [x] OpenCode
- [ ] Claude Code
- [ ] Codex
- [ ] Other

paolomainardi:
## Summary

The opencode plugin registers `session.created` and `tui.prompt.append` as named hook keys, but opencode's plugin API (`@opencode-ai/plugin` `Hooks` interface) does not support these as trigger-able hooks. They exist as internal bus events only — the plugin system never dispatches them to user plugins.

**Result**: the flag file is never written and per-turn reinforcement never fires. Caveman is effectively dead on opencode despite the plugin loading successfully.

## Root cause

opencode's `plugin.trigger(name, input, output)` only dispatches to hooks defined in the `Hooks` interface:
- `chat.message`, `chat.params`, `chat.headers`
- `tool.execute.before` / `tool.execute.after`
- `shell.env`
- `experimental.chat.system.transform`
- `experimental.session.compacting`
- etc.

`session.created` and `tui.prompt.append` are NOT in this list.

Additionally, the trigger pattern is `(input, output) => Promise<void>` with **mutation on output** — return values are discarded. The old code returned `{ append: ... }` which was silently ignored.

## Fix

| Old (broken) | New (working) |
|---|---|
| `session.created` hook | Plugin factory body (runs once at load) |
| `tui.prompt.append` returning `{ append }` | `chat.message` for mode detection + `experimental.chat.system.transform` for injection |

## Testing

- ✅ Tested with opencode 1.15.4 on Linux via `opencode run` headless mode
- ✅ Flag initialization on session start
- ✅ `/caveman ultra` mode switching
- ✅ `stop caveman` natural-language deactivation
- ✅ System prompt reinforcement injection
- ✅ Updated unit test passes

Closes #418

### DIFF
--- a/src/plugins/opencode/plugin.js
+++ b/src/plugins/opencode/plugin.js
@@ -1,22 +1,29 @@
 // caveman — opencode plugin
 //
-// Mirrors the Claude Code SessionStart + UserPromptSubmit hook pair using
-// opencode's lifecycle hook system. Bun ESM module; loads the existing
-// security-hardened helpers from caveman-config.js via createRequire so the
-// symlink-safe flag-write code lives in one place.
+// Provides dynamic caveman mode tracking for opencode:
+// - Writes the mode flag on plugin initialization (session start)
+// - Parses user messages for /caveman commands and natural-language toggles
+// - Injects per-turn reinforcement into the system prompt
 //
 // Layout once installed:
 //   ~/.config/opencode/plugins/caveman/
 //   ├── package.json
 //   ├── plugin.js              ← this file
-//   └── caveman-config.js      ← copied sibling of src/hooks/caveman-config.js
+//   └── caveman-config.cjs     ← copied sibling of src/hooks/caveman-config.js
 //
-// Always-on caveman ruleset is provided separately via
-// ~/.config/opencode/AGENTS.md (Tier-3 base) so this plugin only handles
-// dynamic state — flag writes, slash-command parsing, natural-language
-// activation, and per-prompt reinforcement. opencode's `session.created`
-// payload doesn't expose a documented system-prompt-injection return, so we
-// don't try to emit ruleset content here.
+// The always-on caveman ruleset is provided separately via
+// ~/.config/opencode/AGENTS.md (Tier-3 base). This plugin handles dynamic
+// state only: flag writes, slash-command parsing, natural-language
+// activation, and per-turn reinforcement.
+//
+// Hook mapping (opencode >=1.15.x):
+//   - Plugin factory: session initialization (runs once at load)
+//   - chat.message: intercept user prompts for mode changes
+//   - experimental.chat.system.transform: inject reinforcement per-turn
+//
+// Note: opencode does NOT support 'session.created' or 'tui.prompt.append'
+// as named plugin hooks. Those exist as internal bus events only. See:
+// https://github.com/JuliusBrussee/caveman/issues/418
 
 import { createRequire } from 'node:module';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`22f75e3d`](https://github.com/JuliusBrussee/caveman/commit/22f75e3d) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #424, opened 2026-05-21
caveman-compress: preserve YAML frontmatter

jtsport13:
## Problem

`caveman-compress` strips YAML frontmatter from markdown files despite the SKILL.md preserve-structure rules. Hit when running on Claude Code memory files (`~/.claude/projects/.../memory/*.md`) that all start with `---` blocks defining `name`, `description`, `type`, etc. The LLM rewrites or drops the frontmatter, breaking `[[wiki-link]]` references that key off the `name:` slug.

## Fix

Split frontmatter off before sending body to Claude, then prepend it back to the compressed output. The model never sees frontmatter, so it can't touch it.

- New regex `FRONTMATTER_REGEX` matches `---\n...\n---\n` at file start
- New helper `split_frontmatter()` returns `(frontmatter, body)` tuple
- `compress_file()` splits, compresses body only, reassembles
- Files without frontmatter pass through unchanged (regex returns `("", text)`)

## Testing

Tested across 18 memory files in production use (real Claude Code memory dir):

- Frontmatter preserved byte-exact in all cases (verified via diff against `.original.md` backups)
- Validation pipeline still passes (URL/path/heading/code-block checks)
- Compression savings on body: 4–12% range (unchanged from prior behavior)
- Backwards compatible: files without frontmatter unaffected

Sample logging output added when frontmatter detected:
```
Detected YAML frontmatter (218 chars) — preserving verbatim
```

## Why not just prompt-engineer Claude harder

Tried. The LLM strips frontmatter ~50%+ of the time despite explicit "Preserve ALL headings exactly" and "Preserve Structure: Frontmatter/YAML headers in markdown files" rules in `build_compress_prompt()`. Surgical removal is the only reliable fix.

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -16,6 +16,27 @@ OUTER_FENCE_REGEX = re.compile(
     r"\A\s*(`{3,}|~{3,})[^\n]*\n(.*)\n\1\s*\Z", re.DOTALL
 )
 
+# YAML frontmatter: starts at file start with --- on its own line, ends with --- on its own line.
+# Captures the entire block (including delimiters and trailing newline) and the body after.
+FRONTMATTER_REGEX = re.compile(
+    r"\A(---\r?\n.*?\r?\n---\r?\n)(.*)", re.DOTALL
+)
+
+
+def split_frontmatter(text: str) -> tuple[str, str]:
+    """Split YAML frontmatter from body. Returns (frontmatter, body).
+
+    Memory files (and many other markdown docs) start with a YAML frontmatter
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`e8eae0ff`](https://github.com/JuliusBrussee/caveman/commit/e8eae0ff) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #429, opened 2026-05-23
feat(config): repo-local .caveman/config.json (per-project default mode)

tehalexf:
## Motivation

Today the only ways to override caveman's default mode are:

1. `CAVEMAN_DEFAULT_MODE` env var
2. User-level `~/.config/caveman/config.json` (or XDG / APPDATA equivalent)

Both are *per-machine* / *per-user*. A team that wants caveman in (say)
`lite` for one project and `full` for another has no clean option — they
either pollute every contributor's shell rc, or rely on every contributor
to remember a manual `/caveman lite` on session start.

This PR adds a third resolution layer between env and user config: a
checked-in repo-local config at `<repo>/.caveman/config.json` or
`<repo>/.caveman.json`. The walker climbs from `process.cwd()` to the
filesystem root (bounded to 64 ancestors), so any sub-directory of the
repo picks it up.

## Resolution order (new)

```
1. CAVEMAN_DEFAULT_MODE env var      (unchanged, still highest priority)
2. <repo>/.caveman/config.json       (NEW — preferred at each ancestor)
   <repo>/.caveman.json              (NEW — fallback at each ancestor)
3. user config (XDG / ~/.config / %APPDATA%)   (unchanged)
4. 'full'                             (unchanged)
```

Schema for the repo-local file matches the existing user config:

```json
{ "defaultMode": "lite" }
```

## Backwards compatibility

Purely additive. Every existing path through `getDefaultMode()` is
preserved:

- Env-only users: env wins, never reach repo-local lookup.
- User-config-only users: no `.caveman/` in any ancestor → fall through
  to existing user-config branch.
- Repos with no caveman config: behavior unchanged from `main`.

## Security

`findRepoConfigPath()` rejects symlinked config files via `lstatSync` +
`isSymbolicLink()` check before reading, symmetric with the existing
`safeWriteFlag` / `readFlag` policy in the same module. Malformed JSON
and unknown mode strings silently fall through to the next layer, matching
the existing user-config behavior.

## Test plan

`tests/test_repo_local_config.js` — 12 cases, all passing on Linux Node 22:

- [x] no config anywhere → `full`
- [x] `.caveman/config.json` in cwd
- [x] `.caveman.json` in cwd
- [x] `.caveman/config.json` wins over `.caveman.json` at the same level
- [x] walks up from a nested cwd to find the repo config
- [x] env var beats repo-local
- [x] repo-local beats user config
- [x] no repo config → user config still wins
- [x] invalid mode in repo config falls through
- [x] malformed JSON in repo config falls through
- [x] symlinked `.caveman.json` is refused
- [x] `findRepoConfigPath` returns `null` outside any repo

```
$ node tests/test_repo_local_config.js
repo-local config resolution tests
  ✓ returns "full" when no env, no repo config, no user config
  ✓ reads .caveman/config.json in cwd
  ✓ reads .caveman.json in cwd
  ✓ .caveman/config.json wins over .caveman.json at same level
  ✓ walks up from nested cwd to find repo config
  ✓ env var beats repo-local config
  ✓ repo-local config beats user config
  ✓ falls through to user config when repo config absent
  ✓ invalid mode in repo config falls through to default
  ✓ malformed JSON in repo config falls through to default
  ✓ refuses symlinked .caveman.json (symmetric with readFlag policy)
  ✓ findRepoConfigPath returns null outside any repo
12 passed, 0 failed
```

Existing `test_symlink_flag.js` (12) and `test_caveman_stats.js` (29) still pass;
two pre-existing failures in `test_caveman_init.js` are unrelated (same
failures on `main` before this PR).

## Files

- `src/hooks/caveman-config.js` — new `findRepoConfigPath()`, refactor of the
  config-read into `readModeFromConfigFile()`, updated header doc.
- `tests/test_repo_local_config.js` — new.
- `CLAUDE.md` — updated `getDefaultMode()` doc + new `findRepoConfigPath()` entry.
- `skills/caveman-help/SKILL.md` — updated "Configure Default Mode" section
  so users see all three layers.

## Adversarial review (against my own diff)

- **Risk**: walker climbs forever on a pathological mount → bounded to 64 ancestors.
- **Risk**: a `.caveman.json` planted by a malicious dependency in a subdir of
  `node_modules` could hijack default mode → only candidate file names are checked
  (no glob), and `defaultMode` values are whitelist-validated against `VALID_MODES`,
  so the worst-case is a mode-of-the-attacker's-choice, not arbitrary execution.
  Acceptable given the threat model (defaults are advisory; the user can always
  override per-message).
- **Risk**: behavior change for users who *expected* user config to always win →
  documented as new precedence layer in both CLAUDE.md and the SKILL help; env
  layer remains the escape hatch for any user who wants to override repo intent.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -154,7 +154,8 @@ All hooks honor `CLAUDE_CONFIG_DIR` for non-default Claude Code config locations
 ### `src/hooks/caveman-config.js` — shared module
 
 Exports:
-- `getDefaultMode()` — resolves default mode from `CAVEMAN_DEFAULT_MODE` env var, then `$XDG_CONFIG_HOME/caveman/config.json` / `~/.config/caveman/config.json` / `%APPDATA%\caveman\config.json`, then `'full'`
+- `getDefaultMode()` — resolves default mode in order: `CAVEMAN_DEFAULT_MODE` env var → repo-local config (`<cwd>/.caveman/config.json` or `<cwd>/.caveman.json`, walking up to the filesystem root) → user config (`$XDG_CONFIG_HOME/caveman/config.json` / `~/.config/caveman/config.json` / `%APPDATA%\caveman\config.json`) → `'full'`. Repo-local config lets a team check in a per-project default without polluting every contributor's env or user config.
+- `findRepoConfigPath(start)` — walks up from `start` (default `process.cwd()`) looking for the first `.caveman/config.json` or `.caveman.json`. Bounded to 64 ancestors. Refuses symlinked files (symmetric with `safeWriteFlag` / `readFlag`).
 - `safeWriteFlag(flagPath, content)` — symlink-safe flag write. Refuses if flag target or its immediate parent is a symlink. Opens with `O_NOFOLLOW` where supported. Atomic temp + rename. Creates with `0600`. Protects against local attackers replacing the predictable flag path with a symlink to clobber files writable by the user. Used by both write hooks. Silent-fails on all filesystem errors.
 
 ### `src/hooks/caveman-activate.js` — SessionStart hook
--- a/skills/caveman-help/SKILL.md
+++ b/skills/caveman-help/SKILL.md
@@ -45,14 +45,20 @@ Default mode = `full`. Change it:
 export CAVEMAN_DEFAULT_MODE=ultra
 ```
 
-**Config file** (`~/.config/caveman/config.json`):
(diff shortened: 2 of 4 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`cd4009ef`](https://github.com/JuliusBrussee/caveman/commit/cd4009ef) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #434, opened 2026-05-24
fix(opencode): add missing caveman-compress.md command file and guard missing file in install loop

peppekerstens:
## Changes

Two bugs found during OpenCode install:

### 1. Missing command file
`caveman-compress.md` is listed in `OPENCODE_COMMAND_FILES` but doesn't exist in `src/plugins/opencode/commands/`. The install loop at line 555 lacks a `fs.existsSync(src)` check (present in the agent copy loop at line 566), causing `copyFileSync` to throw ENOENT.

**Fix**: Created `src/plugins/opencode/commands/caveman-compress.md` with command description matching caveman-compress skill docs.

### 2. Missing source existence check
Added `if (!fs.existsSync(src)) continue;` guard to the command file copy loop, matching the pattern used by the agent copy loop. Prevents silent ENOENT crash when a source file is missing.

### 3. Gitignore scope fix
Changed `caveman-compress.md` → `/caveman-compress.md` in `.gitignore` so the static command file under `src/` is tracked while root-level generated output files remain ignored.

## Test
- `npm pack --dry-run` confirms `caveman-compress.md` is included in the opencode plugin commands
- Verified the install loop skips safely when a source file is missing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ __pycache__/
 *.pyc
 .venv/
 .env.local
-caveman-compress.md
+/caveman-compress.md
 **/.DS_Store
 .claude/worktrees/
 evals/snapshots/*.html
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,9 +91,10 @@ caveman/
 | `skills/caveman-help/SKILL.md` | Quick-reference card. One-shot display, not a persistent mode. |
 | `skills/caveman-compress/SKILL.md` | Compress sub-skill behavior. |
 | `skills/cavecrew/SKILL.md` | Cavecrew decision guide — when to delegate to caveman subagents vs vanilla. Edit only here. |
-| `agents/cavecrew-investigator.md` | Read-only locator subagent (haiku). Output contract: `path:line — symbol — note`. |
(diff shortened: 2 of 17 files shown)


pdabney:
+1 confirmed repro on fresh install via `npx caveman-installer`:

```
failed:
  • opencode — ENOENT: no such file or directory, copyfile '.../caveman-installer/src/plugins/opencode/commands/caveman-compress.md' -> '/Users/.../.config/opencode/commands/caveman-compress.md'
```

Applied this PR's `caveman-compress.md` locally — install completes cleanly. The `existsSync` guard on the command copy loop is a nice belt-and-suspenders too; matches the pattern already in the agent loop. LGTM.

SamPlaysKeys:
Tested and confirmed as well, install runs clean after using this PR's caveman-compress.md. I didn't test the `existsSync`, but it looks good, and matches the current workflow pattern.
For reference, no AI was used in this testing/review.

SamPlaysKeys (approved):
The limited review I did looks good, and the two tiered approach appears to be working.

SamPlaysKeys:
The approval was by accident, I haven't looked over the entire PR, just the caveman compress and the base function of the `existsSync` check. That being said, I'm using this fork now, and not currently seeing any issues.

JuliusBrussee:

PR #435, opened 2026-05-24
fix(compress): resolve claude binary via shutil.which for Windows .cmd shims

m9751:
## Summary

`caveman-compress` fails on Windows 11 with `WinError 2: file not found` when invoking the `claude` CLI fallback. Cause: `npm install -g @anthropic-ai/claude-code` puts a `claude.CMD` shim on PATH, and Python's `subprocess.run(["claude", ...])` does not consult PATHEXT, so a bare `"claude"` argv entry never resolves on Windows.

This silently breaks the documented "CLI fallback uses your existing Claude Code desktop auth" path for any Windows user without an `ANTHROPIC_API_KEY` set — which is most Claude Code subscribers.

## Fix

Resolve the binary via `shutil.which("claude")`. On Windows, `shutil.which` consults PATHEXT and returns the full path to `claude.CMD`. On POSIX, it returns the same absolute path the implicit lookup would produce, so behavior is unchanged. No `shell=True`, no string interpolation, no injection surface change.

```diff
+import shutil
 ...
+claude_bin = shutil.which("claude") or "claude"
 result = subprocess.run(
-    ["claude", "--print"],
+    [claude_bin, "--print"],
     input=prompt,
     ...
 )
```

## Test plan

- [x] Reproduced original failure on Win11 (`WinError 2` from `subprocess.run(["claude", ...])`)
- [x] Patched, ran `python3 -m scripts <12KB markdown>` end-to-end
- [x] subprocess exit 0, validation passed first attempt, backup written
- [x] Content integrity verified: 32/32 trigger phrases preserved, 47/47 markdown links preserved, headings + code blocks + inline backticks intact
- [ ] Maintainer to confirm POSIX behavior unchanged (should be a no-op there)

Discovered while running ClawGuard audit + dry-run on a memory file under a Claude Code subscription (no `ANTHROPIC_API_KEY` set).

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -8,6 +8,7 @@ Usage:
 
 import os
 import re
+import shutil
 import subprocess
 from pathlib import Path
 from typing import List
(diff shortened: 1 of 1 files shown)


m9751 (commented):
approved

m9751 (commented):
approve

m9751:
Closing — this fix was folded into upstream `main` via commit e8eae0ff ("fix(compress): utf-8 pin, Windows .cmd resolve, frontmatter + backup-dir"), whose message explicitly credits #435 ("resolve claude via shutil.which for .cmd shims"). The `claude_bin = shutil.which("claude") or "claude"` change is present on main as of 2026-06-01. Thanks for batching it in. Nothing left to merge.

PR #443, opened 2026-05-26
fix(opencode): strip tools array from cavecrew agent frontmatter

### LINKED ISSUE #386: bug(opencode): cavecrew agent files use invalid `tools` format, breaking OpenCode startup
> :robot: _This was written by an AI agent on behalf of @paolomainardi._

## Description

The cavecrew agent files shipped in `agents/` use a YAML array for the `tools` frontmatter field:

```yaml
tools: [Read, Grep, Bash]
```

OpenCode expects `tools` to be an **object** (map of tool name to boolean) or **undefined** — not an array. This causes OpenCode to reject the config on startup:

```
Configuration is invalid at /home/user/.config/opencode/agents/cavecrew-reviewer.md
↳ Expected object | undefined, got ["Read","Grep","Bash"] tools
```

OpenCode refuses to start until the invalid file is removed. The only workaround is `rm -rf ~/.config/opencode/agents/`, which also removes any user-created agent files.

## Affected files

All three cavecrew agents have the same issue:
- `agents/cavecrew-reviewer.md` — `tools: [Read, Grep, Bash]`
- `agents/cavecrew-builder.md` — likely same pattern
- `agents/cavecrew-investigator.md` — likely same pattern

## Root cause

`installOpencode()` in `bin/install.js` copies agent files verbatim from the repo's `agents/` directory into `~/.config/opencode/agents/`. These agent files were written for Claude Code's frontmatter schema, which accepts a `tools` array. OpenCode uses a different schema where `tools` must be an object or omitted entirely.

The correct OpenCode format uses `permission` instead of `tools`:

```yaml
permission:
  read: allow
  grep: allow
  bash: allow
  edit: deny
```

## Suggested fix

Either:
1. **Convert the frontmatter** to use OpenCode's `permission` schema before copying (transform in `installOpencode()`)
2. **Ship separate OpenCode-compatible agent files** (e.g. in `src/plugins/opencode/agents/`)
3. **Strip the `tools` field** when copying to the OpenCode agents dir, since OpenCode defaults to allowing all tools when the field is omitted

## Environment

- OpenCode latest
- caveman: installed via `node bin/install.js --only opencode`
- OS: Linux (Arch)

ousamabenyounes:
Fixes #386

## Summary

OpenCode rejects YAML array form `tools: [...]` for subagent frontmatter and refuses to boot:

```
Configuration is invalid at .../cavecrew-reviewer.md
↳ Expected object | undefined, got ["Read","Grep","Bash"] tools
```

`installOpencode()` in `bin/install.js` copied `agents/cavecrew-*.md` verbatim into `~/.config/opencode/agents/`. Three files affected: `cavecrew-investigator.md`, `cavecrew-builder.md`, `cavecrew-reviewer.md`.

## Fix

New helper `bin/lib/opencode-agent.js` exports `stripOpencodeAgentTools(content)` that removes the `tools:` line (inline-array form **and** multi-line YAML list form) from the leading frontmatter block. Installer's section 3 now reads → strips → writes instead of `copyFileSync`.

OpenCode allows omitting `tools` entirely; it then falls back to its default tool set. The cavecrew subagent prompts already self-restrict in the body ("Read-only locator", "No `Bash` available", etc.), so dropping the array form is safe.

Source files in `agents/` are left untouched — they keep the Claude Code schema for the Claude Code plugin path. Transform applies only on the OpenCode copy.

## Verification

Baseline (main): 47 pass / 4 fail / 51 total — 4 pre-existing opencode failures (caveman-compress.md missing in `src/plugins/opencode/commands/`, separate issue).

Post-fix: 56 pass / 4 fail / 60 total — 9 new tests, **zero regressions**, same 4 pre-existing failures.

**TDD verification**

- RED check (helper missing): test file fails to import → 0/1.
- GREEN check (helper present): 9/9 pass.

Logs persisted locally in `~/.claude/contribute/working/caveman/issue_386_{RED,GREEN}.log`.

## Files changed

| File | Change |
|---|---|
| `bin/lib/opencode-agent.js` | New. `stripOpencodeAgentTools()` helper. |
| `bin/install.js` | Section 3 of `installOpencode` now sanitizes on copy. |
| `tests/installer/opencode-agent.test.mjs` | New. 9 tests: inline array, multi-line list, folded `description: >`, no-frontmatter, no-tools, non-string, real shipped files RED/GREEN, end-to-end installer-equivalent copy. |

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -23,6 +23,7 @@ const readline = require('readline');
 
 const SETTINGS = require('./lib/settings');
 const OPENCLAW = require('./lib/openclaw');
+const { stripOpencodeAgentTools } = require('./lib/opencode-agent');
 
 const REPO = 'JuliusBrussee/caveman';
 const RAW_BASE = `https://raw.githubusercontent.com/${REPO}/main`;
@@ -557,7 +558,11 @@ function installOpencode(ctx) {
       process.stdout.write(`  installed: ${dest}\n`);
     }
 
-    // 3. Subagents.
+    // 3. Subagents. Source files target Claude Code's schema (`tools: [...]`
+    //    YAML array); OpenCode rejects that form and refuses to boot until the
+    //    file is removed. Strip the `tools:` line on copy — OpenCode falls back
+    //    to its default tool set, and subagent prompts already self-restrict in
+    //    the body. Issue 386.
     fs.mkdirSync(agentsDir, { recursive: true });
     const agentSrcDir = path.join(repoRoot, 'agents');
     for (const f of OPENCODE_AGENT_FILES) {
@@ -565,7 +570,7 @@ function installOpencode(ctx) {
       const dest = path.join(agentsDir, f);
       if (!fs.existsSync(src)) continue;
       if (fs.existsSync(dest) && !opts.force) { note(`  skipped ${dest} (exists; --force to overwrite)`); continue; }
-      fs.copyFileSync(src, dest);
+      fs.writeFileSync(dest, stripOpencodeAgentTools(fs.readFileSync(src, 'utf8')));
       process.stdout.write(`  installed: ${dest}\n`);
     }
 
--- a/bin/lib/opencode-agent.js
+++ b/bin/lib/opencode-agent.js
@@ -1,0 +1,42 @@
+'use strict';
+
+// Strip the `tools:` field from a Claude-Code-style subagent frontmatter so
+// the file is valid for OpenCode, whose schema rejects the YAML array form
+// (`tools: [Read, Grep, Bash]`) with:
+//
+//   Configuration is invalid at .../agents/cavecrew-reviewer.md
+//   ↳ Expected object | undefined, got ["Read","Grep","Bash"] tools
+//
+// OpenCode allows `tools` to be a map (`{read: true, grep: true}`) or
+// omitted entirely. Omitting falls back to OpenCode's default tool set,
+// which is what the cavecrew subagent prompts already self-restrict against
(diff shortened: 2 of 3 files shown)


ousamabenyounes:
## Verified RED/GREEN proof (run on `8014b87` against `origin/main`)

Test runner: `node --test tests/installer/opencode-agent.test.mjs`.

Procedure:
1. GREEN — full PR diff applied
2. Replace `bin/lib/opencode-agent.js` with an identity stub (preserves the `require` resolution so unrelated tests still load, kills the strip-behavior so any test that depends on it fails)
3. RED — same test file
4. Restore helper byte-identical

### GREEN (fix applied)

```
# tests 9
# pass 9
# fail 0
```

### RED (helper neutered to identity)

```
# tests 9
# pass 4
# fail 5
```

The 5 failing tests on RED are precisely the ones that exercise the strip-behavior:

  - strips inline `tools: [...]` array from frontmatter
  - strips multi-line `tools:` list with indented continuation
  - preserves folded `description: >` continuation lines when `tools:` follows
  - all shipped cavecrew agent files become OpenCode-safe after transform (GREEN proof)
  - installer-equivalent copy writes OpenCode-safe agent file (issue 386 end-to-end)


The 4 tests that pass on RED are deliberate pass-through guards (no frontmatter, no `tools:` field, non-string input, and the RED-proof check that asserts the source `agents/cavecrew-*.md` files contain the offending array form — which is a property of the source tree, independent of the helper).

Both logs captured locally and available on request.

ousamabenyounes:
Closing as superseded: the fix is already on `main` via commit `ed1fbb7` (issue #386). Thanks!

PR #459, opened 2026-05-28
fix(statusline): render ⛏ correctly on Windows, add spacing

luojeffery:
## Summary

- Set `[Console]::OutputEncoding = UTF8` in `caveman-statusline.ps1` so `⛏` (U+26CF) renders instead of garbled bytes on Windows (PS5.1 defaults to system code page)
- Add `-Encoding UTF8` to `Get-Content` so the suffix file is read correctly
- Add a second space between `⛏` and the token count for visual breathing room

## Test plan

- [ ] Run `/caveman-stats` after install
- [ ] Statusline shows `[CAVEMAN] ⛏  12.4k` with visible space between pickaxe and number
- [ ] No garbled characters on Windows

### DIFF
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -324,7 +324,7 @@ function main() {
     // Routed through safeWriteFlag — the suffix path is predictable and
     // user-owned, same symlink-clobber surface as the .caveman-active flag.
     const agg = aggregateHistory(historyPath, null);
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`8753bf90`](https://github.com/JuliusBrussee/caveman/commit/8753bf90) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #472, opened 2026-05-29
fix(install): prune orphaned managed hooks from settings.json (#471)

### LINKED ISSUE #471: Plugin migration leaves orphaned settings.json hooks → 'Cannot find module caveman-activate.js' (loader:1478) every session
## Summary

After migrating from the **manual hook install** (`~/.claude/hooks/caveman-*.js` wired into `settings.json`) to the **Claude Code plugin** (`claude plugin install caveman@caveman`), the old `settings.json` hook entries are left **orphaned**. The plugin migration disables/removes the local `~/.claude/hooks/caveman-activate.js` (mine are now `*.js.disabled`) but does **not** remove the `settings.json` entries that point at them. Result: every session and every prompt, Claude Code runs `node <abs> /…/caveman-activate.js` against a missing file:

```
Failed with non-blocking status code: node:internal/modules/cjs/loader:1478
Error: Cannot find module '/Users/<me>/.claude/hooks/caveman-activate.js'
```

Same for `caveman-mode-tracker.js` on `UserPromptSubmit`, and the `statusLine` command pointing at `caveman-statusline.sh` (now `.disabled`).

It's non-blocking so caveman still works (the plugin's SessionStart `echo` hook fires), but every session and prompt is noisy with a Node `loader:1478` crash. It also misleads debugging — the error renders directly above the `claude-mem` status banner, so it reads as a claude-mem failure when it is caveman's.

## Root cause

In `bin/lib/settings.js`:

- `rewriteLegacyManagedHookCommands()` only matches **bare** `node <script>` (`const reBare = /^node\s+(...)\s*$/`) and explicitly skips commands with an absolute node path. My orphaned entries are `"/opt/homebrew/Cellar/node/26.0.0/bin/node" "/…/caveman-activate.js"` — absolute node path → never rewritten, never validated against file existence.
- `removeCavemanHooks(settings, marker='caveman')` *would* match these by substring, but it's only invoked on **uninstall**, not on the plugin-install migration path (`bin/install.js` `claude plugin marketplace add` / `plugin install caveman@caveman`).

So the plugin install adds the plugin's own hooks but never reconciles the pre-existing manual-install hooks, leaving dead `node <missing>.js` entries.

## Repro

1. Have caveman installed the old way (settings.json `SessionStart`/`UserPromptSubmit`/`statusLine` → `~/.claude/hooks/caveman-*.js|sh`).
2. Migrate to the plugin: `claude plugin install caveman@caveman` (local hook files get disabled/renamed).
3. Start a new Claude Code session → `loader:1478` / `Cannot find module … caveman-activate.js` every session; same on each prompt.

## Suggested fix

On plugin-based install, reconcile pre-existing manual hooks. Either:

- Call `removeCavemanHooks(settings)` for the managed basenames (`caveman-activate.js`, `caveman-mode-tracker.js`, `caveman-statusline.sh`, `caveman-stats.js`) during plugin install, **OR**
- Extend the legacy walker to drop/repair any `settings.json` hook command whose referenced managed script no longer exists on disk (covers both bare- and absolute-node shapes), **OR**
- At minimum, when the plugin path is taken, strip the manual `settings.json` SessionStart/UserPromptSubmit/statusLine entries the installer itself previously wrote.

A "remove managed hook entries whose target file is missing" pass would be the most robust — it self-heals regardless of how the entry was originally written.

## Workaround

Remove the dead entries from `~/.claude/settings.json`: the `SessionStart` → `caveman-activate.js` block, the `UserPromptSubmit` → `caveman-mode-tracker.js` block, and repoint/remove `statusLine` (point at the existing `…/caveman-statusline.sh.disabled` or a plugin-provided one).

Env: macOS (Darwin 25.5.0), Claude Code v2.1.156, Node v26, caveman plugin `caveman@caveman` enabled.

praxstack:
## What

Adds `pruneOrphanedManagedHooks()` to `bin/lib/settings.js` and runs it during `installClaude()` to remove managed `settings.json` hook / statusLine entries whose target script no longer exists on disk.

Fixes #471.

## Why

Migrating an old **manual** hook install (`settings.json` hooks → `~/.claude/hooks/caveman-*.js`) to the **Claude Code plugin** disables/renames the local scripts (e.g. `caveman-activate.js` → `caveman-activate.js.disabled`) but leaves the `settings.json` entries pointing at them. Claude Code then runs `node <missing>` every `SessionStart` / `UserPromptSubmit`:

```
Failed with non-blocking status code: node:internal/modules/cjs/loader:1478
Error: Cannot find module '…/.claude/hooks/caveman-activate.js'
```

Non-blocking, so caveman still works (plugin manifest fires) — but every session and prompt is noisy, and the error renders directly above unrelated plugins' output, so it misattributes the failure (I chased a different plugin for two sessions before isolating it to caveman).

Existing helpers don't cover it:
- `rewriteLegacyManagedHookCommands` only matches the **bare-node** shape; these orphans are **absolute-node** (`"/opt/…/node" "…/caveman-activate.js"`), which it explicitly skips.
- `removeCavemanHooks` runs only on **uninstall**.

## How

`pruneOrphanedManagedHooks(settings, configDir)`:
- Extracts the script path from any managed-looking hook command — bare- or absolute-node, quoted or not, and `bash …` for the statusline — by matching the managed basenames (`caveman-activate.js`, `caveman-mode-tracker.js`, `caveman-stats.js`, `caveman-statusline.sh`).
- Resolves relative paths against `configDir` (honors `CLAUDE_CONFIG_DIR`).
- Drops the hook entry / `statusLine` **only when the target file is genuinely missing**. A managed hook whose script still exists is left untouched; non-managed hooks (user/superset notify, `rtk`, etc.) are never touched — safe to run on every install.

`installClaude()` runs the prune **unconditionally**, right after the plugin install step and before the hook-wiring decision, so it self-heals an already-dirty config even when standalone wiring is skipped.

## Relationship to #393

Complementary, not overlapping. #393 prevents the **double-wiring** case (fresh install registers hooks via both plugin manifest and settings.json → fire twice). This cleans **pre-existing orphans** left by a migration that happened before any fix. A user who migrated before #393 still has the dead entries; #393 only gates *new* wiring and never removes them. The two work together.

## Tests

6 new unit cases in `tests/installer/unit.settings.test.mjs`:
- removes absolute-node orphan
- removes bare-node orphan
- keeps managed hook whose target exists
- leaves non-managed hooks alone even if missing
- resolves relative target against `configDir`
- drops orphaned managed `statusLine`

`node --test tests/installer/unit.settings.test.mjs` → 25/25 pass. Verified end-to-end against a `settings.json` reproducing the crash: 3 orphans removed (SessionStart + UserPromptSubmit + statusLine), the superset notify hook preserved.

> Note: `tests/installer/opencode.test.mjs` fails on a clean checkout here (Node 26 `JSON.parse` strictness on a JSONC fixture) — pre-existing, unrelated to this change.

CI will resync any mirrors on merge; no source-of-truth skill files touched.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -400,6 +400,29 @@ async function installClaude(ctx) {
     else results.failed.push(['claude', 'claude plugin install failed']);
   }
 
+  // Self-heal: drop managed settings.json hook/statusLine entries whose target
+  // script no longer exists. Migrating an old manual install to the plugin
+  // disables the local ~/.claude/hooks/caveman-*.js scripts but leaves the
+  // settings.json entries pointing at them, so Claude Code crashes every
+  // SessionStart / UserPromptSubmit with `loader:1478 — Cannot find module
+  // …caveman-activate.js` (issue #471). Runs unconditionally (independent of
+  // the hook-wiring decision) so it repairs an already-dirty config even when
+  // we then skip standalone wiring because the plugin manifest handles hooks.
+  {
+    const settingsPath = path.join(SETTINGS.claudeConfigDir(), 'settings.json');
+    const settings = SETTINGS.readSettings(settingsPath);
+    if (settings) {
+      const pruned = SETTINGS.pruneOrphanedManagedHooks(settings, SETTINGS.claudeConfigDir());
+      if (pruned > 0) {
+        note(`  removed ${pruned} orphaned caveman hook entr${pruned === 1 ? 'y' : 'ies'} from settings.json (target script missing)`);
+        if (!opts.dryRun) {
+          SETTINGS.validateHookFields(settings);
+          SETTINGS.writeSettings(settingsPath, settings);
+        }
+      }
+    }
+  }
+
   if (opts.withHooks) {
     say('  → installing hooks (--with-hooks)');
     const r = await installHooks(ctx);
--- a/bin/lib/settings.js
+++ b/bin/lib/settings.js
@@ -230,23 +230,38 @@ function pruneOrphanedManagedHooks(settings, configDir) {
   const baseDir = configDir || claudeConfigDir();
   let removed = 0;
 
-  // Pull the first token that ends in a managed basename out of a command
-  // string. Handles: `node "/a/x.js"`, `"/abs/node" "/a/x.js"`, `bash /a/x.sh`.
-  const managedAlt = [...MANAGED_HOOK_BASENAMES]
-    .map(b => b.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'))
-    .join('|');
-  const reManaged = new RegExp(`(?:"([^"]*(?:${managedAlt}))"|'([^']*(?:${managedAlt}))'|(\\S*(?:${managedAlt})))`);
+  // Split a command into shell-ish tokens, honoring single/double quotes so a
+  // path containing spaces survives intact. Good enough for hook commands we
+  // generate (`node "/a/x.js"`, `"/abs/node" "/a/x.js"`, `bash /a/x.sh`); not
+  // a full shell parser.
+  const tokenize = (command) => {
+    const out = [];
+    const re = /"([^"]*)"|'([^']*)'|(\S+)/g;
+    let m;
+    while ((m = re.exec(command)) !== null) out.push(m[1] ?? m[2] ?? m[3]);
+    return out;
+  };
 
+  // A command is a missing managed target iff some token's BASENAME exactly
+  // equals a managed script (exact match — not substring — so a user hook like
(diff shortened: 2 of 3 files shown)


praxstack:
Closing this — `pruneOrphanedManagedHooks` already landed on `main` (it's wired into `install.js` and covers the same cases, including orphaned managed `statusLine`). The shipped implementation is actually more robust than this PR's: it matches on exact basename (`path.basename(tok)` ∈ managed set) rather than a substring regex, so a user hook like `mycaveman-activate.js` is correctly left untouched, and it handles quoted paths with spaces.

No need to carry a redundant change — thanks for getting the fix in. 🙏

PR #477, opened 2026-05-31
refactor: rename caveman plugin configuration file to plugin.json (in order to be detected by Antigravity 2.0 onwards)

leocura:


JuliusBrussee:
Thanks @leocura — Antigravity now uses the shared provider/profile route, so this root-file rename no longer applies. Closing as superseded.

PR #478, opened 2026-06-01
fix: rename $Args to $InstallArgs in install.ps1 for Windows

vaibhavkumar779:
## Problem

`install.ps1` uses `$Args` as a `param()` name, but `$Args` is a [PowerShell automatic variable](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1#args) that cannot be overwritten inside `Invoke-Expression`. When users run the documented one-liner:

```powershell
irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 | iex
```

PowerShell throws:

```
Invoke-Expression : Cannot overwrite variable Args because the variable has been optimized.
    + CategoryInfo          : WriteError: (Args:String) [Invoke-Expression], SessionStateUnauthorizedAccessException
    + FullyQualifiedErrorId : VariableNotWritableRare,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

This breaks the install on **every Windows machine** using PowerShell 5.1 (the default shell).

### Why it happens

When a script is piped to `Invoke-Expression`, PowerShell compiles the script block in the caller's scope. The automatic `$Args` variable is marked as an optimized/read-only variable in that context, so `param([string[]]$Args)` fails because it tries to reassign it.

Running the script as a file (`& .\install.ps1`) works fine because it gets its own scope — but that defeats the purpose of the one-liner install.

## Environment

- **OS:** Windows 11
- **PowerShell:** 5.1.26100.4061
- **Editor:** VS Code with GitHub Copilot agent
- **Node:** v22.x

### 1. `install.ps1` — Rename parameter (3 occurrences)

| Before | After |
|--------|-------|
| `[string[]]$Args` | `[string[]]$InstallArgs` |
| `@Args` (splat in local path) | `@InstallArgs` |
| `@Args` (splat in npx path) | `@InstallArgs` |

Also updated the header comment to use the download-then-run pattern and added a `NOTE` block explaining why `$Args` cannot be used.

### 2. `README.md` — Updated Windows install command

```diff
-# Windows (PowerShell 5.1+)
-irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 | iex
+# Windows (PowerShell 5.1+)
+$f = "$env:TEMP\caveman-install.ps1"; irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 -OutFile $f; & $f; Remove-Item $f
```

### 3. `INSTALL.md` — Updated install command + troubleshooting

- Updated the Windows install command to match README
- Replaced the `irm | iex` troubleshooting bullet with an explicit warning and the correct command

## Testing

1. **Before fix:** `irm ... | iex` → crashes with `VariableNotWritableRare`
2. **After fix (download-then-run):** Downloads to `%TEMP%`, runs in own scope, installs successfully, cleans up temp file
3. **After fix (local clone):** `& .\install.ps1` still works (no regression)
4. **Verified:** All 7 skills installed successfully for GitHub Copilot on Windows 11

## Checklist

- [x] `$Args` → `$InstallArgs` in `install.ps1` (all 3 occurrences)
- [x] README.md Windows command updated
- [x] INSTALL.md Windows command updated
- [x] INSTALL.md troubleshooting section updated
- [x] No breaking changes for macOS/Linux (`install.sh` untouched)
- [x] No breaking changes for local clone usage (`& .\install.ps1` still works)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,7 +15,7 @@ curl -fsSL https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.
 **Windows (PowerShell 5.1+)**
 
 ```powershell
-irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 | iex
+$f = "$env:TEMP\caveman-install.ps1"; irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1 -OutFile $f; & $f; Remove-Item $f
 ```
 
 What it does:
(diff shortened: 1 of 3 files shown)


JuliusBrussee:
Thanks @vaibhavkumar779 — the reserved-parameter fix is already on main. Closing this duplicate implementation.

PR #481, opened 2026-06-01
fix(compress): harden validation, CLI, benchmark, and default model

brandonwie:
## Summary

Five fixes to the `caveman-compress` skill (in `skills/caveman-compress/scripts/`
and the synced copy). Each closes a correctness gap where the validator could
pass while the compressed output lost structure, or where a helper failed in an
install-only copy.

### Fixes

- **Validation hardening (`validate.py`)** — five gaps:
  - *Variable-length fences*: `extract_inline_codes` only stripped exact 3-char
    ` ``` `/`~~~` fences, so backticks inside a 4+ backtick block were misread as
    inline code → false failures. Now reuses the CommonMark fence-aware scanner
    (closing fence ≥ opening length).
  - *Indented code blocks*: 4-space/tab blocks were treated as prose and could be
    silently rewritten while validation passed. Now extracted and preserved like
    fenced blocks.
  - *Headings exact*: a heading text/order/level change with equal count was only
    a warning (accepted by `compress_file`). Anchors/TOCs/cross-links depend on
    exact headings → now a hard error.
  - *Redundant warning*: a count mismatch fired both an error and the text-change
    warning; the text branch is now `elif` → one error per condition, no noise.
  - *File paths exact*: a dropped/changed referenced path was only a warning → a
    *lost* path is now a hard error; an extra path-like token only in the output
    stays a warning (PATH_REGEX is broad; caveman prose like "pros/cons" can
    coincidentally match).
- **Extensionless build files (`detect.py`)** — `SKIP_EXTENSIONS` only matched the
  rare `*.dockerfile`/`*.makefile` forms, so real `Dockerfile`/`Makefile`
  (extensionless or `Dockerfile.prod`) fell through and could be compressed as
  prose. Adds a case-insensitive basename match that runs before the extension
  rules.
- **Benchmark degrades gracefully (`benchmark.py`)** — glob/no-args mode called
  `sys.exit(1)` when `tests/caveman-compress` was absent (an install copy without
  the repo's `tests/` tree). Benchmarking bundled fixtures is optional → prints a
  note and exits 0. Direct file-pair mode unchanged.
- **CLI usage (`cli.py`)** — corrects the printed usage to the module invocation.
- **Default model (`compress.py`)** — bumps the default to `claude-sonnet-4-6`.

### Tests

- Adds `tests/test_validate_inline.py`: fail/pass coverage for variable-length
  fences, indented blocks, heading-exact failures (text/level/count, no redundant
  warning), and path-loss-vs-prose.
- Corrects the `claude-md-project` compressed fixture, which had mutated a heading
  level (`###` → `##`) — that encoded the old warn-only behavior and now fails the
  headings-exact rule.

Local verification on this branch:
- `python3 -m unittest tests.test_validate_inline tests.test_compress_safety` → **29 tests OK**
- `python3 tests/verify_repo.py` → **all checks passed**

---

Co-authored-by: claude <noreply@anthropic.com>

### DIFF
--- a/plugins/caveman/skills/caveman-compress/scripts/benchmark.py
+++ b/plugins/caveman/skills/caveman-compress/scripts/benchmark.py
@@ -60,8 +60,12 @@ def main():
     # Walk up four dirs: scripts → caveman-compress → skills → repo_root.
     tests_dir = Path(__file__).resolve().parents[3] / "tests" / "caveman-compress"
     if not tests_dir.exists():
-        print(f"❌ Tests dir not found: {tests_dir}")
-        sys.exit(1)
+        # Glob mode is an optional helper that benchmarks the bundled fixtures.
+        # When they are absent (e.g. running from an install copy without the
+        # repo's tests/ tree), degrade gracefully instead of failing.
+        print(f"Note: fixtures dir not found, skipping benchmark: {tests_dir}")
+        print("Pass an original/compressed file pair to benchmark directly.")
+        return
 
     rows = []
     for orig in sorted(tests_dir.glob("*.original.md")):
--- a/plugins/caveman/skills/caveman-compress/scripts/cli.py
+++ b/plugins/caveman/skills/caveman-compress/scripts/cli.py
@@ -3,7 +3,7 @@
 Caveman Compress CLI
 
 Usage:
-    caveman <filepath>
+    python3 -m scripts <filepath>
 """
 
 import sys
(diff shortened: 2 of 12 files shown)


brandonwie:
Closing this for now — it's gone nearly a month without a review, and I've kept these fixes on my own fork in the meantime. The five correctness fixes (fence-aware validation, indented-code-block handling, CLI arg parsing, the benchmark, and the default-model bump) still apply if they're useful upstream. Happy to rebase and reopen if a maintainer wants to pick them up. Thanks for caveman — I use it daily.

PR #485, opened 2026-06-02
fix: prune orphaned settings.json hooks on plugin migration

### LINKED ISSUE #471: Plugin migration leaves orphaned settings.json hooks → 'Cannot find module caveman-activate.js' (loader:1478) every session

code-with-rashid:
## Problem

When migrating from a standalone hook install to the Claude Code plugin, old `settings.json` entries referencing hook files under `~/.claude/hooks/` are not cleaned up. The plugin system takes over hook delivery, but the manual entries remain pointing at files that may no longer exist. Every subsequent session emits noise like:

```
Error: Cannot find module '/Users/<me>/.claude/hooks/caveman-activate.js'
```

Two related root causes:

1. **No cleanup on plugin install** — `removeCavemanHooks()` only runs during `--uninstall`, never during plugin install, so orphaned entries from the previous standalone install persist indefinitely.

2. **`rewriteLegacyManagedHookCommands` only matched bare `node` commands** — the regex `^node\s+...` silently skipped hooks written with an absolute node path (e.g. `/opt/homebrew/Cellar/node/26.0.0/bin/node`), which is the format written by Homebrew, nvm, and the current installer itself. Stale absolute paths from old Node versions were never updated.

Closes #471

### `bin/lib/settings.js`

- **New `extractManagedScriptPath(command)`** — parses any node hook command variant (bare `node`, absolute path, quoted/unquoted) and returns the script path. Used by both functions below.
- **Rewrote `rewriteLegacyManagedHookCommands()`** — now handles all node executable formats, not just bare `node`. Skips only commands already using the exact target binary; rewrites everything else to the canonical `"<absoluteNode>" "<scriptPath>"` form.
- **New `pruneDeadHookFiles(settings)`** — walks all hook entries, checks `fs.existsSync` for each managed script path, and removes entries whose files are gone. Only touches basenames in `MANAGED_HOOK_BASENAMES`; user-authored hooks are always left alone.

### `bin/install.js`

- **`installClaude()`** — after plugin install (fresh or already-installed), reads `settings.json` and calls `pruneDeadHookFiles`. If orphaned entries are found, writes the cleaned file back and logs how many were removed.
- **`installHooks()`** — calls `pruneDeadHookFiles` before `rewriteLegacyManagedHookCommands` so dead entries and stale node paths are both resolved before idempotency checks run.

## Tests

Added 11 new tests in `tests/installer/unit.settings.test.mjs` (all 28 unit tests pass):

- `rewriteLegacyManagedHookCommands` rewrites stale absolute-node scripts (the old "ignores absolute node" behaviour was the bug)
- `rewriteLegacyManagedHookCommands` skips commands already using the target node binary
- `rewriteLegacyManagedHookCommands` rewrites unquoted absolute-node scripts
- `pruneDeadHookFiles` removes entries whose managed scripts are missing
- `pruneDeadHookFiles` keeps entries whose managed scripts exist on disk
- `pruneDeadHookFiles` does not touch user-authored hooks even if their files are missing
- `pruneDeadHookFiles` handles multiple events and mixed live/dead entries
- `pruneDeadHookFiles` handles the bare-node command format
- `pruneDeadHookFiles` is a no-op on empty or missing hooks
- `pruneDeadHookFiles` preserves non-hook settings keys (theme, statusLine, etc.)

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -380,7 +380,7 @@ function absoluteNodePath() {
 
 // ── Per-provider installers ────────────────────────────────────────────────
 async function installClaude(ctx) {
-  const { say, note, warn, ok, opts, results } = ctx;
+  const { say, note, warn, ok, opts, results, configDir } = ctx;
   results.detected++;
   say('→ Claude Code detected');
 
@@ -398,6 +398,23 @@ async function installClaude(ctx) {
     const r2 = runSpawn('claude', ['plugin', 'install', 'caveman@caveman'], null, opts.dryRun);
     if ((r1.status || 0) === 0 && (r2.status || 0) === 0) results.installed.push('claude');
     else results.failed.push(['claude', 'claude plugin install failed']);
+  }
+
+  // Prune settings.json entries for managed scripts that no longer exist on
+  // disk. Common after migrating from standalone hooks to the plugin (issue
+  // #471): the plugin system takes over hook delivery but old settings.json
+  // entries remain, causing "Cannot find module" noise every session.
+  if (!opts.dryRun) {
+    const settingsPath = path.join(configDir, 'settings.json');
+    const settings = SETTINGS.readSettings(settingsPath);
+    if (settings) {
+      const pruned = SETTINGS.pruneDeadHookFiles(settings);
+      if (pruned > 0) {
+        SETTINGS.validateHookFields(settings);
+        SETTINGS.writeSettings(settingsPath, settings);
+        note(`  removed ${pruned} orphaned hook entr${pruned === 1 ? 'y' : 'ies'} from settings.json`);
+      }
+    }
   }
 
   if (opts.withHooks) {
@@ -736,7 +753,13 @@ async function installHooks(ctx) {
   const tracker  = path.join(hooksDir, 'caveman-mode-tracker.js');
   const statusline = path.join(hooksDir, 'caveman-statusline.sh');
 
-  // Migrate any legacy bare-`node` invocations of our managed scripts.
+  // Remove any entries whose managed script files no longer exist on disk
+  // before rewriting/adding ours — keeps settings.json clean across reinstalls
+  // and node-upgrade scenarios.
+  SETTINGS.pruneDeadHookFiles(settings);
+
+  // Migrate stale node invocations (bare `node`, old absolute paths) to the
+  // current node binary. Handles Homebrew, nvm, and other non-PATH installs.
   SETTINGS.rewriteLegacyManagedHookCommands(settings, node);
 
   SETTINGS.addCommandHook(settings, 'SessionStart', {
--- a/bin/lib/settings.js
+++ b/bin/lib/settings.js
@@ -173,12 +173,29 @@ function removeCavemanHooks(settings, marker = 'caveman') {
   return removed;
 }
 
+// ── extractManagedScriptPath ──────────────────────────────────────────────
+// Parse the script path from any node hook command variant:
+//   node /path/to/script.js          (bare node, unquoted)
+//   node "/path/to/script.js"        (bare node, quoted)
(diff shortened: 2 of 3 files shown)


JuliusBrussee:
Thanks @code-with-rashid — current migration prunes only Caveman-owned orphan hooks and preserves foreign config. Closing as implemented.

PR #493, opened 2026-06-04
Fix OpenCode caveman installer

dhruv-anand-aintech:
## Summary
- add the missing OpenCode /caveman-compress command template so the installer no longer aborts mid-install
- convert copied cavecrew agent frontmatter from Claude-style `tools` arrays to OpenCode `permission` config during install
- add regression coverage for OpenCode-installed agent frontmatter

## Verification
- `npm test` (51/51 passing)
- `node --check bin/install.js && node --check src/plugins/opencode/plugin.js`
- temp OpenCode install with `node bin/install.js --only opencode --non-interactive --no-mcp-shrink --force` exits 0 and generates permission-based cavecrew agents

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__/
 .venv/
 .env.local
 caveman-compress.md
+!src/plugins/opencode/commands/caveman-compress.md
 **/.DS_Store
 .claude/worktrees/
 evals/snapshots/*.html
--- a/bin/install.js
+++ b/bin/install.js
@@ -489,6 +489,56 @@ function copyDirRecursive(src, dest) {
   }
 }
 
+function opencodeAgentFrontmatter(body) {
+  // OpenCode currently expects agent tool access as `permission`, not Claude's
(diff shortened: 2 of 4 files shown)


dhruv-anand-aintech:
Amended this branch with one review fix:

- skip OpenCode caveman-shrink MCP registration when no upstream MCP command is configured, because caveman-shrink is a proxy and exits when registered standalone
- added a regression test that runs the OpenCode installer with --with-mcp-shrink and asserts no broken caveman-shrink MCP entry is written

Validation:
- node --test tests/installer/opencode.test.mjs
- node --check bin/install.js
- git diff --check

JuliusBrussee:
Thanks for fixing the OpenCode installer. Equivalent installer and lifecycle fixes are now on main through [f0dd7803](https://github.com/JuliusBrussee/caveman/commit/f0dd7803) and [22f75e3d](https://github.com/JuliusBrussee/caveman/commit/22f75e3d), shipped in [v2.1.0](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected.

PR #498, opened 2026-06-06
fix(claude): SessionStart hook SKILL.md path broken after src/ restructure

### LINKED ISSUE #467: Claude Code SessionStart hook serves weak fallback — src/hooks/ SKILL.md path off by one
## Summary

After the `src/` restructure, the Claude Code **SessionStart hook can no longer find `SKILL.md`**, so it silently serves the weak hardcoded fallback ruleset on every session. The fallback has no intensity table and no examples, so the model drifts back to verbose prose within a few turns — matching the long-standing "drifts back to natural prose mid-session" / "lasts ~4 msgs" reports.

## Root cause: SKILL.md path is off by one

`.claude-plugin` wires the hook at `${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js`, so `__dirname` is now `<plugin_root>/src/hooks/`. But the SKILL.md read still uses:

```js
path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md')
```

From `src/hooks/` that resolves to `<plugin_root>/src/skills/caveman/SKILL.md`, which does not exist (skills live at `<plugin_root>/skills/`). `readFileSync` throws, the `catch` swallows it, and the hook falls through to the minimal hardcoded fallback in the `else` branch. The stale comment on the line still says `__dirname = <plugin_root>/hooks/`, confirming the path was never updated for the move.

## Repro

```bash
D=<plugin_root>   # e.g. ~/.claude/plugins/cache/caveman/caveman/<sha>
CAVEMAN_DEFAULT_MODE=ultra CLAUDE_PLUGIN_ROOT=$D node $D/src/hooks/caveman-activate.js
# Observe: output is the fallback ruleset — no "## Intensity" table, no examples.
```

## Fix

Probe both layouts before falling back:

```js
let skillContent = '';
const skillCandidates = [
  path.join(__dirname, '..', '..', 'skills', 'caveman', 'SKILL.md'), // src/hooks/ (current)
  path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'),       // hooks/ (older)
];
for (const candidate of skillCandidates) {
  try { skillContent = fs.readFileSync(candidate, 'utf8'); break; } catch (e) {}
}
```

Also update the stale `__dirname = <plugin_root>/hooks/` comment.

## Environment

- Claude Code, caveman plugin install, mode `ultra` (reproduces on any level).
- Plugin commit `655b7d9` (current HEAD at time of report).

This is the live cause of the drift reports on Claude Code post-restructure. Happy to open a PR.

ousamabenyounes:
Fixes #467

## Problem

After the `src/` restructure, `caveman-activate.js` moved from `hooks/` to `src/hooks/`, but the relative path to `skills/caveman/SKILL.md` was not updated. The hook resolved `__dirname/../skills/...` → `src/skills/...` — a path that doesn't exist. `readFileSync` threw, the `catch` swallowed it, and **every session silently served the weak hardcoded fallback ruleset** (no intensity table, no examples). Models drifted back to verbose prose within a few turns.

## Fix

Probe both layouts before falling back:

- `src/hooks/ → ../../skills/caveman/SKILL.md` (current layout)
- `hooks/     → ../skills/caveman/SKILL.md`    (legacy layout)

## Test — real RED → GREEN

`tests/test_activate_skillmd_path.js` spawns the real hook and asserts the full SKILL.md is loaded (the `## Intensity` table is SKILL.md-only — never in the fallback).

**GREEN (fix in place):**
```
✓ activation loads full SKILL.md (not the weak fallback)
✓ intensity table is filtered to the active level
2 passed, 0 failed
```

**RED (candidate path reverted to the stale `../skills`):**
```
✗ activation loads full SKILL.md (not the weak fallback)
    expected SKILL.md content ("## Intensity"), got fallback:
    Respond terse like smart caveman. … (hardcoded fallback, no ## Intensity)
✗ intensity table is filtered to the active level
0 passed, 2 failed
```
The reverted hook emits the weak fallback — the exact silent-degradation this issue reports. The test proves the SKILL.md path is actually resolved.

Run: `node tests/test_activate_skillmd_path.js`

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -53,9 +53,11 @@ const modeLabel = mode === 'wenyan' ? 'wenyan-full' : mode;
 // Standalone installs:                __dirname = $CLAUDE_CONFIG_DIR/hooks/, SKILL.md won't exist — falls back to hardcoded rules.
 let skillContent = '';
 const skillCandidates = [
-  path.join(__dirname, '..', '..', 'skills', 'caveman', 'SKILL.md'), // src/hooks/ → plugin root (current layout)
   path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'),       // hooks/ → plugin root (legacy layout)
 ];
+if (path.basename(path.dirname(__dirname)) === 'src') {
+  skillCandidates.unshift(path.join(__dirname, '..', '..', 'skills', 'caveman', 'SKILL.md')); // src/hooks/ → plugin root (current layout)
+}
 for (const candidate of skillCandidates) {
   try { skillContent = fs.readFileSync(candidate, 'utf8'); break; } catch (e) {}
 }
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -48,14 +48,17 @@ if (INDEPENDENT_MODES.has(mode)) {
 const modeLabel = mode === 'wenyan' ? 'wenyan-full' : mode;
 
 // Read SKILL.md — the single source of truth for caveman behavior.
-// Plugin installs: __dirname = <plugin_root>/hooks/, SKILL.md at <plugin_root>/skills/caveman/SKILL.md
-// Standalone installs: __dirname = $CLAUDE_CONFIG_DIR/hooks/, SKILL.md won't exist — falls back to hardcoded rules.
+// Plugin installs (post-restructure): __dirname = <plugin_root>/src/hooks/, SKILL.md at <plugin_root>/skills/caveman/SKILL.md
+// Plugin installs (pre-restructure):  __dirname = <plugin_root>/hooks/,    SKILL.md at <plugin_root>/skills/caveman/SKILL.md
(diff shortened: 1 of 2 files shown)


davidzha712:
+1 — independently hit this on a plugin install (caveman@caveman, commit-pinned). Confirmed root cause: `src/hooks/caveman-activate.js` reads `__dirname/../skills/caveman/SKILL.md` = `src/skills/...` which doesn't exist (SKILL.md is at `<root>/skills/caveman/`), so SessionStart silently serves the weak hardcoded fallback every session. Verified this PR's fix locally — the `src/hooks → ../../skills` candidate resolves correctly and the hook now emits the full SKILL.md ruleset (`## Intensity` table + examples) instead of the fallback. Keeping the legacy `../skills` candidate as a fallback is the right call for pre-restructure installs. Would be great to get this merged.

ousamabenyounes:
Closing as superseded: the fix already landed on `main` (commit `6d71b9e`, issue #467). Avoiding a redundant diff. Thanks!

PR #483, opened 2026-06-01
feat(caveman): extend write-normal carve-out to issue/PR/MR titles and descriptions

paolomainardi:
> :robot: _This was written by an AI agent on behalf of @paolomainardi._

## Summary

The active-mode reminder and skill boundaries already tell the model to write **code, commits, and security warnings** in normal prose. They say nothing about **issue, pull-request, and merge-request titles and descriptions** — which are prose artifacts that should not be compressed into caveman fragments. In practice the model writes terse fragmented PR/MR/issue text while caveman mode is active.

This extends the existing boundary line to explicitly cover issue/PR/MR titles and descriptions, across every source-of-truth surface that carries the carve-out (CI mirrors under `plugins/` are left untouched and regenerate from these).

## Changes

- `src/hooks/caveman-mode-tracker.js` — per-turn `UserPromptSubmit` reinforcement reminder
- `src/hooks/caveman-activate.js` — `SessionStart` full ruleset `## Boundaries`
- `skills/caveman/SKILL.md` — caveman skill behavior doc `## Boundaries`
- `src/rules/caveman-activate.md` — auto-activation rule body (Cursor/Windsurf/Cline/Copilot)
- `src/tools/caveman-init.js` — per-repo init template
- `src/plugins/opencode/plugin.js` and `src/plugins/opencode/commands/caveman.md` — opencode reminder + command

Each edit changes the single boundary line, e.g. `Code/commits/security: write normal.` becomes `Code/commits/security/issue+PR+MR titles+descriptions: write normal prose.`

## Testing

- `node --test tests/installer/*.test.mjs` — 51 pass, 0 fail
- `node --check` on all edited JS hooks — clean

PR #411, opened 2026-05-19
feat(codebuddy): add CodeBuddy Code as first-class agent target

studyzy:
## Summary

Adds full CodeBuddy Code support — identical to Claude Code in plugin architecture, hook system, and installer flow.

Closes #408

### Core implementation (`feat` commit)
- Add `.codebuddy-plugin/` with `plugin.json` + `marketplace.json`
- Add `CODEBUDDY.md` (auto-discovered maintainer instructions)
- Add `codebuddy` provider to `PROVIDERS` array in `bin/install.js`
- Add `installCodebuddy()` mirroring `installClaude()`
- Make `installHooks` / `installMcpShrink` accept provider param
- Add codebuddy to `uninstall()` (plugin, MCP, hooks, settings)
- Add `getAgentConfigDir()` to `caveman-config.js` — detects runtime via `CODEBUDDY_PLUGIN_ROOT` / `CODEBUDDY_CONFIG_DIR` env vars
- Update all hook scripts (activate, mode-tracker, stats, statusline.sh, statusline.ps1) to use `getAgentConfigDir()`

### Tests
- `tests/test_codebuddy_hooks.js` (13 cases): `getAgentConfigDir` priority, activate/mode-tracker/statusline under `CODEBUDDY_CONFIG_DIR`, `CODEBUDDY_PLUGIN_ROOT` fallback, isolation from Claude dir
- `tests/installer/codebuddy.test.mjs` (6 cases): `--list`, dry-run, fresh install, idempotency, uninstall cleanup, Claude dir isolation

### Documentation (`docs` commit)
- `INSTALL.md`: add to install table, update flags, uninstall, troubleshooting, and privacy sections
- `README.md`: add to intro and auto-activate list
- `CONTRIBUTING.md`: add to quick orientation

## Testing

All 19 new test cases pass. Existing test suite unaffected (pre-existing 2 failures in `test_caveman_init.js` are unrelated — agent count mismatch from prior commits).

```
$ node tests/test_codebuddy_hooks.js
13 passed, 0 failed

$ node --test tests/installer/codebuddy.test.mjs
# tests 6, pass 6, fail 0
```

## Dry-run verification

```
$ node bin/install.js --only codebuddy --dry-run --non-interactive
→ CodeBuddy Code detected
  would run: codebuddy plugin marketplace add JuliusBrussee/caveman
  would run: codebuddy plugin install caveman@caveman
  → installing hooks (--with-hooks)
  would mkdir -p ~/.codebuddy/hooks
[...]

PR #285, opened 2026-04-25
Fix caveman mode persistence and improve plan mode handling

beatboxes:
## Summary

This PR fixes caveman mode persistence issues by refactoring hook outputs to use structured `hookSpecificOutput` JSON payloads instead of plain stdout, and adds explicit per-turn reinforcement for plan mode to prevent drift in the model's behavior.

## Key Changes

- **Structured SessionStart output**: `caveman-activate.js` now emits canonical `hookSpecificOutput` JSON with `hookEventName: "SessionStart"` and `additionalContext` containing the ruleset, making the contract explicit and forward-compatible with stricter hook validation.

- **Per-turn reinforcement with level-specific rules**: `caveman-mode-tracker.js` now emits structured `UserPromptSubmit` output on every turn when caveman mode is active (except independent modes). The reinforcement includes:
  - Active level's intensity-table row and example from SKILL.md
  - Explicit plan-mode clause stating that plan documents stay caveman (addressing the headline regression)
  - Priority sentence to ensure the instruction lands ahead of competing defaults

- **Extracted rule parsing logic**: New `extractLevelRules()` and `extractLevelSummary()` helpers in `caveman-config.js` centralize SKILL.md parsing so both hooks stay in sync as the table format evolves.

- **Clarified Auto-Clarity boundaries**: Updated SKILL.md and fallback rulesets to:
  - Distinguish "irreversible multi-command shell sequences" (where fragment order risks data loss) from plan mode (read at leisure, no execution risk)
  - Explicitly state plan mode is NOT a clarity exception
  - Add plan-mode boundary rules for structured sections (steps, paths, commands, code blocks) vs. chat-side replies

- **Comprehensive test coverage**: Added `HookOutputShapeTests` class with 6 new tests validating:
  - SessionStart emits canonical JSON with level-specific rules
  - "off" mode still emits plain "OK" (no-op signal)
  - UserPromptSubmit includes level-specific reinforcement
  - Plan mode clause is present in reinforcement
  - Independent modes skip reinforcement
  - Missing flag skips reinforcement

## Implementation Details

- The `emit()` function in `caveman-activate.js` wraps output once at the bottom to ensure statusline nudge concatenation works correctly.
- Per-turn reinforcement is intentionally short (intensity row + example + override sentence) to stay within context budgets while remaining level-specific.
- The fallback ruleset (when SKILL.md is unavailable) now includes the plan-mode boundary rules inline.
[...]

PR #501, opened 2026-06-08
fix(hooks): load SKILL.md from plugin layout

### LINKED ISSUE #467: Claude Code SessionStart hook serves weak fallback — src/hooks/ SKILL.md path off by one

chrisvaillancourt:
## Summary

- Load `skills/caveman/SKILL.md` from the current plugin `src/hooks` layout before falling back to the legacy flat hook layout.
- Gate the two-level plugin probe to real `src/hooks` installs so standalone hooks under `$CLAUDE_CONFIG_DIR/hooks` cannot accidentally read unrelated `~/skills` content.
- Add regression tests that verify plugin layout loads the real skill file, cwd-local skills are ignored, and standalone layout still falls back when only home-level skills exist.

## Root Cause

The Claude Code plugin manifest runs `${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js`, so `__dirname` is `<plugin_root>/src/hooks`. The hook still looked for `../skills/caveman/SKILL.md`, which resolves to `<plugin_root>/src/skills/caveman/SKILL.md`. That file does not exist, so the hook silently served the hardcoded fallback instead of the source-of-truth skill body.

That fallback is missing the intensity table and wenyan instructions, so plugin installs did not get the full behavior documented in `skills/caveman/SKILL.md`.

## Validation

- `uv run python -m unittest tests.test_hooks`
- `node src/hooks/caveman-activate.js | rg '## Intensity|Current level:'`
- `git diff --check`
- `uv run python -m unittest tests.test_compress_safety`
- `node tests/test_symlink_flag.js`

Additional documented checks were run and failed for existing baseline issues unrelated to this change:

- `npm test` fails in OpenCode installer tests (`caveman-compress.md` missing / temp OpenCode files absent / JSONC parse expectation). This same failure was observed on clean upstream `main` before this branch.
- `node tests/test_caveman_init.js` fails count-based assertions around current init targets/OpenClaw handling. This same failure was observed on clean upstream `main`.

## Review Notes

Additional review found two hardening gaps in the first draft: the ungated `../../skills` probe could escape standalone hook installs into `~/skills`, and the plugin-layout test used `cwd=plugin_root`. This version gates the plugin probe to `src/hooks`, runs the regression from an unrelated cwd with a sentinel cwd skill, and adds a standalone-home sentinel test.

## Overlap

Several open PRs touch `src/hooks/caveman-activate.js`, including #483, #448, #411, #407, #401, and #377. This PR keeps the scope to the SKILL.md path lookup and its regression test so it can be rebased independently. Older dirty PRs such as #285 and #269 touch the pre-`src` hook path and may need the same path logic if revived.

Fixes #467

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -48,14 +48,26 @@ if (INDEPENDENT_MODES.has(mode)) {
 const modeLabel = mode === 'wenyan' ? 'wenyan-full' : mode;
 
 // Read SKILL.md — the single source of truth for caveman behavior.
-// Plugin installs: __dirname = <plugin_root>/hooks/, SKILL.md at <plugin_root>/skills/caveman/SKILL.md
-// Standalone installs: __dirname = $CLAUDE_CONFIG_DIR/hooks/, SKILL.md won't exist — falls back to hardcoded rules.
+// Plugin layout: hook at <plugin_root>/src/hooks/, SKILL.md two levels up at
+// <plugin_root>/skills/caveman/SKILL.md.
+// Legacy/flat layout: hook one level below skills/.
+// Standalone installs ($CLAUDE_CONFIG_DIR/hooks/) ship no SKILL.md — both
+// probes miss and fall through to the hardcoded fallback below.
 let skillContent = '';
-try {
-  skillContent = fs.readFileSync(
-    path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
-  );
-} catch (e) { /* standalone install — will use fallback below */ }
+const skillRoots = [];
+if (path.basename(__dirname) === 'hooks' && path.basename(path.dirname(__dirname)) === 'src') {
+  skillRoots.push(['..', '..', 'skills']);
+}
+skillRoots.push(['..', 'skills']);
+
+for (const rel of skillRoots) {
+  try {
+    skillContent = fs.readFileSync(
+      path.join(__dirname, ...rel, 'caveman', 'SKILL.md'), 'utf8'
+    );
+    break;
+  } catch (e) { /* try next candidate layout */ }
+}
 
 let output;
 
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -1,5 +1,6 @@
 import json
 import os
+import shutil
 import subprocess
 import tempfile
 import unittest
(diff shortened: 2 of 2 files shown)


JuliusBrussee:

PR #511, opened 2026-06-11
fix: clean up temp files leaked by failed flag renames

M3EMO:
## Bug

`safeWriteFlag` in `src/hooks/caveman-config.js` writes the mode flag via an atomic temp + rename (`.caveman-active.<pid>.<timestamp>` → `.caveman-active`). On Windows, `renameSync` over an existing file fails with `EPERM`/`EBUSY` whenever another process holds the target open — the statusline script reading the flag, or a second hook instance writing concurrently. The silent-fail `catch` (correct per the hook rules) swallows the error, but the temp file is never removed.

Net effect: one orphaned temp file per failed rename, accumulating indefinitely in `$CLAUDE_CONFIG_DIR`. Observed in normal daily use on Windows 11: **28 orphans in two weeks**.

## Fix

Two parts, both silent-fail per the hook rules in CLAUDE.md:

1. **`safeWriteFlag` cleans up after itself.** The write + rename now runs inside a `try/finally`; if the rename did not complete, the temp is unlinked. No new failure modes — a failed unlink is swallowed like every other fs error.

2. **`sweepOrphanTemps(flagDir)`** removes temps left by older versions or hard crashes mid-write. Called once per session from `caveman-activate.js` (SessionStart). Deletion criteria are conservative:
   - file older than 24h, **or**
   - owning PID is dead **and** the file is >60s old (grace window so an in-flight write from a live concurrent hook is never touched).

   The strict pattern `^\.caveman-active\.(\d+)\.(\d+)$` can never match the live flag (`.caveman-active`, no suffix). Symlinks matching the pattern are skipped (consistent with the symlink-safety posture of the rest of the module). Nonexistent dirs and fs errors silent-fail.

## Tests

`tests/test_temp_leak.js` (follows `tests/test_symlink_flag.js` conventions), 9 cases:

- successful write leaves no temps
- failed rename (simulated `EPERM`) leaks nothing and leaves the original flag intact
- sweep removes >24h orphans and dead-PID orphans past the grace window
- sweep keeps fresh live-PID temps, dead-PID temps inside the grace window, the live flag, and unrelated files
- sweep skips symlinks matching the temp pattern
- sweep silent-fails on a nonexistent directory

```
9 passed, 0 failed
```

`tests/test_symlink_flag.js` is unchanged by this PR: 5 passed / 7 failed both on `main` and on this branch in my environment — the 7 failures are `EPERM: operation not permitted, symlink` (Windows without Developer Mode cannot create symlinks), pre-existing and unrelated.

## Repro (Windows)

1. Configure the caveman statusline (reads the flag frequently) and use Claude Code normally for a few days, or run two hook instances concurrently.
2. `ls $CLAUDE_CONFIG_DIR/.caveman-active.*` → orphaned temps accumulate.

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -9,11 +9,14 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
-const { getDefaultMode, safeWriteFlag } = require('./caveman-config');
+const { getDefaultMode, safeWriteFlag, sweepOrphanTemps } = require('./caveman-config');
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
 const flagPath = path.join(claudeDir, '.caveman-active');
 const settingsPath = path.join(claudeDir, 'settings.json');
+
+// Once per session: clear temp files orphaned by failed atomic-rename writes.
+sweepOrphanTemps(claudeDir);
 
 const mode = getDefaultMode();
 
--- a/src/hooks/caveman-config.js
+++ b/src/hooks/caveman-config.js
@@ -130,17 +130,67 @@ function safeWriteFlag(flagPath, content) {
     const tempPath = path.join(realFlagDir, `.caveman-active.${process.pid}.${Date.now()}`);
     const O_NOFOLLOW = typeof fs.constants.O_NOFOLLOW === 'number' ? fs.constants.O_NOFOLLOW : 0;
     const flags = fs.constants.O_WRONLY | fs.constants.O_CREAT | fs.constants.O_EXCL | O_NOFOLLOW;
-    let fd;
+    // On Windows, renameSync over an existing file fails with EPERM/EBUSY when
+    // another process holds the target open (statusline read, concurrent hook).
+    // Without the cleanup below, every failed rename leaks one temp file.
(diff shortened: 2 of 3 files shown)


HetCreep:
Related edge case worth noting alongside the orphan-cleanup fix:

The temp path `\config.js:181\`:

`javascript
const tempPath = path.join(realFlagDir, \.caveman-active.\.\\`);
`

PID+timestamp is unique across separate processes (different PIDs), but if the **same process** calls `safeWriteFlag` twice within the same millisecond (e.g. two rapid slash-command triggers in a single hook invocation), `Date.now()` returns the same value and both writes target the same temp path. The `O_EXCL` flag on the open would cause the second write to throw - caught silently, so the second flag write is silently dropped.

Low probability in practice (requires sub-millisecond re-entrancy in a single process), but the fix is cheap: swap `Date.now()` for `crypto.randomUUID()` or `crypto.randomBytes(8).toString('hex')`. No PID needed if the name is truly random.

`javascript
const { randomBytes } = require('crypto');
const tempPath = path.join(realFlagDir, .caveman-active..tmp);
`

This also makes the orphan pattern in `sweepOrphanTemps` slightly simpler to match.

JuliusBrussee:

PR #515, opened 2026-06-12
fix: feat: Auto-Clarity exception for the final task answer (clear prose for the turn-closing summary) (closes #510)

### LINKED ISSUE #510: feat: Auto-Clarity exception for the final task answer (clear prose for the turn-closing summary)
Caveman is perfect for the intermediate messages an agent emits while working (status notes, quick acks between tool calls): that is where the token volume is. But the turn-closing summary, the one message the user actually reads to understand what happened, carries all the comprehension load. Saving ~75% there costs more in re-reading than it saves in tokens.

Proposal: extend Auto-Clarity in `skills/caveman/SKILL.md` with one more drop case:

> Also drop caveman for FINAL TASK ANSWER: the turn-closing message that delivers results, summary, or explanation after work done. User reads only that message to understand outcome: write it in normal clear prose, full sentences, explain properly. Caveman stay for intermediate progress notes, status lines, quick acks between tool calls.

**Before** (turn-closing summary, `full`):

> Bug in auth middleware. Token expiry check use `<` not `<=`. Edge: token accepted at exact expiry second. Fix pushed, tests green. Side effect: sessions live 1s longer, no security impact.

**After**:

> Found it: the auth middleware checks token expiry with `<` instead of `<=`, so a token is still accepted at the exact expiry second. Fix is pushed and tests are green. Side effect: sessions now live one second longer, which has no security impact.

Why better: intermediate chatter keeps the full token cut, and the single message per task that a human reads stays clear. If you prefer it opt-in, it could also be a config flag (e.g. `"clearFinalAnswer": true` next to `defaultMode`).

I run this as a local patch and it works well. Happy to send a PR if you are open to it (one paragraph in `skills/caveman/SKILL.md`).

sseshachala:
Fixes #510

---
This fix was implemented autonomously by **[Conduct AI](https://conductai.ai)** — an open-source platform that turns AI agents into reusable team automations via YAML playbooks.
- 🔗 Platform: [conductai.ai](https://conductai.ai)
- 📦 Source: [github.com/sseshachala/conductai](https://github.com/sseshachala/conductai)
> Want Conduct AI to automate fixes in your repo too? [Get started free →](https://conductai.ai)

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -59,6 +59,7 @@ Drop caveman when:
 - Multi-step sequences where fragment order or omitted conjunctions risk misread
 - Compression itself creates technical ambiguity (e.g., `"migrate table drop column backup first"` — order unclear without articles/conjunctions)
 - User asks to clarify or repeats question
+- FINAL TASK ANSWER: the turn-closing message that delivers results, summary, or explanation after work done. User reads only that message to understand outcome: write it in normal clear prose, full sentences, explain properly. Caveman stay for intermediate progress notes, status lines, quick acks between tool calls.
 
 Resume caveman after clear part done.
 


JuliusBrussee:
Thanks @sseshachala — making every final answer non-Caveman would silently disable the mode on its main output surface. Closing this contradictory exception.

PR #527, opened 2026-06-14
feat(cavecrew): add per-agent model overrides

### LINKED ISSUE #521: feat: per-agent model override for cavecrew subagents (reviewer hardcoded to haiku, too weak as a review gate)
## Problem

The cavecrew subagents pin `model:` in their frontmatter — `cavecrew-reviewer` is `model: haiku`, and builder/investigator likewise. Haiku is the right *default* (cheap, fast), but when cavecrew-reviewer is wired in as a real pre-merge / pre-push review gate, haiku misses semantic bug-classes that a stronger model catches (subtle invariant violations, cross-file reasoning, "this looks fine but is wrong" cases). Users who depend on the reviewer for correctness want to opt it up to `sonnet`/`opus` **per agent**, without touching the others.

There is currently no clean way to do that. Every override path is bad:

| Path | Why it fails |
|---|---|
| Shadow agent in `~/.claude/agents/<name>.md` | Same-`name:` is a **whole-file replacement**, not a field merge. To change only `model:` you must copy the plugin's entire system-prompt body — which then **freezes and drifts stale** every time the plugin updates the reviewer (the body has already changed once: `25d22f864ad6` → `655b7d9c5431`). You silently lose upstream review improvements. |
| `CLAUDE_CODE_SUBAGENT_MODEL` env | Global — bumps **every** subagent and main-thread delegation, not just the reviewer. Too broad; pays opus/sonnet cost everywhere. |
| Per-call `model` param on the Task tool | Only applies when the agent is **explicitly** invoked; no effect on auto-delegation. Also can't be set once and forgotten. |

## Request

A per-agent model override that leaves the plugin-maintained body live (so it keeps getting upstream improvements). Any of:

1. **Env vars** read by the SessionStart hook, e.g. `CAVECREW_REVIEWER_MODEL`, `CAVECREW_BUILDER_MODEL`, `CAVECREW_INVESTIGATOR_MODEL`. The hook rewrites only the `model:` line of the generated agent files at install/session start.
2. **A small config block** (e.g. in the existing settings the hook already reads) mapping agent name → model.
3. At minimum, **document** the supported override and pick a default that's a deliberate quality/cost trade rather than always-haiku for the reviewer.

## Why this is the one agent worth it

The reviewer is the cavecrew agent whose output users most depend on for *correctness*, and the one where a model bump most changes results. Builder/investigator tolerate haiku far better. A field-level model knob (vs. all-or-nothing shadow/env) is the missing piece.

## Related

- #473 — also surfaces the hardcoded-model problem, from the portability angle (the pinned Anthropic model breaks the Gemini CLI backend). A per-agent model config would help there too.

Happy to PR the SessionStart-hook env-var approach if that's the preferred direction.

Ijtihed:
Closes #521

## What

Add env-var model overrides for cavecrew subagents. Patch only the installed agent frontmatter at SessionStart ΓÇö prompt body untouched, keeps receiving upstream updates.

```sh
export CAVECREW_REVIEWER_MODEL=sonnet   # reviewer on sonnet
export CAVECREW_INVESTIGATOR_MODEL=sonnet
export CAVECREW_BUILDER_MODEL=opus
```

## Files

- `src/hooks/cavecrew-model-overrides.js` ΓÇö pure helper, exported functions
- `src/hooks/caveman-activate.js` ΓÇö best-effort call at SessionStart
- install/uninstall manifests (sh + ps1) ΓÇö add new file to HOOK_FILES
- `tests/test_cavecrew_model_overrides.js` ΓÇö 24 tests, all pass
- `skills/cavecrew/README.md` ΓÇö model overrides section

## Notes

- Plugin installs only ΓÇö standalone has no agents dir, silent no-op
- Override persists in installed file until plugin update/reinstall
- No new config schema ΓÇö env vars only, as requested in issue


Ijtihed:
@JuliusBrussee thoughts?

Ijtihed:
:)

Issue #492, opened 2026-06-04
Gemini compatibility issue with Caveman


Ya-Pondo:
**What happen**

Invalid tool name (compatibility issue with Gemini)

**Expected**

Debug the following files
\caveman\agents\cavecrew-builder.md
\caveman\agents\cavecrew-investigator.md
\caveman\agents\cavecrew-reviewer.md
**Before/after example**
```
Input: gemini extension install
Got:  
[ExtensionManager] Error loading agent from caveman: Failed to load agent from C:\Users\XXX\.gemini\extensions\caveman\agents\cavecrew-builder.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
tools.1: Invalid tool name
tools.2: Invalid tool name
tools.3: Invalid tool name
tools.4: Invalid tool name
[ExtensionManager] Error loading agent from caveman: Failed to load agent from C:\Users\XXX\.gemini\extensions\caveman\agents\cavecrew-investigator.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
tools.1: Invalid tool name
tools.2: Invalid tool name
tools.3: Invalid tool name
[ExtensionManager] Error loading agent from caveman: Failed to load agent from C:\Users\XXX\.gemini\extensions\caveman\agents\cavecrew-reviewer.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
tools.1: Invalid tool name
tools.2: Invalid tool name
Failed to clone Git repository from https://github.com/wshobson/agents Error: spawn git ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:483:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
```

**Platform**
- [ ] Claude Code
- [ ] Codex
- [X] Other: __Gemini_

**Version / install method**
gemini extensions install (latest version)

JuliusBrussee:
Thanks for the Gemini error log and affected-file list. This is the same provider-specific cavecrew tool-name incompatibility tracked in #373. Closing this duplicate so the Gemini schema transform and runtime proof remain in one thread.

Issue #473, opened 2026-05-29
cavecrew subagents fail to load in Gemini CLI extension (invalid tool names + Anthropic model)


zackchiutw:
## Summary

When caveman is installed as a **Gemini CLI extension**, all three `cavecrew` subagents fail to load. Gemini CLI rejects the agent definitions because their frontmatter uses **Claude Code tool names** (`Read`, `Edit`, `Write`, `Grep`, `Glob`, `Bash`) and an **Anthropic model** (`model: haiku`), neither of which is valid in Gemini CLI.

The core caveman compression modes still work — only the `cavecrew` multi-agent delegation feature is broken on Gemini.

## Environment

- caveman Gemini extension: release tag `v1.8.2` (note: `gemini-extension.json` still says `version: 1.0.1` — minor aside)
- Gemini CLI: `0.44.1`
- OS: Windows 11 (also applies to macOS/Linux — this is platform-independent validation)

## Reproduction

```
gemini extensions install https://github.com/JuliusBrussee/caveman
gemini extensions list
```

Output:

```
[ExtensionManager] Error loading agent from caveman: Failed to load agent from
  .../.gemini/extensions/caveman/agents/cavecrew-builder.md:
  Validation failed: Agent Definition:
    tools.0: Invalid tool name
    tools.1: Invalid tool name
    ...
[ExtensionManager] Error loading agent from caveman: ...cavecrew-investigator.md: ... Invalid tool name ...
[ExtensionManager] Error loading agent from caveman: ...cavecrew-reviewer.md: ... Invalid tool name ...
```

## Root cause (verified against gemini-cli source)

1. The Gemini extension loader reads agents from a **hardcoded** path — `path.join(effectiveExtensionPath, "agents")` — with no config option to relocate it. So the extension always loads the repo-root `agents/` directory verbatim.

2. Gemini validates each entry in an agent's `tools:` array with `isValidToolName(name, { allowWildcards: true })`, which only accepts:
   - `ALL_BUILTIN_TOOL_NAMES` (Gemini's own tool ids),
   - `TOOL_LEGACY_ALIASES` (currently empty),
   - the `*` wildcard, MCP/discovered tool prefixes.

   Gemini's builtin tool ids differ from Claude Code's. The relevant mapping:

   | Claude Code (current) | Gemini CLI |
   |---|---|
   | `Read`  | `read_file` |
   | `Edit`  | `replace` |
   | `Write` | `write_file` |
[...]

Issue #373, opened 2026-05-12
Compatibility Issue with Gemini CLI: 'Invalid tool name' errors when loading agents


xilu0:
When using Gemini CLI (v0.41.2, darwin), the ExtensionManager fails to load the agents in the caveman extension.

**Error Log:**
```
[ExtensionManager] Error loading agent from caveman: Failed to load agent from ~/.gemini/extensions/caveman/agents/cavecrew-builder.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
tools.1: Invalid tool name
tools.2: Invalid tool name
tools.3: Invalid tool name
tools.4: Invalid tool name
```

**Observations:**
The agents (`cavecrew-builder.md`, `cavecrew-investigator.md`, `cavecrew-reviewer.md`) define their tools using Claude Code's naming conventions (e.g., `[Read, Edit, Write, Grep, Glob]`).
Gemini CLI's ExtensionManager validates these tool names. Since Gemini CLI uses different tool names (e.g., `read_file`, `replace`, `write_file`, `grep_search`, `glob`), it rejects these agents as invalid.

**Expected Behavior:**
The extension should provide compatible tool names for the Gemini CLI environment or include a mapping layer.

**Environment:**
- macOS (darwin)
- Gemini CLI v0.41.2

ehurrn:
Adding impact & following...

```
 % uname -v
Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:26 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T8132
```

```
 % gemini -v    
0.42.0

```

```
% gemini extensions install https://github.com/JuliusBrussee/caveman
Installing extension "caveman".
This extension will append info to your gemini.md context using GEMINI.md

Agent Skills:

This extension will install the following agent skills:

  * caveman-stats: Show real token usage and estimated savings for the current session. Reads directly from the Claude Code session log — no AI estimation. Triggers on /caveman-stats. Output is injected by the mode-tracker hook; the model itself does not compute the numbers.

    (Source: /var/folders/y_/s_lvq4wx5ks7s7x2fh9t1h480000gn/T/gemini-extensionktu1dG/skills/caveman-stats/SKILL.md) (2 items in directory)

[...]

PR #530, opened 2026-06-15
fix(install): remap cavecrew agent tool names to Gemini ids on install

apollion69:
## Problem

Installing caveman as a **Gemini CLI extension** fails to load all three `cavecrew` subagents:

```
[ExtensionManager] Error loading agent from caveman: Failed to load agent from
  ~/.gemini/extensions/caveman/agents/cavecrew-builder.md:
  Validation failed: Agent Definition: tools.0: Invalid tool name ...
```

Gemini's `ExtensionManager` loads agents from the hardcoded `<ext>/agents/` path and validates every entry in an agent's `tools:` array with `isValidToolName(name, { allowWildcards: true })`, which only accepts Gemini's own builtin tool ids. The shipped `agents/cavecrew-*.md` use **Claude Code** tool names, so all entries are rejected and the agents never load. `model: haiku` is also Anthropic-only and fails at runtime.

Reproduces on every platform (verified against `@google/gemini-cli@0.44.1`). Tracked in #373, #473, #492.

## Why it's awkward

`agents/cavecrew-*.md` is a single source of truth **shared with the Claude Code plugin** (`.claude-plugin/plugin.json` → `source: "./"`), which needs the Claude names. The two tool namespaces are disjoint, so the source files can't satisfy both — they must stay Claude-canonical.

## Fix

Rewrite the **installed copy** at Gemini-install time, exactly mirroring the opencode path's `stripOpencodeAgentTools`:

- New `bin/lib/gemini-agent.js` → `mapGeminiAgentFrontmatter()` maps tool names and `model: haiku`.
- `installGemini()` post-processes `~/.gemini/extensions/caveman/agents/*.md` after a successful `gemini extensions install`.
- **Source `agents/*.md` are untouched** — Claude plugin behavior is unchanged.

Tool-name map (verified against the gemini-cli `ALL_BUILTIN_TOOL_NAMES`):

| Claude | Gemini |
|---|---|
| `Read` | `read_file` |
| `Edit` | `replace` |
| `Write` | `write_file` |
| `Grep` | `grep_search` |
| `Glob` | `glob` |
| `Bash` | `run_shell_command` |

`model: haiku` → `gemini-2.5-flash` (sonnet/opus → `gemini-2.5-pro`). Unknown tool names pass through unchanged.

## Tests

`tests/installer/gemini-agent.test.mjs` — 10 new tests (inline-array mapping, `Bash`→`run_shell_command`, folded `description: >` preservation, unknown-name passthrough, non-string defensive, RED proof the shipped files carry Claude names today, GREEN proof they transform to Gemini-valid ids with byte-identical body, installer-equivalent end-to-end).

```
node --test tests/installer/*.test.mjs   ->  83 pass / 0 fail
```

## Known limitation

The patch runs through `bin/install.js`. A direct `gemini extensions install <url>` (without the caveman installer) won't be patched, and a Gemini auto-update re-clones the source tree and reverts it — same constraint as the opencode mirror. A fuller fix (CI-generated Gemini agent mirror) is noted in #473 option 1; this keeps the change small and focused per CONTRIBUTING.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -25,7 +25,7 @@ const crypto = require('crypto');
 const SETTINGS = require('./lib/settings');
 const OPENCLAW = require('./lib/openclaw');
 const { stripOpencodeAgentTools } = require('./lib/opencode-agent');
-const { mapGeminiAgentFrontmatter } = require('./lib/gemini-agent');
+const { mapGeminiAgentFrontmatter, geminiExtAgentsDir } = require('./lib/gemini-agent');
 
 const REPO = 'JuliusBrussee/caveman';
 // Pin remote fetches to an immutable release tag, not the moving `main`
@@ -529,6 +529,11 @@ function installGemini(ctx) {
     const r = captureSpawn('gemini', ['extensions', 'list']);
     if (r.status === 0 && /caveman/i.test(r.stdout || '')) {
       note('  caveman extension already installed (use --force to reinstall)');
+      // Still rewrite the installed agents to Gemini tool ids: upgraders who
+      // installed before the #473 fix have the broken Claude tool names on
+      // disk and would never get the fix without --force. The rewrite is
+      // idempotent (only writes when the frontmatter actually changes).
+      patchGeminiAgents(ctx);
       results.skipped.push(['gemini', 'extension already installed']);
       process.stdout.write('\n');
       return;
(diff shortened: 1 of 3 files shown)


apollion69:
Thanks for the review — both P2s fixed in `b796bfa`:

**`bin/install.js:559` — honor `GEMINI_CLI_HOME`** ✅
The agents-dir was hard-coded to `os.homedir()/.gemini`, so under the `GEMINI_CLI_HOME` override the rewrite skipped as "install dir not found" and left the rejected Claude tool names on disk. Extracted `geminiExtAgentsDir()` into `lib/gemini-agent.js` using `process.env.GEMINI_CLI_HOME || os.homedir()`, mirroring the existing `SETTINGS.claudeConfigDir` / `CLAUDE_CONFIG_DIR` pattern. Unit-tested both branches (override + fallback).

**`bin/install.js:540` — patch existing installs** ✅
The already-installed branch returned before `patchGeminiAgents` ran, so anyone who installed before the #473 fix never got the rewrite without `--force`. It now runs in that branch too — and since the rewrite only writes when the frontmatter actually changes, it's a safe idempotent no-op for already-correct installs.

`npm test`: 85 pass (was 83), `node --check` clean on both files.

apollion69:
Friendly ping — this is rebased on `main` and still applies cleanly. Happy to adjust the Gemini tool-id mapping if you'd prefer a different naming convention; otherwise it's ready whenever you have a moment to review.

PR #532, opened 2026-06-15
fix(mode-tracker): honor off default in per-turn reinforcement

triggrr-io:
## Problem

`CAVEMAN_DEFAULT_MODE=off` (env) / `defaultMode: "off"` (config) makes the **SessionStart** hook skip activation and delete the flag. But the **UserPromptSubmit** per-turn reinforcement in `src/hooks/caveman-mode-tracker.js` read only the global flag file (`~/.caveman-active`) and never consulted the resolved default mode.

Because that flag is **global** (shared across all sessions/repos), a session in any other repo where caveman is on sets it to a non-off mode — and it then bleeds into repos/sessions that explicitly opted out, re-injecting `CAVEMAN MODE ACTIVE (...)` on every turn. A per-repo / per-context off cannot be expressed when the per-turn check is flag-only.

## Fix

Bring the per-turn path in line with SessionStart: if `getDefaultMode() === 'off'`, clear the stale global flag and emit nothing. `getDefaultMode()` already reads env → config → default, so this respects per-context opt-out. No behavior change when caveman is genuinely active.

## Test

- env off + flag=full → no emit, stale flag cleared ✅
- env unset + flag=full → emits as before ✅
- `node --check` passes ✅

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -119,16 +119,28 @@ process.stdin.on('end', () => {
     // If the flag is missing, corrupted, oversized, or a symlink pointing at
     // something like ~/.ssh/id_rsa, readFlag returns null and we emit nothing
     // — never inject untrusted bytes into model context.
-    const activeMode = readFlag(flagPath);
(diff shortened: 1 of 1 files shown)


JuliusBrussee:

PR #534, opened 2026-06-17
fix(caveman-compress): add UTF-8 encoding to all file I/O

### LINKED ISSUE #533: `caveman-compress` crashes on Windows with `charmap` codec error (Unicode characters)
**What happen**

`caveman-compress` crashes with a `charmap` codec error when the target `.md` file contains non-ASCII Unicode characters (e.g. `→` U+2192, em-dash `—`). Python on Windows defaults to `cp1252` encoding instead of UTF-8.

Worse: the output file gets **truncated to 0 bytes** before the error surfaces. The `.original.md` backup is not yet written at that point. Result: original file is wiped with no recovery path.

```
Detected: natural_language
Starting caveman compression...

Processing: C:\path\to\CLAUDE.md
Compressing with Claude...

❌ Error: 'charmap' codec can't encode character '→' in position 4310: character maps to <undefined>
```

**Expected**

- Read and write files with explicit `encoding='utf-8'`
- On any error: report **before** writing — leave original untouched
- Write `.original.md` backup **before** any destructive write

**Before/after example**

```
Input:  .md file containing → (U+2192)
Got:    crash + file wiped to 0 bytes
Want:   either compress successfully (UTF-8 I/O) or fail safely with original intact
```

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**

Installed as Claude Code plugin via superpowers. Invoked with `uv run python -m scripts <filepath>` (cpython-3.14-windows-x86_64). OS: Windows 11 Pro 10.0.26200.

**Suggested fix**

Add `encoding='utf-8'` to all `open()` calls in `scripts/__main__.py`:

```python
# read
with open(filepath, encoding='utf-8') as f:
    content = f.read()

# write
with open(filepath, 'w', encoding='utf-8') as f:
    f.write(compressed)
```

Workaround until fixed: set `PYTHONUTF8=1` env var.

rish106-hub:
Windows defaults to cp1252, crashing on non-ASCII chars (→ —) with "charmap" codec error and truncating the target file to 0 bytes.

- Add encoding='utf-8' to all read_text/write_text calls in compress.py, validate.py, and detect.py
- Add _atomic_write() (temp file + rename) to prevent truncation if a write fails mid-way
- Update test helpers and add unicode round-trip safety test

Closes #533

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -11,6 +11,7 @@ import re
 import shutil
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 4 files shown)


rish106-hub:
@JuliusBrussee Hey! Just opened this PR to fix the Windows encoding crash reported in #533. The issue was that Windows defaults to cp1252, which breaks on non-ASCII characters like → and —. I've added `encoding='utf-8'` to all file I/O calls in `compress.py`, `validate.py`, and `detect.py`, and also added an `_atomic_write()` helper to prevent file truncation on failed writes. Would love your review!

rish106-hub:
@JuliusBrussee Updated with fixes for both Copilot feedback items:

1. `_atomic_write()` now uses `tempfile.mkstemp()` for a unique temp filename 
   — avoids clobbering existing *.tmp files and handles sources that already 
   end in .tmp
2. Restore-on-failure path now calls `_atomic_write()` too — restores from 
   the verified on-disk backup atomically

Ready for review whenever you have time!

JuliusBrussee:

PR #540, opened 2026-06-19
fix: add version field to plugin.json

zeral189:
CLI tools and plugin managers use the `version` field to display installed versions. Without it, `plugin.json` shows `vundefined`. This adds `"version": "1.9.0"` to match the release tag.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,5 +1,6 @@
 {
   "name": "caveman",
+  "version": "1.9.0",
   "description": "Ultra-compressed communication mode. Cuts ~75% of tokens while keeping full technical accuracy by speaking like a caveman.",
   "author": {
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
**Adopted in spirit, superseded by #1015** — you were right about the defect, thank you for filing it.

Two reasons the branch didn't take it as-is: the literal `1.9.0` has since gone stale (the release is 2.6.0), so it would have reproduced the same drift a few releases later; and the Claude manifest was one of three that had drifted — `gemini-extension.json` was at `1.0.1` and the Codex manifest at `0.1.0`.

So #1015 sets the version from the release and adds a build check that fails when any host manifest stops matching `package.json`. You're credited as co-author on the commit.

---

JuliusBrussee:

Issue #445, opened 2026-05-26
caveman: preserve user's language instead of forcing English


thallesdomician:
## Problem
Caveman tends to start in English even when the user is writing in another language.

## Expected
Keep the user's dominant language and only compress style.

## Scope
- no forced English openings/status phrases
- keep technical terms, code, commands, and exact error strings unchanged unless translation is requested

I already have a small patch ready for the skill text and help docs.

PR #541, opened 2026-06-19
fix(hooks): preserve user language in per-turn reminder and fallback ruleset

grigory51:
## What

  Caveman already says "preserve the user's language — compress the style, not the language." But that line lives only in the SessionStart ruleset. Two places
  that run every turn never got the memo:

  - **`caveman-mode-tracker.js`** re-injects a short English reminder on *every* prompt. It's the freshest instruction the model sees, and it's English-only — so
  in long sessions (especially after a `/compact` drops the SessionStart block) replies quietly drift back to English.
  - **`caveman-activate.js`** fallback ruleset (the one used when SKILL.md isn't found) skips the language rule too.

  ## Fix

  One line added to each. Caveman keeps crushing tokens; it just stops switching the language while doing it.

  ## Impact

  Nothing changes for English users. For the rest of us, caveman finally stays in our language. I'm Russian — `/caveman` kept flipping me to English mid-task,
  and this was the leak.

  Two-line diff, hook tests green (`tests/test_hooks.py`).

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -101,6 +101,7 @@ if (skillContent) {
     '## Rules\n\n' +
     'Drop: articles (a/an/the), filler (just/really/basically/actually/simply), pleasantries (sure/certainly/of course/happy to), hedging. ' +
     'Fragments OK. Short synonyms (big not extensive, fix not "implement a solution for"). Technical terms exact. Code blocks unchanged. Errors quoted exact.\n\n' +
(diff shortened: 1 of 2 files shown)


SS-42:
Hit this exact gap independently — the SessionStart ruleset already carries the language rule (merged in f06348c, closing #445), but this per-turn reminder and the no-SKILL.md fallback ruleset were both still English-only, and the reminder is the freshest thing in context on every turn. Confirmed this patch closes both. Applied it locally on top of current main, ran the full suite, all green.

PR #542, opened 2026-06-19
fix(hooks): isolate flag files per session

### LINKED ISSUE #184: Concurrent sessions conflict — .caveman-active flag file is global, not per-session
## Problem

`.caveman-active` is single global file. When multiple Claude Code sessions run simultaneously with different caveman levels, they clobber each other.

Session A sets ultra → writes `~/.claude/.caveman-active` = "ultra"
Session B starts → overwrites file = "full"
Session A statusline now shows wrong mode.

Both `caveman-activate.js` (SessionStart) and `caveman-mode-tracker.js` (UserPromptSubmit) write to same global path.

## Fix

Use `session_id` from hook stdin JSON to write per-session flag files:

```
~/.claude/.caveman-active-{session_id}
```

`session_id` is available in:
- Hook stdin JSON (`data.session_id`)
- Statusline stdin JSON (`.session_id`)

### Changes needed

1. **`caveman-activate.js`** — parse stdin for `session_id`, write `~/.claude/.caveman-active-{session_id}`, cleanup stale files >24h
2. **`caveman-mode-tracker.js`** — extract `session_id` from already-parsed stdin, write per-session file
3. **`caveman-statusline.sh`** — extract `session_id` via jq, read per-session file, fallback to global

## Environment

- macOS, Claude Code CLI
- Multiple terminal sessions with different caveman levels

### LINKED ISSUE #334: Feature request: Per-session mode isolation for tmux users
Hello there, this is a feature I'd love to see implemented for caveman as I really like using it in claude code but would love different tmux panes to use different modes. All of the below is the problem description & proposed solution written by Claude Code:
  
## Problem

  When running multiple Claude Code sessions in tmux panes, all
  sessions share a
  single flag file (~/.claude/.caveman-active). Running `/caveman
  ultra` in one
  pane silently changes the mode in every other pane.

  ## Root Cause

  `caveman-mode-tracker.js` and `caveman-activate.js` both resolve
   flagPath as:

    const flagPath = path.join(claudeDir, '.caveman-active');

  Single global file, no per-session discrimination.

  ## Proposed Fix

  Key flag file by $TMUX_PANE (e.g. %0, %1) when inside tmux:

  // caveman-config.js
  function getFlagPath(claudeDir) {
    const base = path.join(claudeDir, '.caveman-active');
    const pane = process.env.TMUX_PANE;
    if (pane) {
      const safe = pane.replace(/[^a-zA-Z0-9%]/g, '');
      if (safe) return base + '-' + safe;
    }
    return base;
  }

  Replace hardcoded flagPath in caveman-mode-tracker.js and
  caveman-activate.js
  with getFlagPath(claudeDir). Same treatment for
  caveman-statusline.sh with
  a fallback to the global flag if pane-specific file doesn't
  exist.

  ## Behavior After Fix

  - Each tmux pane gets its own flag (.caveman-active-%0,
  .caveman-active-%1)
  - /caveman ultra in pane 1 does not affect pane 2
  - Non-tmux sessions fall back to .caveman-active (unchanged
  behavior)
  - $CAVEMAN_DEFAULT_MODE still works as per-pane override at
  launch time

  ## Environment

  macOS, tmux, multiple Claude Code sessions per window

popematt:
Single global `.caveman-active` caused two bugs:
- Concurrent sessions clobbered each other's mode
- Context compression (compact/resume) reset mid-session changes

Per-session flag keyed by `session_id` from hook stdin. SessionStart skips write on compact/resume if per-session file exists. Global file kept as best-effort fallback for external tools.

UUID validation on all session ID inputs. `safeUnlinkFlag` helper for symlink-safe deletion. Stale per-session files cleaned after 24h.

BREAKING CHANGE: mid-session mode changes (`/caveman ultra`, etc.) no longer update the global `.caveman-active` file. External tools reading that file see the session's startup default, not mid-session changes. Migrate to per-session files via `CLAUDE_CODE_SESSION_ID`.

Fixes #184. Fixes #334.

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -1,143 +1,172 @@
 #!/usr/bin/env node
 // caveman — Claude Code SessionStart activation hook
 //
-// Runs on every session start:
-//   1. Writes flag file at $CLAUDE_CONFIG_DIR/.caveman-active (statusline reads this)
-//   2. Emits caveman ruleset as hidden SessionStart context
-//   3. Detects missing statusline config and emits setup nudge
+// Runs on every session start (startup, resume, compact, clear):
+//   1. Reads stdin JSON for session_id and source
+//   2. Writes per-session flag file (conditional on source)
+//   3. Writes global flag file (best-effort for external tools)
+//   4. Emits caveman ruleset as hidden SessionStart context
+//   5. Detects missing statusline config and emits setup nudge
+//   6. Cleans up stale per-session flag files on fresh startup
 
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
-const { getDefaultMode, safeWriteFlag } = require('./caveman-config');
+const { getDefaultMode, safeWriteFlag, readFlag, getFlagPath, getGlobalFlagPath, safeUnlinkFlag, cleanupStaleFlagFiles } = require('./caveman-config');
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
-const flagPath = path.join(claudeDir, '.caveman-active');
 const settingsPath = path.join(claudeDir, 'settings.json');
 
-const mode = getDefaultMode();
+let input = '';
+process.stdin.on('data', chunk => { input += chunk; });
+process.stdin.on('end', () => {
+  let sessionId = null;
+  let source = 'startup';
+  try {
+    const data = JSON.parse(input);
+    sessionId = data.session_id || null;
+    source = data.source || 'startup';
+  } catch (e) {
+    // No stdin or invalid JSON — proceed with defaults
+  }
 
-// "off" mode — skip activation entirely, don't write flag or emit rules
-if (mode === 'off') {
-  try { fs.unlinkSync(flagPath); } catch (e) {}
-  process.stdout.write('OK');
-  process.exit(0);
-}
+  const mode = getDefaultMode();
+  const perSessionFlagPath = getFlagPath(claudeDir, sessionId);
+  const globalFlagPath = getGlobalFlagPath(claudeDir);
 
-// 1. Write flag file (symlink-safe)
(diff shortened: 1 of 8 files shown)


AmanKRoy:
Nice fix for the concurrent-session clobbering and compact/resume reset — `sanitizeSessionId` UUID check, symlink-safe `safeUnlinkFlag`, and the stale-file cleanup are solid additions.

One gap I noticed diffing this against #634/#532 (both open, both patching the same reinforcement check in `caveman-mode-tracker.js`): this PR's final hunk on that block is

```js
const activeMode = readFlag(flagPath);
if (activeMode && !INDEPENDENT_MODES.has(activeMode)) {
```

— no off-gate at all. #532 and #634 both add a `getDefaultMode(...) !== 'off'` condition here to stop a flag (global or, post-this-PR, per-session) from re-injecting "CAVEMAN MODE ACTIVE" into a repo/session that's explicitly configured off. If this merges as-is, that regression comes back regardless of which of #532/#634 lands first or second, since all three touch the identical line.

Given #542 also removes the concurrent-foreign-session motivation behind #532/#634's problem statements, might be worth landing this first and then rebasing the off-gate on top — but happy to help reconcile whichever order the maintainer prefers. Also flagging it shows `CONFLICTING`/`dirty` against current `main`, so a rebase is needed either way before it's mergeable.

AmanKRoy:
@JuliusBrussee no rush at all, know you've got a big queue — just flagging this one since it touches the same hooks as #634/#532 (which I opened/reviewed), so wanted to make sure it's on the radar in case a rebase against `main` would help move it along. Happy to help with that reconciliation if useful, otherwise no worries either way!

Also still fairly new to contributing to open source, so apologies if I've missed any convention or overstepped anywhere — happy to adjust.

JuliusBrussee:
Thanks @popematt — mode state is now isolated per session and hardened for concurrency. Closing as implemented.

PR #554, opened 2026-06-23
Fix: CJK text immediately before a tool call can truncate the turn

GhostXia:
## Problem

When caveman is active and the user works in a **CJK language** (Chinese/Japanese/Korean), turns that include **tool calls** are frequently cut off at the tool-call boundary. The visible truncation always spans from a garbled fragment of the tool-call wrapper through `</invoke>` — i.e. the structured tool-call markup itself is corrupted/truncated, not the prose.

Key signals from a user report:

- Pure **English** workflows never truncate.
- Only **CJK + caveman + tool calls** truncates, with high frequency.
- A user-confirmed workaround — *work/narrate in English, write the CJK summary only after tools finish* — eliminated the truncation entirely.

## Root cause (best current understanding)

CJK characters placed immediately before a tool call abut the structured tool-call markup; the model's tool-call emission destabilizes at that boundary and the turn is cut. This is fundamentally a model/client tool-call-emission interaction with multibyte text — but **caveman is the trigger**, because it pushes the model to interleave terse CJK prose tightly against tool calls.

## Fix

Encode the confirmed workaround as a rule, so users do not have to discover/repeat it:

- `skills/caveman/SKILL.md` — never place CJK text immediately before a tool call; narrate around tools in English (or omit), write the CJK summary only after all tool calls finish.
- `src/hooks/caveman-mode-tracker.js` — same rule appended to the per-turn reinforcement, so it survives context compaction in long sessions.

Mitigation is from the skill side (steering prose/tool-call layout); it does not require any client change.

## Repro

1. Enable caveman (`full`), converse in Chinese.
2. Ask for a task needing several tool calls (write a file, run a command).
3. Observe the turn cut off at the tool-call boundary: CJK prose immediately followed by a truncated tool call.
4. Apply this patch (or the English-narration workaround) → no more truncation.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -24,6 +24,8 @@ Preserve user's dominant language. User write Portuguese → reply Portuguese ca
 
 No self-reference. Never name or announce the style. No "caveman mode on", "me caveman think", no third-person caveman tags. Output caveman-only — never normal answer plus "Caveman:" recap. Exception: user explicitly ask what the mode is.
 
+CJK + tool calls: never place CJK (Chinese/Japanese/Korean) text **immediately before a tool call** or structured XML/code block. CJK characters abutting tool-call markup can corrupt or truncate the emitted call, cutting the turn off at the tool boundary. When a turn uses tool calls, keep surrounding narration in English (or omit it); write the CJK summary only **after** all tool calls finish.
+
 Pattern: `[thing] [action] [reason]. [next step].`
 
 Not: "Sure! I'd be happy to help you with that. The issue you're experiencing is likely caused by..."
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -126,7 +126,9 @@ process.stdin.on('end', () => {
           hookEventName: "UserPromptSubmit",
           additionalContext: "CAVEMAN MODE ACTIVE (" + activeMode + "). " +
             "Drop articles/filler/pleasantries/hedging. Fragments OK. " +
-            "Code/commits/security: write normal."
+            "Code/commits/security: write normal. " +
+            "CJK (Chinese/JP/KR) + tool calls: never put CJK text right before a tool call " +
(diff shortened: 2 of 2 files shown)


GhostXia:
**Alternative / lighter mitigation: newline separator instead of switching to English narration**

Current fix steers the model to narrate in English and write CJK only after tools finish. Works, but degrades UX for CJK users. A lighter variant that may achieve the same result:

> Require a newline (blank line) between any CJK prose and the immediately-following tool call, so the last token before the tool-call markup is ASCII `\n` rather than a multibyte CJK character.

Rationale, consistent with this PR's stated root cause (CJK chars *abutting* the tool-call markup destabilize emission): a newline is a single-byte ASCII token. Inserting it removes the "abutting" condition — the model transitions into the markup from `\n`, not from a multibyte char — while keeping CJK narration.

Trade-offs vs. the current fix:
- Keeps CJK narration (better UX)
- Strictly lighter constraint
- The root cause is "best current understanding" / unconfirmed. If instability triggers on CJK *anywhere near* the markup (not just the immediately-preceding char), a single `\n` will not be enough.
- The English-narration workaround is **user-confirmed**; the newline variant is **untested**. Needs validation against the same repro before relaxing the rule.

**Fresh repro captured in the wild** — CJK prose immediately abutting the tool call, turn truncated, with a garbled wrapper fragment (`count`) appearing where the tool-call markup should start:

```
...生成占位图标源图(1024×1024 PNG, CI 里用 tauri icon 生成全套)。
count
<invoke name="PowerShell">
...
```

Note the stray `count` fragment immediately before the (corrupted) tool-call wrapper — this is the markup-emission corruption the PR describes.

Suggested path: keep this PR as the safe default, but test the newline variant against the repro; if it holds, relax the rule to "CJK + blank line before tool call" instead of English-only. Posting for the maintainer to decide.

JuliusBrussee:
Thanks @GhostXia — forcing English is not a valid fix for an alleged tool-call transport issue and violates language preservation. Closing this mechanism.

PR #563, opened 2026-06-24
fix(caveman-stats): read desktop session logs, CWD-scope CLI, fall through empty transcript

asicmaster:
## Problem

`/caveman-stats` reported **"No conversation yet"** even deep into active sessions. Three independent causes:

### 1. Desktop app sessions are never found (main bug)
The Claude **desktop app** ("host loop") does NOT write the CLI's `~/.claude/projects/<slug>/<uuid>.jsonl` transcript. It writes a live, per-turn audit log at:
```
~/.config/Claude/local-agent-mode-sessions/<acct>/<space>/local_<id>/audit.jsonl
```
`findRecentSession` only scanned `~/.claude/projects/`, so on the desktop app there was no file to read. Verified across **115 audit.jsonl files / 862 result entries** on a live install (`output_tokens`, `cache_read_input_tokens`, `total_cost_usd`, `session_id` present 100%).

### 2. CLI fallback picked the wrong project (latent)
`findRecentSession` walked **all** `~/.claude/projects/**` and picked the globally newest `.jsonl` by mtime — a busy sibling project shadowed the active session.

### 3. Empty transcript_path masked discovery
The UserPromptSubmit hook forwards the harness `transcript_path` as `--session-file`. On the desktop app that path can point at an empty / not-yet-written file (the live session isn't flushed to disk), so trusting it unconditionally produced "No conversation yet" and skipped the audit.jsonl discovery entirely.

## Fix (`src/hooks/caveman-stats.js`)

**Desktop discovery:** `auditFromCwd()` (exact current session via cwd walk-up), `newestDesktopAudit()` (recency fallback), `desktopSessionsRoot()` (`~/.config/Claude`, override `CLAUDE_DESKTOP_DIR`, kept separate from CLI's `CLAUDE_CONFIG_DIR`), `newerFile()` (CLI vs desktop by recency).

**Dual-format `parseSession()`:** desktop `audit.jsonl` carries 1–83 `type=="result"` entries with cumulative per-run usage; the interleaved `type=="assistant"` entries are streaming snapshots (`output_tokens≈1`) that must NOT be summed. Sum the result entries, derive model from `modelUsage` keys (strip `[1m]` marker), count error runs. CLI assistant-summing unchanged.

**Candidate fall-through in `main()`:** build an ordered candidate list (explicit arg → cwd desktop audit → newer of CLI/desktop) and keep the FIRST candidate with `turns > 0`. A real transcript still wins immediately; an empty `transcript_path` no longer hides the live audit log.

**Also:** `priceForModel()` strips `[1m]` before rate-card match; history keyed by result `session_id`; CLI lookup CWD-scoped before the global walk.

## Test
`node tests/test_caveman_stats.js` → **31 passed, 0 failed.** Added desktop result-entry coverage (snapshot exclusion, `[1m]` strip, error runs, `session_id`) and CLI-unchanged assertion; isolated `CLAUDE_DESKTOP_DIR` in the no-session test. Verified through the real hook path (mode-tracker → stats) with an empty `transcript_path` → renders real usage:
```
Turns: 17 · Output: 5,681 · Cache-read: 883,239 · ~$0.264 (claude-opus-4-8)
```

## Known limitation
A claude.ai **remote-control** (web-bridged) session writes no local transcript, so stats fall through to the most recent desktop/CLI session rather than that exact chat. Normal desktop host-loop and CLI sessions both work.

### DIFF
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -54,6 +54,29 @@ function formatUsd(amount) {
 
 function findRecentSession(claudeDir) {
   const projectsDir = path.join(claudeDir, 'projects');
+
+  // Session dirs are flat (UUID.jsonl files directly inside project slug dir).
+  // Try CWD-scoped project first so multi-project setups don't bleed across.
+  // CWD slug: /home/dave → -home-dave  (replace every / with -)
+  function newestInDir(dir) {
+    let names;
+    try { names = fs.readdirSync(dir); } catch { return null; }
+    let best = null;
+    for (const name of names) {
+      if (!name.endsWith('.jsonl')) continue;
+      const p = path.join(dir, name);
+      let st;
+      try { st = fs.statSync(p); } catch { continue; }
+      if (!best || st.mtimeMs > best.mtime) best = { file: p, mtime: st.mtimeMs };
+    }
+    return best ? best.file : null;
+  }
+
+  const cwdSlug = process.cwd().replace(/\//g, '-');
+  const cwdResult = newestInDir(path.join(projectsDir, cwdSlug));
+  if (cwdResult) return cwdResult;
+
+  // Fall back: global walk (original behaviour, covers non-standard CWDs)
   let entries;
   try { entries = fs.readdirSync(projectsDir, { withFileTypes: true }); }
   catch { return null; }
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -40,8 +40,11 @@ const MODEL_OUTPUT_PRICE_PER_M = [
 
 function priceForModel(model) {
   if (!model) return null;
+  // Desktop audit logs tag ids with a context-window marker, e.g.
+  // "claude-opus-4-8[1m]". Strip it before prefix-matching the rate card.
+  const id = model.replace(/\[[^\]]*\]\s*$/, '');
   for (const [prefix, price] of MODEL_OUTPUT_PRICE_PER_M) {
-    if (model.startsWith(prefix)) return price;
+    if (id.startsWith(prefix)) return price;
   }
   return null;
 }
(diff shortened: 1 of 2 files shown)


src/hooks/caveman-stats.js
```diff
@@ -75,28 +101,107 @@ function findRecentSession(claudeDir) {
...
 
+// ── Claude desktop app (Electron) session logs ───────────────────────────
+// The desktop "host loop" does NOT write the CLI's
+//   ~/.claude/projects/<slug>/<uuid>.jsonl
+// transcript. It writes a live, per-turn audit log at
+//   ~/.config/Claude/local-agent-mode-sessions/<acct>/<space>/local_<id>/audit.jsonl
+// Each audit.jsonl holds one or more type=="result" entries carrying
+// cumulative usage for that run; parseSession() prefers them.
+function desktopSessionsRoot() {
+  const base = process.env.CLAUDE_DESKTOP_DIR || path.join(os.homedir(), '.config', 'Claude');
```
AmirF194, line 112: `desktopSessionsRoot()` (src/hooks/caveman-stats.js:111-112) hardcodes `path.join(os.homedir(), '.config', 'Claude')` with no `win32` branch, so on Windows this looks for `<homedir>/.config/Claude/local-agent-mode-sessions` instead of wherever the desktop app's actual per-user data directory is. Nothing in this diff or the rest of the file sets `CLAUDE_DESKTOP_DIR` automatically, so a Windows user gets no error, just a silent fall-through to the old CLI-only lookup, which is exactly the "No conversation yet" bug this PR sets out to fix.

The file's own siblings already have the pattern for this: `caveman-config.js:50-58` and `caveman-activate.js:96-97` both check `process.platform === 'win32'` and fall back to `process.env.APPDATA`. Might be worth the same branch here rather than a Linux/macOS-only default.

Issue #444, opened 2026-05-26
caveman-shrink: nested protected-segment sentinels not restored, enum values lost (e.g. `( 0 )` instead of `(STARTER/BUSINESS)`)


alejandronava562:
## Summary

`caveman-shrink` (v0.1.0) corrupts parenthetical enumeration values in MCP tool descriptions because nested protected-segment sentinels are not recursively restored. The numeric placeholder leaks into the final output, destroying semantic information the model needs.

## Repro

Wrap the Todoist remote MCP server and dump `tools/list`:

```jsonc
{
  "mcpServers": {
    "todoist-shrunk": {
      "command": "npx",
      "args": ["-y", "caveman-shrink", "cmd", "/c", "npx -y mcp-remote https://ai.todoist.net/mcp"]
    }
  }
}
```

### Observed

`list-workspaces` description:

```
Before: ... plan type (STARTER/BUSINESS), user role (ADMIN/MEMBER/GUEST), ...
After:  ... plan type ( 0 ), user role ( 1 ), ...
```

`user-info`:

```
Before: ... user plan (Free/Pro/Business).
After:  ... user plan ( 1 ).
```

The enum values (`STARTER/BUSINESS`, `ADMIN/MEMBER/GUEST`, `Free/Pro/Business`) are lost. From the model's perspective the tool no longer documents its return enums.

### Expected

Either the enum stays intact, or compression is skipped on segments that mix `/` separators with parens. Anything except numeric placeholders surviving into the model-visible output.

## Root cause

In `src/mcp-servers/caveman-shrink/compress.js`, `withProtectedSegments` iterates `PROTECTED_PATTERNS` in order and stores matches in a single `segments[]` array, using ` <index> ` as the sentinel. Two patterns fire on the same span:

1. The **path** pattern `\b[\w.-]*[\/\\][\w.\/\\\-]+` matches `STARTER/BUSINESS` (the `/` qualifies it as a path). → `segments[0] = "STARTER/BUSINESS"`, working text becomes `plan type ( 0 )`.
2. The **function-call** pattern `[A-Za-z_][A-Za-z0-9_]*\s*\([^)]*\)` then matches `type ( 0 )`. → `segments[1] = "type ( 0 )"`, working text becomes `plan  1 `.

Restoration is a **single pass**:

```js
out = out.replace(/ (\d+) /g, (_, i) => segments[+i]);
```

[...]

PR #576, opened 2026-06-26
fix(shrink): skip ARTICLES removal in CJK-mixed text

### LINKED ISSUE #575: [Bug] CJK/Chinese text — ARTICLES regex deletes English articles that carry semantic weight in mixed-language context
## Problem

`compress.js` has zero CJK-awareness. All regex patterns (`ARTICLES`, `FILLERS`, `PLEASANTRIES`, `HEDGES`, `LEADERS`) are English-only, but the `ARTICLES` regex `/\b(?:a|an|the)\s+(?=[a-z])/gi` incorrectly strips English articles from CJK-mixed text where they carry semantic weight.

In Chinese technical conversations, English articles before technical terms are **meaningful** — they signal specificity or quantity:

| Original | After compress | Problem |
|----------|---------------|---------|
| "这是一个 **a** LLM 模型" | "这是一个 LLM 模型" | Lost "a" = loses "one/some" nuance |
| "这个 **the** MCP server" | "这个 MCP server" | Lost "the" = loses "this specific" nuance |
| "the Agent 的状态" | "Agent 的状态" | Same — specific reference lost |

Even worse, `LEADERS` regex `^(?:i'll|i will|...)\s+` can accidentally delete English verb phrases that are embedded in Chinese sentences:

| Original | After compress |
|----------|---------------|
| "i will 检查这个 bug" | "检查这个 bug" — lost intent statement |

## Root Cause

1. `ARTICLES` regex uses `\b` word boundary which matches at CJK↔English transitions, so "a LLM" gets caught
2. In pure English, removing "a/an/the" is fine — they're filler. In **CJK-mixed text**, English articles are **not filler** — they're semantic modifiers for technical terms embedded in Chinese
3. Chinese has no article concept (中文没有冠词), so the English article is the **only** way to express specificity for the embedded English term
4. `compressProse()` runs all regex regardless of whether the text contains CJK characters

## Suggested Fix

In `compress.js`, add CJK-awareness to `compressProse()`:

```js
// Detect CJK-mixed text — Chinese, Japanese, Korean characters
const CJK_RE = /[一-鿿　-〿ぁ-ヶ가-힣]/;

function compressProse(text) {
  const hasCJK = CJK_RE.test(text);

  let s = text;
  s = s.replace(LEADERS, );
  s = s.replace(PLEASANTRIES, );
  s = s.replace(HEDGES, );
  s = s.replace(FILLERS, );

  // In CJK-mixed text, English articles are semantic — skip removal
  if (!hasCJK) {
    s = s.replace(ARTICLES, );
  }

  // ... rest unchanged
}
```

This is a **minimal, safe change** — it only skips ARTICLES removal when CJK characters are detected. All other compression (fillers, pleasantries, hedges) still applies in CJK-mixed text, which is correct — those are genuinely filler in any language context.

## Related

- #8 (Multilingual support) — too broad, already closed
- #34 (Classical Chinese mode) — different scope (wenyan is a deliberate style, not a bug)

## Environment

- caveman v18e45320 (latest)
- Claude Code, Chinese + English mixed conversations
- macOS, Node.js v22

lg320531124:
## Problem

In Chinese/Japanese/Korean mixed-language text, the `ARTICLES` regex `/\b(?:a|an|the)\s+(?=[a-z])/gi` incorrectly strips English articles that carry semantic weight:

| Original | After compress | Problem |
|----------|---------------|---------|
| "这是一个 **a** LLM 模型" | "这是一个 LLM 模型" | Lost "a" = loses "one/some" nuance |
| "这个 **the** MCP server" | "这个 MCP server" | Lost "the" = loses "this specific" nuance |

Chinese has no article concept — the English article is the **only** way to express specificity for embedded technical terms.

## Fix

Add CJK detection before `ARTICLES` removal in `compressProse()`. When CJK characters are present in the text, skip the `ARTICLES` regex entirely. All other compression (fillers, pleasantries, hedges, leaders) still applies — those are genuinely filler in any language context.

Also replaces NUL-byte sentinels (`\x00`) with Unicode PUA (`U+E000`) in `withProtectedSegments()`. The original code used NUL bytes as placeholder sentinels, which causes Git to misidentify `compress.js` as binary, making the diff invisible on GitHub. No functional change.

## Diff (GitHub shows binary due to NUL bytes in base)

```diff
+// CJK detection — Chinese, Japanese, Korean Unicode ranges.
+// In CJK-mixed text, English articles ("a/an/the") carry semantic weight
+// as specificity/quantity markers for embedded technical terms (e.g. "a LLM 模型",
+// "the MCP server"). Chinese has no article concept, so the English article is
+// the only way to express specificity. Skip ARTICLES removal when CJK is present.
+const CJK_RE = /[一-鿿　-〿ぁ-ヶ가-힣]/;
+
 const ARTICLES = /\b(?:a|an|the)\s+(?=[a-z])/gi;

+// Sentinel for protected segments: uses Unicode Private Use Area character
+// instead of NUL byte, which causes Git to misidentify the file as binary.
+const SENTINEL = "";

 function withProtectedSegments(text, transform) {
   for (const re of PROTECTED_PATTERNS) {
     working = working.replace(re, m => {
-      return `\0${i}\0`;
+      return `${SENTINEL}${i}${SENTINEL}`;
     });
   }
+  const sentinelRe = new RegExp(`${SENTINEL}(\\d+)${SENTINEL}`, "g");
   let out = transform(working);
-  out = out.replace(/\0(\d+)\0/g, (_, i) => segments[+i]);
+  out = out.replace(sentinelRe, (_, i) => segments[+i]);
 }

 function compressProse(text) {
+  const hasCJK = CJK_RE.test(text);
   let s = text;
   s = s.replace(LEADERS, "");
   s = s.replace(PLEASANTRIES, "");
   s = s.replace(HEDGES, "");
   s = s.replace(FILLERS, "");
-  s = s.replace(ARTICLES, "");
+  if (!hasCJK) {
+    s = s.replace(ARTICLES, "");
+  }
```

## Testing

```js
// Pure English — ARTICLES still removed (unchanged behavior)
compressProse("I will fix the bug") → "Fix bug"

// CJK-mixed — ARTICLES preserved, other compression still applies
compressProse("这是一个 a LLM 模型，I think") → "这是一个 a LLM 模型"
```

Fixes #575


lg320531124:
Gentle ping — open ~3 days. This fixes ARTICLES removal mangling CJK-mixed text (a correctness bug, not style). Small, focused diff. No CI gating. Ready for review when you have a moment.

lg320531124:
Rebased onto latest `main` to resolve conflicts with upstream changes.

- Squashed the two original commits (NUL→PUA sentinel + CJK-skip) into a single commit, since the binary-conflict resolution applied both changes together.
- Sentinel is now `U+E000` (PUA) — zero NUL bytes remain in `compress.js`.
- Preserved upstreams iterative `MAX_RESTORE_PASSES` restore loop (#444) on top of the CJK-skip change.
- `node --check` passes; functional test confirms CJK-mixed text keeps articles (`"a LLM 模型 the bug"` unchanged) while pure English drops them.

Mergeable state: CLEAN.

JuliusBrussee:
Reviewed. Mechanism confirmed: `ARTICLES` is `/\b(?:a|an|the)\s+(?=[a-z])/gi`, and the `i` flag makes the lookahead case-insensitive too, so `a LLM 模型` → `LLM 模型`. Your report is accurate.

Not merged this run because it's a policy call rather than a defect with one right answer: whether an English article inside Chinese text is filler or a semantic modifier is a judgment about compression aggressiveness, and this run only ships reproducible bug fixes. Flagged for the maintainer in #1012.

The code change itself is sound and minimal. The one question I'd raise for that decision: gating on *any* CJK codepoint disables ARTICLES for a mostly-English document containing a single Chinese word. That errs in the safe direction, but it's worth being deliberate about.

---

JuliusBrussee:
**Can't review this as sent** — GitHub renders the diff as `Binary files a/src/mcp-servers/caveman-shrink/compress.js and b/... differ`, so there's nothing to read.

Not your fault: `compress.js` on `main` already contains six literal NUL bytes (the `\x00${i}\x00` protect/restore sentinels around line 75), and git classifies any file containing NUL as binary. Anything you change in that file will show up this way.

Also see #575 — the underlying question (do English articles carry semantic weight in CJK-mixed prose?) is a design call and needs the maintainer's decision before an implementation can be judged right or wrong. Flagged in #1055.

The related `\b`-boundary bug in the same regexes — filler words being eaten out of hyphenated compounds — is fixed in #1055, so expect a conflict in this region.

---

JuliusBrussee:
**Correction to my comment above — `main` already carries a fix for #575.**

My earlier reply was written against a stale checkout. `5506ba8` ("fix(shrink): leave mixed CJK strings byte-identical") landed since and takes a broader approach than this PR: rather than skipping only `ARTICLES`, it short-circuits the whole compressor when CJK is detected, so mixed text is returned byte-identical. It also covers Hiragana, Katakana, Hangul and Bopomofo, not Han alone.

So this PR is superseded rather than unreviewable — thanks for filing both it and the analysis in #575, which is what the landed fix is built on.

The binary-diff note still holds for anything you send against `compress.js` in future: the file carries literal NUL sentinel bytes, so git classifies it as binary and GitHub will not render the diff.

---

AmirF194 (commented):
GitHub shows this file's diff as binary (no patch, the base file's NUL-byte sentinels this PR is also removing), so I pulled base and head content directly at `fa4815625eba0dbfc0e4cd22ef1f59b3251aecb5` and ran the real `compress()` export in a clean container.

`CJK_RE` (`src/mcp-servers/caveman-shrink/compress.js:49`) replaces the base file's `\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}\p{Script=Hangul}\p{Script=Bopomofo}` detection with an explicit character-range class, and it is narrower: it drops halfwidth katakana (U+FF61-FF9F) and CJK Unified Ideographs Extension-A (U+3400-4DBF) and beyond. Base used to skip compression for the whole string whenever any of those scripts appeared anywhere in it, so this narrowing reopens #575's exact bug for text in the scripts it no longer detects, since `compress()` now always runs `compressProse` and only gates `ARTICLES` on `hasCJK`.

```
$ node -e '
const { compress } = require("./compress.js");
const s = "using a LLM ｺﾝﾋﾟｭｰﾀｰ tool";
console.log(JSON.stringify(compress(s).compressed));
'
"Using LLM ｺﾝﾋﾟｭｰﾀｰ tool"
```

"a " is gone even though the sentence is CJK-mixed, same failure shape as the "a LLM 模型" example in the PR description, just with the CJK term expressed in halfwidth katakana instead of full-width Han/Kana.

Widening the character class back out to cover halfwidth katakana and the CJK extension blocks (or reusing the base file's `\p{Script=...}` form, which already covered both) would close this without giving up the narrower per-clause ARTICLES gate.

JuliusBrussee:
**Superseded by `5506ba8` on `main` — no action needed from you.**

That commit settled the design question your PR raised, and the other way from what this PR proposes: mixed-CJK text now passes through **byte-identically** instead of being selectively compressed. I re-verified all four rows from #575 against current `main` today — every one is identical in and out (`before === after`), including the `LEADERS` case.

So the outcome you wanted is in, by a broader route than skipping `ARTICLES` alone. Your analysis of the `\b` boundary is what pointed at the fix, and the same sloppiness turned out to be eating hyphen-joined compounds in ordinary English too.

Flagged #575 and this PR as closeable in #1070; the close is the maintainer's call, not mine. Thanks for the write-up and for the patience on the round trip.

---

JuliusBrussee:
**Already fixed on `main`** — `src/mcp-servers/caveman-shrink/compress.js` carries a `CJK` short-circuit that cites #575 by number and returns mixed-script text byte-for-byte unchanged, spacing included, before any English deletion rule runs.

Also checked for a sibling gap and found none: the Python compress tool in `skills/caveman-compress/scripts/` does no mechanical article stripping, so there was nothing to mirror the guard into.

Nothing left to change. Flagged in #1082 for the maintainer to close both this and #575; I don't close anything.

---

JuliusBrussee:
Superseded — the CJK guard is already on main (see my note on #575). `compress()` now skips the entire prose pass on CJK-mixed text, which is stricter than skipping `ARTICLES` alone and also covers the `"i will 检查这个 bug"` case from your issue.

For the record I did try to adopt this branch rather than close it out: it shares no merge base with current `main`, so it can't be cherry-picked.

Thanks for both the report and the patch — the diagnosis in #575 is what the shipped fix is built on.

---

JuliusBrussee:
**Superseded — the fix is already on `main`.** `5506ba8` ("fix(shrink): leave mixed CJK strings byte-identical (#575)") landed the same guard: `compress()` short-circuits on `CJK.test(text)` and returns the input unchanged, so articles, intent phrases, case and whitespace all survive in mixed text.

Your analysis in #575 is what that fix was built from — the root cause you identified (word boundaries matching at CJK↔English transitions, and English articles carrying specificity for embedded technical terms) is restated in the code comment. Thank you for it.

Nothing to do on your side. Whether to close this PR is the maintainer's call; I don't close anything.

---

PR #578, opened 2026-06-27
fix(hooks): prevent safeWriteFlag temp-file leak on Windows rename co…

edubraqd:
…ntention

safeWriteFlag wrote a temp file then renameSync'd it onto the flag path to publish atomically. On Windows, renameSync onto an existing target throws EPERM/EBUSY/EEXIST/EACCES when a concurrent session or the statusline holds the flag open without FILE_SHARE_DELETE. The catch swallowed the error and left the temp file orphaned, so .caveman-active.<pid>.<ts> files accumulated in the config dir (64 found under the multi-session Claude desktop app).

Hoist tempPath to an outer scope, retry the rename a few times to ride out brief lock contention, and always unlink the temp file in a finally so no litter is left regardless of outcome. Concurrent writers publish the same mode value, so a missed write is harmless. Update checksums.sha256 to match.


rajpratham1 (approved):
This is a focused bug fix with a clear purpose:

Fixes temporary file leaks on Windows when renameSync fails due to file locks.
Adds cleanup in finally, preventing orphaned temp files.
Retries transient rename failures (EPERM, EBUSY, EEXIST, EACCES).
Updates the checksum for the modified file.
Small, localized change with low regression risk.

JuliusBrussee:

PR #579, opened 2026-06-27
feat: input-side token-saving upgrades + eval quality gate

edubraqd:
## Why

In agentic Claude Code the weekly token limit is dominated by **input** — the whole context (system prompt, tool schemas, prior prose, every tool result) is re-sent every turn; caveman's output compression is the smallest slice. These changes target the input side and add a quality gate so compression never silently trades away correctness. A token saved on the re-sent prefix is re-billed every later turn (compounds); a token saved on output is billed once.

## What

**1. Agentic-loop output discipline + gated reinforcement + one-shot nudge** (`24c8e03`)
- `skills/caveman/SKILL.md`: new `## Agentic Loop` section — result-first, no recap of files just read, plan once then deltas, one intent clause per tool batch. Cross-references Auto-Clarity so security/irreversible/multi-step stay full prose.
- `caveman-mode-tracker.js`: per-turn reinforcement now fires every 3rd turn (+ on mode (re)activation), not every turn — the counter lives in a separate flag file so the injected string stays byte-stable and keeps hitting the prompt cache.
- `caveman-activate.js`: statusline-setup nudge is one-shot (a `.caveman-nudged` marker) instead of re-appending ~90 tokens to the cached SessionStart prefix every session.

**2. Eval quality gate — fidelity axis + two-gate accept rule** (`5681437`, `5c7e19a`)
- `measure.py` only counts tokens, which is gameable (a skill replying `k` scores -99% and "wins"). Adds `evals/judge.py` (LLM judge for rubric facts + deterministic verbatim check -> `fidelity.json`) and `evals/gate.py` (accept only if tokens hold AND fidelity holds; high-risk prompts like the git-rebase "rewrites history" warning allow zero fidelity loss).
- `evals/prompts/rubrics.json` for all 10 prompts; a committed `fidelity.json` baseline; `tests/test_eval_gate.py` (incl. the canonical "reply k -> REJECT"); `.github/workflows/eval-gate.yml`.

**3. Opt-in PostToolUse tool-result trim** (`de7da10`)
- `caveman-trim-tool-result.js`: when enabled, replaces oversized Read/Bash/Grep/Glob results (via the public `updatedToolOutput`) with a deterministic, lossless-first trimmed version before they enter context. OFF by default (`CAVEMAN_TRIM_TOOL_RESULTS=1`); wired via `--with-trim`, independent of the plugin (no double-fire).

**4. cavecrew as the default locate route** (`e2e7076`)
- Nudge the main thread to delegate locate-shaped work to `cavecrew-investigator` (its verbose Grep/Read stays out of main context); cap investigator/reviewer output at 25 rows.

**5. Structural ultra + honest input/output split in /caveman-stats** (`b01efd3`)
- `skills/caveman/SKILL.md`: the `ultra` level now collapses multi-sentence explanation into one causal chain and cuts transition sentences (not just word abbreviation), keeping the verbatim code-symbol/error-string exemption.
- `caveman-stats.js`: `/caveman-stats` reports the input-vs-output token split — output is the demo, input is where the weekly budget goes — and points at the input-side levers. Number formatting pinned to `en-US` so output is locale-deterministic.

## Testing
- `tests/test_hook_output.js` (8), `tests/test_trim_tool_result.js` (15), `tests/test_eval_gate.py` (7), `tests/test_caveman_stats.js` (31) all pass.
- Trim hook adversarially reviewed (determinism, data corruption, fail-open) — no issues.
- Gate verified end-to-end on the committed snapshot: caveman saves ~50% tokens at 100% fidelity.

## Notes
- Clean-room: hook mechanisms are from the public Claude Code hook docs.
- The Windows `safeWriteFlag` temp-leak fix is a separate PR (#578) — this branch deliberately leaves `caveman-config.js` untouched.
- `fidelity.json` was judged inline (an authenticated `claude -p` was unavailable in the author's environment); regenerate with `evals/judge.py` once CLI auth is set up.


edubraqd:
### Heads-up on item 3 (the opt-in PostToolUse trim hook)

Verified empirically on a live Claude Code session (logging wrapper on the hook): **the trim hook currently has ~no real-world effect**, for two reasons:

1. **Claude Code does not apply a string `updatedToolOutput` when the `tool_response` is a structured object** — and every built-in tool returns one: Bash `{stdout, stderr, interrupted, …}`, Read `{type, file}`, Grep `{mode, content, numLines, …}`, Glob likewise. The hook fires, extracts the text, trims it, and emits a valid `updatedToolOutput` string — but CC shows the original result unchanged. So it's a no-op for all four matched tools.
2. **Claude Code already persists oversized results natively** (~30KB → saved to disk + a ~2KB preview), more aggressively than this hook would.

The hook code is correct and unit-tested; the limitation is in the PostToolUse API surface, not the implementation. It ships **OFF by default**, so it harms nothing — but it won't save tokens on current Claude Code. A doc note was added (`src/hooks/README.md`). If a future CC honors `updatedToolOutput` for object results, the hook would need to return the object shape with a trimmed text field (not a string).

The other items in this PR (agentic-loop discipline, gated reinforcement, one-shot nudge, eval quality gate, cavecrew locate routing, structural ultra, /caveman-stats input/output split) are unaffected.

rajpratham1 (changes requested):
This is a very large PR (+2,196/-49 across 31 files). While the ideas are good, it's introducing multiple independent features in one PR:

Eval framework (judge.py, gate.py, rubrics, snapshots, CI)
New input-bounding hook
Installer changes
Mode changes (auto)
Statusline/session behavior
Documentation updates
Cavecrew behavior changes

These should be split into separate PRs for easier review and lower regression risk.

JuliusBrussee:
Thanks @edubraqd — input compression and quality gates now belong in Engine/proxy, not legacy Claude-only hooks. Closing this superseded architecture.

PR #590, opened 2026-07-01
fix(hooks): resolve SKILL.md path from src/hooks correctly

### LINKED ISSUE #589: SessionStart hook path off-by-one — SKILL.md not found, silently falls back to English-only ruleset
**What happen**
`src/hooks/caveman-activate.js` reads SKILL.md via `path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md')`. Hook file lives at `<plugin_root>/src/hooks/`, so this resolves to `<plugin_root>/src/skills/caveman/SKILL.md` — doesn't exist. Real file is `<plugin_root>/skills/caveman/SKILL.md`, one level further up. `readFileSync` fails, caught silently (empty `catch` at line 58), falls back to hardcoded ruleset baked into the hook — which is missing the "Preserve user's dominant language" rule that exists in the real SKILL.md. Result: every Claude Code plugin install replies in English caveman regardless of the user's language, even though SKILL.md says otherwise.

**Expected**
Hook finds real SKILL.md, so language-preservation (and any other rule added since the fallback was last hand-copied) actually applies.

**Before/after example**
```
Input: user writes prompt in Portuguese, caveman active
Got:   caveman reply in English
Want:  caveman reply in Portuguese (per SKILL.md "Preserve user's dominant language" rule)
```

**Platform**
- [x] Claude Code

**Version / install method**
Claude Code plugin (marketplace install). Confirmed on current `main` as of commit `1ce38a2295186e4fe5f5ed0c72853e0ee2de2c27`.

**Fix**
Line 56: `path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md')` → `path.join(__dirname, '..', '..', 'skills', 'caveman', 'SKILL.md')`.

marcoshmendes:
## Summary
- `caveman-activate.js` reads `SKILL.md` via `path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md')`. The hook lives at `<plugin_root>/src/hooks/`, so this resolved to `<plugin_root>/src/skills/...`, which doesn't exist — the real file is `<plugin_root>/skills/caveman/SKILL.md`.
- `readFileSync` failed, caught silently, and the hook fell back to the hardcoded ruleset — which is missing rules added to `SKILL.md` since (e.g. "Preserve user's dominant language"), so every plugin install replied in English regardless of the user's language.
- Fixed the path (added one more `..`) and corrected the stale comment above it.
- Added a regression test in `tests/test_hooks.py` asserting the activate hook's stdout contains "Preserve user's dominant language" — proves it's reading the real `SKILL.md`, not the fallback.

Fixes #589.

## Test plan
- [x] `npm test` — 73/73 pass
- [x] `python3 -m unittest tests.test_hooks` — 5/5 pass, including new regression test

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -48,12 +48,12 @@ if (INDEPENDENT_MODES.has(mode)) {
 const modeLabel = mode === 'wenyan' ? 'wenyan-full' : mode;
 
 // Read SKILL.md — the single source of truth for caveman behavior.
-// Plugin installs: __dirname = <plugin_root>/hooks/, SKILL.md at <plugin_root>/skills/caveman/SKILL.md
+// Plugin installs: __dirname = <plugin_root>/src/hooks/, SKILL.md at <plugin_root>/skills/caveman/SKILL.md
 // Standalone installs: __dirname = $CLAUDE_CONFIG_DIR/hooks/, SKILL.md won't exist — falls back to hardcoded rules.
 let skillContent = '';
 try {
   skillContent = fs.readFileSync(
-    path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
+    path.join(__dirname, '..', '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
   );
 } catch (e) { /* standalone install — will use fallback below */ }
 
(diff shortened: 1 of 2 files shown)


AmirF194:
Good diagnosis, and the off-by-one is exactly right for the `src/hooks/` layout.

Heads up that main already fixed this in #591, but with candidate-path resolution rather than a single hardcoded `../../`: it tries `$CLAUDE_PLUGIN_ROOT/skills`, then `../../skills`, then `../skills`. The reason it uses the list is the standalone install where the skill sits one level up, which a fixed `../../` would miss.

The regression test is a good idea, but there's an equivalent already on main (`test_activate_emits_skill_md_not_fallback_from_repo_layout`, asserting the intensity table shows up), so this looks superseded end to end. Probably cleanest to close, unless there's something in your test worth folding into the existing one.

JuliusBrussee:

PR #604, opened 2026-07-02
fix(install): treat spawn ENOENT as failure, not success

### LINKED ISSUE #592: install.js: spawn ENOENT treated as success — missing claude CLI reports "installed", skips standalone-hook fallback, nothing installed
**What happen**

On a machine where the `claude` CLI is not on PATH, `node bin/install.js --only claude --non-interactive` reports success and installs **nothing** — no plugin, no hooks, no settings.json. Worse, the standalone-hook fallback is skipped because the installer believes the plugin install succeeded.

Verified in a clean `node:20-bookworm` Docker container (no `claude` binary, `HOME=/root` empty):

```
→ Claude Code detected
  $ claude plugin marketplace add JuliusBrussee/caveman
  $ claude plugin install caveman@caveman
  hooks: plugin manifest handles SessionStart + UserPromptSubmit
🪨 done
  installed:
    • claude
```

Afterwards: `~/.claude/` does not exist at all.

**Root cause**

`spawnSync` on ENOENT returns `{ status: null, error: Error }`. `bin/install.js:445` coerces that to success:

```js
if ((r1.status || 0) === 0 && (r2.status || 0) === 0) {   // null || 0 === 0 → "success"
```

So `pluginInstallSucceeded = true`, and the hook-wiring decision matrix (issue #392 logic) takes the "plugin manifest handles hooks" branch — the exact fallback built for this situation never fires.

The `(r.status || 0) === 0` pattern appears at 8 sites: `bin/install.js:445` (claude), `:537` (gemini), `:558` (all npx-skills providers), `:946`, `:965`, `:977` (init), `:1081` (uninstall), `:1355` (skills-auto). Every one converts "binary not found / spawn failed" into success.

**Expected**

Spawn failure = install failure: report `claude` as failed, fall back to standalone hook wiring (which works offline from the local clone), exit non-zero.

**Fix**

```js
const spawnOk = (r) => !r.error && r.status === 0;
```
and use it at all 8 sites.

**Platform**: Linux (node:20-bookworm container), repro is OS-independent. main @ 25d22f8.

AmirF194:
## What happen

`spawnSync` reports a missing binary as `{ status: null, error }`. The installer's `(r.status || 0) === 0` checks coerce that `null` to success. Verified in a clean `node:20-bookworm` container: `node bin/install.js --only claude --non-interactive` on a box without the `claude` CLI printed `installed: • claude`, skipped the standalone-hook fallback with `hooks: plugin manifest handles SessionStart + UserPromptSubmit`, and left `~/.claude/` nonexistent — success reported, nothing installed.

Fixes #592.

## Fix

- New `spawnOk(r)` helper: `!!r && !r.error && r.status === 0`. Dry-run results (`{status: 0}`, no `error`) still pass.
- Used at all 8 coercion sites: claude plugin install, gemini extension install, npx-skills providers, mcp-shrink registration, `runInit` (local + curl-fallback paths), claude plugin uninstall, and the skills auto-detect fallback.
- `installClaude` now warns `claude CLI not found on PATH (or could not be spawned)` before recording the failure, so the summary isn't mysterious.

With this, the existing hook-wiring decision matrix does the right thing automatically: plugin install fails → `falling back to standalone wiring` → hooks copied from the local clone (works offline) and merged into settings.json.

## Tests

New e2e test in `tests/installer/e2e.freshinstall.test.mjs` runs the real installer with an **empty PATH** (guaranteed ENOENT even on machines that have `claude`), asserting: failure is reported, `claude` is not listed as installed, the standalone fallback triggers, hook files exist on disk, and `settings.json` has the SessionStart wiring. Passes everywhere since it doesn't depend on the CLI being present.

Full suites: 29/29 dry-run + unit, new test green, `node --check` clean.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -418,6 +418,15 @@ function captureSpawn(cmd, args) {
   catch (_) { return { status: 1, stdout: '', stderr: '' }; }
 }
 
+// spawnSync reports a missing binary as { status: null, error }, so the old
+// `(r.status || 0) === 0` checks read ENOENT as success — a machine without
+// the `claude` CLI got "installed: claude" with nothing installed and the
+// standalone-hook fallback skipped (issue #592). Every spawn result must pass
+// through here before being treated as "it worked".
+function spawnOk(r) {
+  return !!r && !r.error && r.status === 0;
+}
+
 function absoluteNodePath() {
   return process.execPath;
 }
@@ -442,10 +451,13 @@ async function installClaude(ctx) {
   } else {
     const r1 = runSpawn('claude', ['plugin', 'marketplace', 'add', REPO], null, opts.dryRun);
     const r2 = runSpawn('claude', ['plugin', 'install', 'caveman@caveman'], null, opts.dryRun);
-    if ((r1.status || 0) === 0 && (r2.status || 0) === 0) {
+    if (spawnOk(r1) && spawnOk(r2)) {
       results.installed.push('claude');
       pluginInstallSucceeded = true;
     } else {
+      if (r1.error || r2.error) {
+        warn('  claude CLI not found on PATH (or could not be spawned)');
+      }
       results.failed.push(['claude', 'claude plugin install failed']);
     }
   }
@@ -534,7 +546,7 @@ function installGemini(ctx) {
     }
   }
   const r = runSpawn('gemini', ['extensions', 'install', `https://github.com/${REPO}`], null, opts.dryRun);
-  if ((r.status || 0) === 0) results.installed.push('gemini');
+  if (spawnOk(r)) results.installed.push('gemini');
   else results.failed.push(['gemini', 'gemini extensions install failed']);
   process.stdout.write('\n');
 }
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This fixes a genuine bug where spawnSync() returning { status: null, error: ENOENT } was incorrectly treated as success because of (r.status || 0) === 0. Introducing a shared spawnOk() helper makes the success criteria explicit and consistent across the installer, and the regression test verifies the important fallback behavior when the claude CLI is unavailable.

rajpratham1 (approved):
This is a focused correctness fix that addresses a real bug:

Introduces a single spawnOk() helper to correctly distinguish successful execution from spawnSync failures (ENOENT results in status === null).
Replaces the unsafe (r.status || 0) === 0 pattern throughout the installer.
Adds an end-to-end regression test verifying that a missing claude binary is reported as a failure and that the standalone fallback path is executed.
Low risk: behavior only changes for failed process spawns; successful executions remain unchanged.

PR #605, opened 2026-07-02
fix(uninstall): match managed hook basenames, not substring

### LINKED ISSUE #593: uninstall deletes unrelated user hooks whose command merely contains the substring "caveman"
**What happen**

`node bin/install.js --uninstall` deletes **any** hook entry in `settings.json` whose command string contains the substring `caveman` anywhere — including hooks that have nothing to do with this project.

`bin/install.js:1052` calls `removeCavemanHooks(settings, 'caveman')`, and `bin/lib/settings.js:157-174` matches by `command.includes('caveman')`.

**Repro (confirmed)**

User-authored hook in settings.json:

```json
{ "type": "command", "command": "node /Users/me/Projects/caveman-notes/my-hook.js" }
```

Run `--uninstall` → "removed 1 caveman hook entry" — the user's unrelated PreToolUse hook is gone. Any path/repo/username containing "caveman" (including a clone of this very repo used in a hook) triggers deletion.

**Expected**

Uninstall removes only hooks this installer manages. The codebase already has the right pattern: `pruneOrphanedManagedHooks` (`bin/lib/settings.js:249-259`) deliberately matches exact basenames from `MANAGED_HOOK_BASENAMES` to avoid this exact problem.

**Fix**

Make `removeCavemanHooks` match on `MANAGED_HOOK_BASENAMES` (caveman-activate.js / caveman-mode-tracker.js / caveman-statusline.*) instead of a substring.

main @ 25d22f8.

AmirF194:
## What happen

`--uninstall` deleted **any** hook entry in settings.json whose command contained the substring `caveman` anywhere — `bin/install.js` called `removeCavemanHooks(settings, 'caveman')` and `bin/lib/settings.js` matched with `command.includes(marker)`. A user-authored hook like

```json
{ "type": "command", "command": "node /Users/me/Projects/caveman-notes/my-hook.js" }
```

was silently destroyed ("removed 1 caveman hook entry"). Any path, repo, or username containing the word — including a clone of this repo used in a hook — triggered deletion.

Fixes #593.

## Fix

`removeCavemanHooks` now tokenizes each command (quote-aware, same tokenizer the prune pass already used — hoisted to module scope and shared) and removes an entry only when a token's **exact basename** is one of `MANAGED_HOOK_BASENAMES`. This is the same design `pruneOrphanedManagedHooks` deliberately adopted to avoid this exact false-positive class; uninstall now follows it.

Details:
- `path.win32.basename` (splits on both `/` and `\`) so a settings.json written on Windows is handled correctly anywhere.
- `caveman-statusline.ps1` added to `MANAGED_HOOK_BASENAMES` — the installer wires it on Windows but it was missing from the set, so Windows statusline entries were invisible to both removal and the #471 orphan-prune self-heal.
- `mycaveman-activate.js` (superstring basename) and `caveman-notes/` paths are never matched.

## Tests

Unit tests updated to exercise real managed-script commands, plus two new regression tests: user hooks that merely mention "caveman" survive uninstall, and Windows-style `caveman-stats.js` wiring is removed. 31/31 unit + dry-run suites pass. Also verified end-to-end: synthetic settings.json with one managed hook + one `caveman-notes` user hook → `--uninstall` removes exactly the managed entry.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -1049,7 +1049,7 @@ function uninstall(ctx) {
   if (fs.existsSync(settingsPath)) {
     const settings = SETTINGS.readSettings(settingsPath);
     if (settings) {
-      const removed = SETTINGS.removeCavemanHooks(settings, 'caveman');
+      const removed = SETTINGS.removeCavemanHooks(settings);
       // Drop our statusline if it points at our script
       if (settings.statusLine) {
         const cmd = typeof settings.statusLine === 'string' ? settings.statusLine : (settings.statusLine.command || '');
--- a/bin/lib/settings.js
+++ b/bin/lib/settings.js
@@ -149,12 +149,52 @@ function addCommandHook(settings, event, opts) {
   return true;
 }
 
+// ── Managed hook scripts ──────────────────────────────────────────────────
+// The exact script basenames this installer wires into settings.json. Every
+// helper that decides "is this hook ours?" must match against these — never
+// against a bare "caveman" substring, which also matches user-authored hooks
+// that merely mention the word in a path (issue #593).
+const MANAGED_HOOK_BASENAMES = new Set([
+  'caveman-activate.js',
+  'caveman-mode-tracker.js',
+  'caveman-stats.js',
+  'caveman-statusline.sh',
+  'caveman-statusline.ps1',
+]);
+
+// Split a command into shell-ish tokens, honoring single/double quotes so a
+// path containing spaces survives intact. Good enough for hook commands we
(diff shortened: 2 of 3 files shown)


rajpratham1 (approved):
This improves uninstall safety by matching only the exact managed hook script basenames instead of any command containing "caveman". That prevents user-authored hooks (e.g. paths like caveman-notes/ or mycaveman-activate.js) from being removed accidentally. The shared helper also reduces duplicated parsing logic, and the new tests cover quoted paths, Windows paths, malformed hook structures, and the regression scenario.

PR #607, opened 2026-07-02
fix(settings): string-aware trailing-comma removal in JSONC reader

### LINKED ISSUE #595: settings.js: stripJsonComments trailing-comma sweep corrupts string values containing ,} or ,]
**What happen**

`stripJsonComments` in `bin/lib/settings.js` ends with a trailing-comma sweep that runs over the whole output **including string contents** (`bin/lib/settings.js:62`):

```js
out.replace(/,(\s*[}\]])/g, '$1')
```

The comment above claims this is safe after comment-stripping; it isn't — string bodies survive comment-stripping intact.

**Repro (confirmed)**

```jsonc
{ "cmd": "echo ,}", // comment
}
```

parses to `"cmd": "echo }"` — the comma inside the quoted string is silently deleted. Any JSONC settings whose string values contain `,}` or `,]` (shell brace expansion `file{,}`, inline JSON passed to a hook, jq/awk snippets) is corrupted on read — and then **persisted corrupted** by `writeSettings` on the next install/uninstall.

It only triggers for files that fail strict `JSON.parse` first, i.e. real JSONC files — exactly the files this module exists to protect.

**Fix**

Do the trailing-comma removal in the same character-scan that already tracks in-string state for comment stripping, instead of a global regex afterward.

main @ 25d22f8.

AmirF194:
## What happen

`stripJsonComments` ended with a global regex sweep `out.replace(/,(\s*[}\]])/g, '$1')` over the whole comment-stripped output — **including string contents**. A JSONC settings.json whose string values contain `,}` or `,]` (shell brace expansion `file{,}`, inline JSON in hook args, jq/awk snippets) was silently corrupted on read: `{"cmd": "echo ,}"}` parsed to `"cmd": "echo }"` — then persisted corrupted by `writeSettings` on the next install/uninstall. It only triggers for files that fail strict `JSON.parse` first, i.e. real JSONC files — exactly what this module exists to protect.

Fixes #595.

## Fix

Replace the regex with `stripTrailingCommas()` — a scan that tracks string state (same approach as the comment stripper above it, including backslash escapes) and drops a comma only when it's outside a string and the next non-whitespace char is `}` or `]`.

## Tests

Three new unit tests: `,}` / `,]` inside string values survive, real trailing commas after strings are still removed, and escaped quotes before `,}` inside strings are handled. 30/30 unit suite passes; full installer suites (77 tests) and `verify_repo.py` green locally **and in a clean node:20-bookworm container**.

### DIFF
--- a/bin/lib/settings.js
+++ b/bin/lib/settings.js
@@ -57,9 +57,38 @@ function stripJsonComments(src) {
     if (c === '/' && next === '*') { inBlock = true; i += 2; continue; }
     out += c; i++;
   }
-  // Trailing-comma sweep — only outside strings, but stripping happened above
-  // so a regex over the comment-free output is safe.
-  return out.replace(/,(\s*[}\]])/g, '$1');
+  return stripTrailingCommas(out);
+}
+
+// ── stripTrailingCommas ────────────────────────────────────────────────────
+// Remove `,` when the next non-whitespace char is `}` or `]` — but only
+// OUTSIDE strings. The old global regex ran over string contents too and
+// silently corrupted values like `"echo ,}"` → `"echo }"` (issue #595);
+// comment-stripping does not sanitize string bodies, so a string-aware scan
+// is required here as well.
+function stripTrailingCommas(src) {
+  let out = '';
+  let i = 0;
+  const n = src.length;
+  let inString = false;
+  let stringChar = '';
+  while (i < n) {
+    const c = src[i];
+    if (inString) {
+      out += c;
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This looks good. The new string-aware trailing-comma removal avoids modifying content inside string literals while still stripping genuine trailing commas from objects and arrays. The regression tests cover the reported issue along with escaped quotes and normal trailing-comma handling, making the change well validated.

PR #610, opened 2026-07-02
fix(opencode): --force migrates legacy AGENTS.md, never wipes it

### LINKED ISSUE #594: opencode: --force on legacy AGENTS.md replaces the whole file — user content destroyed, and the installer's own hint steers users into it
**What happen**

For opencode, when `~/.config/opencode/AGENTS.md` contains the legacy un-fenced sentinel, the installer prints:

> re-run with `--force` to replace it with a fenced block

But `--force` executes a **full-file replace** (`bin/install.js:707-716`: `fs.writeFileSync(agentsMd, fencedBlock)`). Everything the user wrote in AGENTS.md above/below the legacy block is destroyed. No backup is taken.

**Repro (confirmed)**

```
# My precious user rules
Always use tabs.

Respond terse like smart caveman. ...   <- legacy sentinel
```

→ `--force` → file now contains **only** the fenced caveman block. User rules gone.

Note `--help` describes `--force` as merely "Re-run even if a target reports already installed" — nothing warns it rewrites user files. The installer actively steers users with mixed legacy files into data loss.

**Expected**

Replace only the legacy caveman block in place (or append fenced block + strip just the sentinel paragraph), and/or write a `.bak` before any destructive rewrite.

main @ 25d22f8.

AmirF194:
## What happen

When `~/.config/opencode/AGENTS.md` contains the legacy un-fenced sentinel, the installer prints *"re-run with `--force` to replace it with a fenced block"* — and `--force` then executed a **full-file replace** (`fs.writeFileSync(agentsMd, fencedBlock)`). All user-authored content around the legacy block was destroyed, no backup taken. The installer's own hint steered users with mixed files into data loss, and `--help` describes `--force` as merely "Re-run even if a target reports already installed".

Fixes #594.

## Fix

Migrate instead of wipe:

1. One-time backup to `AGENTS.md.bak` before any rewrite (same once-only pattern as `opencode.json.bak`).
2. Remove **only the legacy block**: exact match of the current rule body when possible; otherwise cut from the sentinel's paragraph start to EOF — the legacy path *appended* the block, so user content precedes it, and the heuristic path says so explicitly and points at the backup.
3. Write the fenced block after the preserved user content.

## Tests

New e2e test drives the real installer against a mixed legacy AGENTS.md (user rules + legacy body): user content survives, the fence is present, the legacy un-fenced copy is gone, and the backup contains the original. The older-revision heuristic path verified manually (user notes above sentinel survive, `.bak` written). Full suites green locally **and in a clean node:20-bookworm container**, including a targeted container repro of the exact reported scenario.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -705,14 +705,36 @@ function installOpencode(ctx) {
       if (alreadyFenced) {
         note(`  ${agentsMd} already contains caveman ruleset`);
       } else if (alreadyByLegacySentinel) {
-        note(`  ${agentsMd} contains a legacy (un-fenced) caveman block — leaving as-is`);
-        note('  re-run with --force to replace it with a fenced block');
+        if (!opts.force) {
+          note(`  ${agentsMd} contains a legacy (un-fenced) caveman block — leaving as-is`);
+          note('  re-run with --force to migrate it to a fenced block');
+        }
         if (opts.force) {
-          // Replace the entire file with a clean fenced version. The legacy
-          // path didn't fence, so we can't isolate the block — full rewrite is
-          // the only safe option under --force.
-          fs.writeFileSync(agentsMd, fencedBlock, { mode: 0o644 });
-          process.stdout.write(`  rewrote ${agentsMd} with fenced caveman block\n`);
+          // Migrate, don't wipe (issue #594): the old code replaced the whole
+          // file, destroying any user-authored content around the legacy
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
The change fixes a real data-loss issue by migrating legacy AGENTS.md content instead of overwriting the entire file when --force is used. Creating a backup and preserving user-authored content is the safer behavior, and the added end-to-end test covers the regression well.

PR #524, opened 2026-06-14
feat(install): add Hermes Agent support

Vagesh21:
Adds Hermes Agent as a supported provider with native skill installation.

## Changes
- Added 

╭─ Hermes Agent v0.16.0 (2026.6.5) · upstream 6b76284c · local 5a819d39 (+1 ca─╮
│                                       Available Tools                        │
│    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣀⡀⠀⣀⣀⠀⢀⣀⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀     browser: browser_back, browser_click,  │
│    ⠀⠀⠀⠀⠀⠀⢀⣠⣴⣾⣿⣿⣇⠸⣿⣿⠇⣸⣿⣿⣷⣦⣄⡀⠀⠀⠀⠀⠀⠀     ...                                    │
│    ⠀⢀⣠⣴⣶⠿⠋⣩⡿⣿⡿⠻⣿⡇⢠⡄⢸⣿⠟⢿⣿⢿⣍⠙⠿⣶⣦⣄⡀⠀     browser-cdp: browser_cdp,              │
│    ⠀⠀⠉⠉⠁⠶⠟⠋⠀⠉⠀⢀⣈⣁⡈⢁⣈⣁⡀⠀⠉⠀⠙⠻⠶⠈⠉⠉⠀⠀     browser_dialog                         │
│    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣴⣿⡿⠛⢁⡈⠛⢿⣿⣦⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀     clarify: clarify                       │
│    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠿⣿⣦⣤⣈⠁⢠⣴⣿⠿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀     code_execution: execute_code           │
│    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠉⠻⢿⣿⣦⡉⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀     computer_use: computer_use             │
│    ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⢷⣦⣈⠛⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀     cronjob: cronjob                       │
[...]

PR #611, opened 2026-07-02
fix(install): make Hermes skill install safe

freQuensy23-coder:
## Summary
- Adds Hermes Agent as a native caveman skill target.
- Installs the seven caveman skills into `$HERMES_HOME/skills/productivity` or `~/.hermes/skills/productivity`.
- Makes Hermes reinstall safe: existing local/self-edited skills are preserved unless `--force` is passed.
- Adds Hermes cleanup to `--uninstall`.
- Adds Hermes-specific installer regression tests.

## Fixes from review
- Avoids deleting Hermes skill directories on normal reinstall.
- Removes Hermes-installed skills during uninstall.
- Labels Hermes as a native copy provider in `--list`.
- Fails the Hermes install path if shipped source skill directories are missing.

## Test plan
- [x] `node --test tests/installer/hermes.test.mjs`
- [x] `npm test`
- [x] `node -c bin/install.js`
- [x] `git diff --check origin/main...HEAD`
- [x] `python3 -m pytest tests -q`
- [x] Manual isolated install with `HERMES_HOME=$(mktemp -d)/hermes-home node bin/install.js --only hermes` and `hermes skills list`

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -228,8 +228,8 @@ const PROVIDERS = [
   // CLI agents — require the binary. The `||dir:~/.foo` fallbacks were the
   // main source of false positives (warp, kiro, junie etc. leave config dirs
   // behind on uninstall).
-    { id: 'hermes',     label: 'Hermes Agent',        mech: 'npx skills add (hermes)',       detect: 'command:hermes',        profile: 'hermes' },
-    { id: 'aider-desk', label: 'Aider Desk',          mech: 'npx skills add (aider-desk)',   detect: 'command:aider', profile: 'aider-desk' },
+  { id: 'hermes',     label: 'Hermes Agent',        mech: 'native Hermes skills copy',   detect: 'command:hermes',        profile: 'hermes' },
+  { id: 'aider-desk', label: 'Aider Desk',          mech: 'npx skills add (aider-desk)',   detect: 'command:aider', profile: 'aider-desk' },
   { id: 'amp',        label: 'Sourcegraph Amp',     mech: 'npx skills add (amp)',          detect: 'command:amp',             profile: 'amp' },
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
Looks good. The Hermes installer now copies only the expected skill directories, preserves existing user modifications unless --force is specified, supports clean uninstall, and includes thorough end-to-end test coverage for install, reinstall, uninstall, and provider detection. The changes are well-contained and improve installer safety.

AmirF194:
The fail-fast guard here is a genuine improvement worth keeping. Main's `installHermes` warns "skill dir not found" per directory but still does `results.installed.push('hermes')`, so a missing source tree currently reports a successful install having copied nothing (same false-success class as the spawn-ENOENT fix). Your pre-check that returns failed instead is the right behavior.

The catch is that main already merged full Hermes support in #524: `installHermes`, `hermesConfigDir` (with `HERMES_HOME`), the PROVIDERS row, uninstall pruning, the dispatch, and `tests/installer/hermes.test.mjs` are all there now. That's why this shows as conflicting; it re-adds all of it from scratch. The high-value move is probably to rebase onto current main and reduce this to just the delta over #524, which is really the missing-source fail-fast plus any test for it that main doesn't already have.

Two smaller notes. The PROVIDERS row here adds `profile: 'hermes'`, which main omits; since Hermes is native-dispatched the profile never fires, and a slug that isn't in vercel-labs/skills can bite if anything ever falls through to the skills path, so I'd drop it. And the skip-existing-unless-force behavior is a real change from main's overwrite: it preserves user edits, but it also means a caveman upgrade won't refresh the Hermes skills on a normal reinstall. Worth calling out so the maintainer can pick the tradeoff deliberately.

JuliusBrussee:
Thanks @freQuensy23-coder — Hermes install/uninstall ownership is now symmetric and verified. Closing as implemented.

PR #615, opened 2026-07-02
fix: include spawn-options.js in caveman-shrink npm package

### LINKED ISSUE #597: caveman-shrink: package.json files omits spawn-options.js — next npm publish ships a package that crashes on launch
**What happen**

The next `npm publish` of `caveman-shrink` will ship a package that crashes on every launch.

`src/mcp-servers/caveman-shrink/package.json:26-30`:

```json
"files": ["index.js", "compress.js", "README.md"]
```

but `index.js:43` does `require('./spawn-options')` — and `spawn-options.js` is not in `files`.

Verified with `npm pack --dry-run`: tarball contains only 4 files, no `spawn-options.js` → `MODULE_NOT_FOUND` at startup.

**Compounding**

Registry `caveman-shrink@0.1.0` is stale — its published `index.js` predates the Windows `getSpawnOptions` fix (verified by diffing the downloaded tarball against repo source) — yet the repo `version` is still `0.1.0`. So `npx -y caveman-shrink` users (the exact wiring `bin/install.js --with-mcp-shrink` writes) still get the old Windows-broken spawn, and the fix cannot be published without a version bump anyway.

**Fix**

Add `spawn-options.js` to `files`, bump to `0.1.1`, republish.

main @ 25d22f8.

shubthenub:
Closes #597

### DIFF
--- a/src/mcp-servers/caveman-shrink/package.json
+++ b/src/mcp-servers/caveman-shrink/package.json
@@ -1,6 +1,6 @@
 {
   "name": "caveman-shrink",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "MCP proxy that compresses prose fields (tool descriptions, etc.) using caveman rules. Same accuracy, fewer context tokens.",
   "license": "MIT",
   "homepage": "https://github.com/JuliusBrussee/caveman",
@@ -25,6 +25,7 @@
   "files": [
     "index.js",
     "compress.js",
-    "README.md"
+    "README.md",
+    "spawn-options.js"
   ]
 }


rajpratham1 (approved):
This fixes a packaging issue by including spawn-options.js in the published npm package. Without it, consumers installing from npm could encounter runtime module resolution errors. The accompanying patch version bump is appropriate for this packaging fix.

AmirF194:
Confirmed the underlying gap was real: `spawn-options.js` is required at runtime but was missing from the published `files[]`. The thing is it already landed on main in a884696, where the version is 0.1.1 and `spawn-options.js` is in the `files` array, so this diff is a no-op against current main and that's why it's showing as conflicting. Nothing wrong with the change itself, it just looks already done.

If you want a follow-up that still adds value, a packaging guard test that asserts every path in `files[]` exists on disk before publish would catch the next drift of this kind. #625 already takes that angle if it helps to coordinate. Thanks for chasing it.

JuliusBrussee:
Thanks for catching the missing package file. spawn-options.js is now included in the published file list through [a8846964](https://github.com/JuliusBrussee/caveman/commit/a8846964), shipped in [v2.1.0](https://github.com/JuliusBrussee/caveman/tree/v2.1.0), with a packaging regression test. I’m closing this PR as superseded, not rejected.

Issue #603, opened 2026-07-02
commands/caveman-init.toml runs repo-relative src/tools path — fails for every installed Codex/Gemini user


AmirF194:
**What happen**

`commands/caveman-init.toml:2` tells the agent to run

```
node src/tools/caveman-init.js {{args}}
```

"in the current repo". For every installed Codex / Gemini-extension user, cwd is **their** project — `src/tools/caveman-init.js` only exists inside the caveman clone / extension directory. The command fails with `Cannot find module` for everyone except caveman developers.

**Fix**

Use the runner that works from anywhere:

```
npx -y github:JuliusBrussee/caveman -- --with-init {{args}}
```

or resolve the extension's absolute install path.

main @ 25d22f8.

AmirF194:
Fixed on main, caveman-init.toml now runs the standalone script via curl|node for installed users (and the local path only inside a caveman checkout). This issue can be closed.

JuliusBrussee:
Thanks for reporting this. This was addressed on main in [`5f079ab7`](https://github.com/JuliusBrussee/caveman/commit/5f079ab7) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this as implemented. If the current release still reproduces it, reply with your Caveman version, platform, command, and a minimal reproduction and I’ll reopen it.

PR #616, opened 2026-07-02
Fix caveman-init command runner

Shevanio:
# Context

The `/caveman-init` command currently depends on the current working directory containing `src/tools/caveman-init.js`, limiting its portability and usability across different user projects.

# Solution

This pull request updates the `commands/caveman-init.toml` file to replace the repository-relative Node invocation with a portable `npx` command, preserving existing metadata and argument placeholders. The prompt now instructs users to run `npx -y github:JuliusBrussee/caveman -- --with-init {{args}}`, ensuring the command is user-project independent.

# Scope

- Update `commands/caveman-init.toml` to use the portable runner.
- Add tests to verify the updated command does not reference `src/tools/caveman-init.js`.

# Tests Run

Focused tests for the caveman-init command were added, which check the command's structure against the new requirements.

# Results

Checks passed, but no tests ran and no static analysis validated the change — review carefully. 

# Risks

Minimal risk due to the focused nature of the updates. However, there could be edge cases not considered in user environments.

# Related Issue

None

### DIFF
--- a/commands/caveman-init.toml
+++ b/commands/caveman-init.toml
@@ -1,2 +1,2 @@
 description = "Drop the always-on caveman activation rule into the current repo for every IDE agent"
-prompt = "Run `node src/tools/caveman-init.js {{args}}` in the current repo and report the result. Use --dry-run first if the user did not pass --force, so we never silently overwrite an existing rule file."
+prompt = "Run `npx -y github:JuliusBrussee/caveman -- --with-init {{args}}` in the current repo and report the result. Use --dry-run first if the user did not pass --force, so we never silently overwrite an existing rule file."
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This change makes the caveman-init command portable by invoking the published GitHub package instead of referencing a repository-local script path. The added test also guards against regressions by verifying the command template. The implementation is small, focused, and aligns with the command's intended usage.

Shevanio:
Closing this — the same problem was already fixed upstream in 51e1990 (fix for #603), which makes `caveman-init` work outside a caveman checkout by gating the repo-relative path on the file existing and falling back to the standalone `curl … | node -` script.

This PR proposed a different approach (`npx -y github:JuliusBrussee/caveman -- --with-init`), but since the underlying issue is already resolved on `main`, there's no need for it. Thanks!

PR #309, opened 2026-04-28
fix(validate): add inline backtick span validation

hireblackout:
## Summary

- Add `extract_inline_codes()` function that strips fenced code blocks before extracting inline backtick spans
- Add `validate_inline_codes()` validator that errors when inline codes are lost and warns when added
- Wire validation into `validate()` function alongside existing checkers
- Add 11 unit tests covering extract, validate, and integration scenarios

## Problem

The `validate()` function performs structural integrity checks on compressed markdown by validating:
- Fenced code blocks
- URLs
- File paths
- Headings
- Bullet counts

However, inline backtick spans (`code`) are not validated. A compressor that drops or modifies inline references such as:
- `rm -rf /`
- `~/.claude/settings.json`
- `ANTHROPIC_API_KEY`

...passes validation undetected, causing silent data loss.

## Solution

Implement two new functions:

### extract_inline_codes(text)
```python
def extract_inline_codes(text):
    text_without_fences = re.sub(r"^```[\s\S]*?^```", "", text, flags=re.MULTILINE)
    text_without_fences = re.sub(r"^~~~[\s\S]*?^~~~", "", text_without_fences, flags=re.MULTILINE)
    return re.findall(r"\`([^\`]+)\`", text_without_fences)
```
- Strips fenced blocks first to avoid double-counting
- Extracts all inline backtick spans using regex

### validate_inline_codes(orig, comp, result)
- Compares sets of inline codes between original and compressed
- Errors if any code present in original is absent in compressed (data loss)
- Warns if any code in compressed is absent in original (unintended addition)

## Files Changed

- `caveman-compress/scripts/validate.py` — Add extraction + validation functions + wire into validate()
- `tests/test_validate_inline.py` — Add 11 test cases

## Testing

```bash
$ python3 -m unittest tests.test_validate_inline -v
test_fenced_blocks_excluded ... ok
test_inline_only ... ok
test_mixed_content ... ok
test_empty ... ok
test_match ... ok
test_lost ... ok
test_added ... ok
test_empty_orig ... ok
test_both_empty ... ok
test_validate_inline_codes_wired ... ok
```

PR #619, opened 2026-07-02
fix(caveman-compress): protect code by construction (mask-and-splice), force UTF-8 file I/O + integrity gate, fix inline-code extractor fence bug

FatalMerlin:
**Title:** fix(caveman-compress): protect code by construction (mask-and-splice), force UTF-8 file I/O + integrity gate, fix inline-code extractor fence bug

**Disclosure:** this PR (diagnosis, patches, and tests) was drafted by Claude (Anthropic's Claude Code). I (@FatalMerlin) hit each of these failures directly on real files, worked through the fixes with Claude, reviewed the diagnosis and the diffs myself, and I'm signing off on them as accurate and correct before submitting — this is a reviewed and approved contribution, not raw unreviewed model output. (The third commit, the UTF-8 fix, was surfaced the same way: a real batch-compression run on a set of memory files.)

## Summary

- `extract_inline_codes()` used a fragile, column-0-anchored regex to strip fenced blocks before scanning for inline code. Any fence not flush-left (e.g. indented under a list item) was invisible to it, and the fallback then paired backticks *across* the fence, producing garbage tokens instead of the real inline code (repro + regression test in the diff).
- More fundamentally: code preservation during compression was only ever a **prompt instruction** to the compression model (`build_compress_prompt`'s "do NOT modify code blocks/backticks" rules), then checked byte-exact after the fact by `validate.py`. That's asking a non-deterministic model to behave, not guaranteeing it — on my install this currently means a hard fail + discard after 2 retries on real code-dense files, rather than silent corruption, but the underlying gap is the same one #112 reported.
- **Sibling gap in the file layer:** the same "protect it, but only softly" problem existed for **non-ASCII prose**, one level down from the model. The skill's file I/O read/wrote with no explicit `encoding=`, so on Windows it fell back to the locale codepage (cp1252). Prose characters outside cp1252 (`→`, `≠`) raised `UnicodeEncodeError` mid-write; characters *inside* cp1252 but not ASCII (em-dash `—`, curly quotes, bullet, ellipsis) were silently mis-encoded and the output became invalid UTF-8 while the run still exited 0 with "Validation passed" — the mask/validate layer only guards code and frontmatter, not general text integrity.

**Prior art / why this isn't a resubmit:** #309, #307, and #293 each tried fixing this bug class before (adding an inline-code check with its own from-scratch regex) and were closed unmerged; #481 fixed `extract_code_blocks` itself but didn't touch `extract_inline_codes`, so the bug above is still live on `main`. All of those were *detect-after-the-fact* approaches. This PR's masking commit is a different strategy — *prevent by construction* — so it's meant to complement/supersede that line of attempts, not repeat it.

## What this PR does

1. **Fix `extract_inline_codes`** (`validate.py`) to reuse the existing line-based `extract_code_blocks` fence extractor instead of its own regex, so both checks agree on what counts as "inside a fence."
2. **Add mechanical code protection** (`compress.py`): before the compression prompt is built, `mask_code()` replaces every fenced block and inline `` `span` `` with an opaque placeholder token (`⟦CVMN-CODE-n⟧` — non-markdown-active Unicode brackets, chosen so a whitespace-normalizing compressor can't fracture them). The model compresses prose + placeholders only, never real code. `unmask_code()` splices the originals back verbatim by index afterward.
3. **Fail closed on code**: if a placeholder is dropped, duplicated, or otherwise altered by the model, `unmask_code` raises `PlaceholderIntegrityError` and `compress_file` aborts before any backup/write — original file untouched, no partial or best-guess splice.
4. **Force UTF-8 file I/O + fail closed on text** (`compress.py`, `validate.py`, `detect.py`): pin `encoding="utf-8"` on every read/write in the skill so file I/O no longer depends on the host locale, and add a `UnicodeIntegrityError` gate before overwrite — the compressed output must round-trip encode/decode as UTF-8 and introduce no new `U+FFFD` versus the source, else abort without writing. Also reconfigure `stdout`/`stderr` to UTF-8 so the tool's own emoji status `print()`s don't crash on a cp1252 console.
5. **SKILL.md**: documents the masking step, and fixes a doc/behavior mismatch — the backup was described as `FILE.original.md` beside the source, but the code actually writes it out-of-tree under `$XDG_DATA_HOME/caveman-compress/backups/...` (intentional, so auto-loaders don't re-ingest it — just undocumented).

## Why this approach

Masking makes `validate_code_blocks` / `validate_inline_codes` structurally un-failable — the spliced-back text is byte-identical to the original by construction, so those checks become a backstop rather than the primary defense. The UTF-8 gate applies the same philosophy to prose: guarantee it at the I/O boundary and fail loudly if a future regression breaks it, instead of trusting the environment. No change needed to the validator itself. Existing `MAX_RETRIES` fix-pass logic is untouched.

## Testing

```
python3 -m unittest discover -s tests
```

- Added `tests/test_mask_code.py`: round-trip fidelity through a simulated aggressive rewrite, no real code leaking into the masked prompt text, fence-first masking so inline code inside a fence isn't double-masked, and three fail-closed cases (dropped placeholder, duplicated placeholder, no-op when nothing was masked).
- Added `tests/test_unicode_prose.py`: a deterministic mask → rewrite → write → reread round-trip over a fixture containing `— → ≠ … " ' •`, asserting the written file re-reads as valid UTF-8 with every character intact (fails on the pre-fix code).
- Added a regression test to `tests/test_validate_inline.py` (`test_indented_fence_not_leaked_as_inline_code`) pinning the extractor bug above.
- The encoding fix also cleared 4 pre-existing errors in `test_compress_safety.py` (emoji status `print()`s crashing on cp1252 stdout — same root cause).
- Full suite: **23 tests pass**, 0 failures, 0 errors.
- Beyond the unit tests, also ran the real `compress_file` pipeline end-to-end (via `claude --print`) against a synthetic code-dense markdown file (inline commands + a fenced YAML block with a comment) — output validated byte-identical on fenced blocks and inline code, including preserving a YAML comment the model would otherwise be inclined to strip. Repro file available on request if useful as a fixture; kept out of the diff since it's synthetic and not needed for the unit tests to pass.

### DIFF
--- a/skills/caveman-compress/SKILL.md
+++ b/skills/caveman-compress/SKILL.md
@@ -3,7 +3,7 @@ name: caveman-compress
 description: >
   Compress natural language memory files (CLAUDE.md, todos, preferences) into caveman format
   to save input tokens. Preserves all technical substance, code, URLs, and structure.
-  Compressed version overwrites the original file. Human-readable backup saved as FILE.original.md.
+  Compressed version overwrites the original file. Original backed up out-of-tree (see backup path below).
   Trigger: /caveman-compress FILEPATH or "compress memory file"
 ---
 
@@ -11,7 +11,7 @@ description: >
 
 ## Purpose
 
-Compress natural language files (CLAUDE.md, todos, preferences) into caveman-speak to reduce input tokens. Compressed version overwrites original. Human-readable backup saved as `<filename>.original.md`.
+Compress natural language files (CLAUDE.md, todos, preferences) into caveman-speak to reduce input tokens. Compressed version overwrites original. Original is backed up out-of-tree as `<filename>.original.md` under a platform-aware data dir (`$XDG_DATA_HOME/caveman-compress/backups/...` or `%LOCALAPPDATA%\caveman-compress\backups\...`), not beside the source file — this keeps skill/rule auto-loaders from re-ingesting the backup as a live file.
 
 ## Trigger
 
@@ -27,7 +27,12 @@ python3 -m scripts <absolute_filepath>
 
 3. The CLI will:
 - detect file type (no tokens)
-- call Claude to compress
+- mechanically mask fenced code blocks and inline `code` spans with opaque placeholder
+  tokens before sending anything to Claude — this is automatic and needs no model action;
+  it exists so code can't be mangled during compression, only spliced back verbatim after
+- call Claude to compress (it sees prose + placeholders, never the real code)
+- splice the real code back in by placeholder; abort with no changes if a placeholder was
+  dropped, duplicated, or altered (fail closed — never guess where code belonged)
 - validate output (no tokens)
 - if errors: cherry-pick fix with Claude (targeted fixes only, no recompression)
 - retry up to 2 times
@@ -107,5 +112,5 @@ Compressed:
 - NEVER modify: .py, .js, .ts, .json, .yaml, .yml, .toml, .env, .lock, .css, .html, .xml, .sql, .sh
 - If file has mixed content (prose + code), compress ONLY the prose sections
 - If unsure whether something is code or prose, leave it unchanged
-- Original file is backed up as FILE.original.md before overwriting
+- Original file is backed up as `<filename>.original.md` (out-of-tree, see backup path above) before overwriting
 - Never compress FILE.original.md (skip it)
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -14,6 +14,21 @@ import sys
 from pathlib import Path
 from typing import List
 
+# Force UTF-8 for interactive console output. Windows text streams default to
+# the system codepage (commonly cp1252) when Python isn't in UTF-8 mode (see
+# the encoding notes throughout this file) -- without this, a bare print() of
+# a non-ASCII status glyph (the checkmark/cross-mark used on the pass/fail
+# paths below) raises UnicodeEncodeError and crashes the whole run instead of
+# reporting cleanly. errors="backslashreplace" only affects what a human sees
+# in a legacy console; it never touches file contents, which go through the
+# strict, gated UTF-8 file I/O below.
+for _stream in (sys.stdout, sys.stderr):
+    if hasattr(_stream, "reconfigure"):
+        try:
+            _stream.reconfigure(encoding="utf-8", errors="backslashreplace")
+        except (AttributeError, ValueError, OSError):
+            pass
+
 OUTER_FENCE_REGEX = re.compile(
     r"\A\s*(`{3,}|~{3,})[^\n]*\n(.*)\n\1\s*\Z", re.DOTALL
 )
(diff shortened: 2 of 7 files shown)


rajpratham1 (approved):
Nice improvement. Replacing prompt-only code preservation with placeholder masking makes the behavior much more robust, and the fail-closed integrity checks are a sensible safeguard. The regression fix for inline code extraction and the accompanying test coverage help make the change easier to trust. The documentation updates also stay in sync with the implementation. I didn't notice any issues in the changes shown.

FatalMerlin:
> Nice improvement. Replacing prompt-only code preservation with placeholder masking makes the behavior much more robust, and the fail-closed integrity checks are a sensible safeguard. The regression fix for inline code extraction and the accompanying test coverage help make the change easier to trust. The documentation updates also stay in sync with the implementation. I didn't notice any issues in the changes shown.

Thank you, I am glad to hear that. It was important to me to only submit this as a proper fix. I already managed to verify this "in production" by using the patch locally to correctly compress the files it had previously tripped on repeatedly :)

AmirF194:
This is a good change, and mask-and-splice is the right call: enforcing preservation structurally instead of asking the model and validating after is a real step up from the prior attempts you cite. A few things I checked.

- The `extract_inline_codes` regression test genuinely fails on main. I ran the old column-0 extractor against your indented-fence input and it returns `['sh\n  echo ', ...]` garbage instead of `['kubectl get pods']`, so that test is doing real work.
- The fail-closed logic in `unmask_code` covers dropped, duplicated, and unexpected placeholders, and aborting before backup/write is the right default.
- One scope note: the retry fix-pass (`build_fix_prompt`) runs on the already-unmasked body, so the "model never sees real code" guarantee only holds for the first pass. It stays safe because `validate()` byte-checks code blocks against the backup every iteration and restores the original on final failure, so the worst case is a clean abort rather than corruption. A one-line comment noting the guarantee is first-pass and the validator backstops the retry would save a future reader from assuming masking covers the whole loop.
- Minor: the inline regex `` `([^`\n]+)` `` won't handle double-backtick spans, but that matches the validator's own extractor, so at least they agree. Fine to leave.

Nice work, and thanks for laying out the prior art.

FatalMerlin:
@AmirF194 updated to add the requested comment :)

FatalMerlin:
Found a Unicode-corruption bug in this branch during real use on Windows while batch-compressing a set of memory files, fixed in 63008e9.

**Symptom:** compressing files that contain non-ASCII prose characters (em-dash `—`, arrows `→`, `≠`, curly quotes, bullet, ellipsis) either silently corrupted them or hard-crashed, depending on the character:

- characters that exist in cp1252 (em-dash, curly quotes, ...) were silently mis-encoded and the file became invalid UTF-8, while the run still exited 0 with "Validation passed"
- characters outside cp1252 (`→`, `≠`) raised `UnicodeEncodeError` mid-write

The mask-and-splice validator only checked code-fragment / frontmatter loss, so the silent case slipped straight through.

**Root cause:** the skill's file I/O (`compress.py`, `validate.py`, `detect.py`) called `read_text()` / `write_text()` with no explicit `encoding=`. On Windows that defaults to the locale codepage (cp1252 here, `PYTHONUTF8` unset), not UTF-8. The `claude --print` subprocess decode was already correctly pinned to UTF-8; only the file layer was missing it.

**Fix (63008e9):**

- pin `encoding="utf-8"` on every file read/write in the skill
- add a fail-closed `UnicodeIntegrityError` gate before overwrite: the compressed output must round-trip encode/decode as UTF-8 and introduce no new `U+FFFD` versus the source, else abort without writing
- reconfigure stdout/stderr to UTF-8 so the tool's own diagnostic `print()` lines (emoji status) stop crashing on cp1252 consoles (this also fixed 4 pre-existing test errors from the same root cause)

Tests: 22 → 23 (added `test_unicode_prose.py`, a deterministic mask → rewrite → write → reread round-trip over `— → ≠ … " ' •`), all green.

JuliusBrussee:

PR #621, opened 2026-07-03
feat(hooks): propagate caveman mode to subagents via SubagentStart

tompassarelli:
## Problem

SessionStart hook context reaches the parent thread only — it never propagates to Task-spawned subagents. So in an active caveman session, every subagent (Explore, custom agents, etc.) runs without the caveman ruleset and emits full-verbosity output, forfeiting the token savings exactly where output volume is highest.

## Fix

A SubagentStart hook (`src/hooks/caveman-subagent.js`) that re-reads the active caveman level from the same flag file the other hooks use and re-injects the ruleset into each spawned subagent via `hookSpecificOutput.additionalContext`. When caveman is off or the flag is absent it exits silently with no output. Fail-open everywhere: a hook error can never block a subagent spawn.

The pattern is borrowed from ponytail (MIT, [DietrichGebert/ponytail](https://github.com/DietrichGebert/ponytail)), which fixed the same parent-only-context gap for its own mode (their issue #252).

## Changes

- `src/hooks/caveman-subagent.js` — the hook (new)
- `.claude-plugin/plugin.json` — SubagentStart entry
- `bin/install.js` + `src/hooks/install.sh` — standalone-install wiring (idempotent, marker-guarded like existing hooks)
- `src/hooks/checksums.sha256` — new hook entry

## Verification

- `node --check` passes
- Flag absent → exit 0, no output
- Flag `full`/`ultra` → valid SubagentStart JSON, additionalContext starts with "CAVEMAN MODE ACTIVE — level: ..." (works both with plugin SKILL.md present and standalone fallback)
- `sha256sum -c src/hooks/checksums.sha256` passes
- `.codex/hooks.json` intentionally untouched — Codex has no SubagentStart equivalent


rajpratham1 (approved):
Looks good overall. The implementation consistently adds SubagentStart support across the plugin configuration, installers, and hook implementation, while preserving graceful fallback behavior if the hook cannot run. The approach of reusing the existing mode flag and injecting context through hookSpecificOutput.additionalContext fits the intended feature well. I didn't notice any functional issues in the changes shown. Nice addition!

PR #622, opened 2026-07-03
fix(hooks): deliver /caveman-stats via additionalContext for macOS app compat

### LINKED ISSUE #618: /caveman-stats shows nothing in Claude Code macOS app (works in terminal)
## Bug
`/caveman-stats` works in the terminal Claude Code but produces **no output** in the Claude Code macOS app. The hook fires (the per-turn caveman reminder still injects), but the stats block never appears.

## Root cause
`src/hooks/caveman-mode-tracker.js` delivers stats via `{"decision":"block","reason": <stats>}` on `UserPromptSubmit`. The terminal renders a blocked prompt's `reason`; the macOS app **honors the block (swallows the prompt) but never renders `reason`** → stats vanish silently.

## Fix
Deliver the stats via `hookSpecificOutput.additionalContext` instead and let the model print the block verbatim. `additionalContext` is honored by both the terminal and the app.

```diff
-        const out = execFileSync(process.execPath, argv, { encoding: 'utf8', timeout: 5000 });
-        process.stdout.write(JSON.stringify({ decision: 'block', reason: out.trim() }));
-      } catch (e) {
-        process.stdout.write(JSON.stringify({
-          decision: 'block',
-          reason: 'caveman-stats: could not run stats script.\nTry manually: node hooks/caveman-stats.js'
-        }));
-      }
+        const out = execFileSync(process.execPath, argv, { encoding: 'utf8', timeout: 5000 });
+        context = 'The user ran /caveman-stats. Print the following stats block ' +
+          'VERBATIM inside a code block. Do not summarize, reformat, or add ' +
+          'commentary (this overrides any active brevity/caveman style):\n\n' + out.trim();
+      } catch (e) {
+        context = 'The user ran /caveman-stats but the stats script failed. ' +
+          'Tell them to run it manually: node hooks/caveman-stats.js';
+      }
+      process.stdout.write(JSON.stringify({
+        hookSpecificOutput: {
+          hookEventName: "UserPromptSubmit",
+          additionalContext: context
+        }
+      }));

Hardik-369:
`/caveman-stats` works in terminal Claude Code but shows nothing in the macOS app.

**Root cause:** the hook delivered stats via `{decision:'block', reason:...}` on `UserPromptSubmit`. Terminal renders the blocked prompt's `reason`; the macOS app honors the block (swallows the prompt) but never renders `reason` → stats vanish silently.

**Fix:** deliver the stats via `hookSpecificOutput.additionalContext` instead, which is honored by both terminal and app. The model prints the block verbatim on behalf of the hook.

Fixes #618

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -41,9 +41,10 @@ process.stdin.on('end', () => {
     // /caveman-stats [--share] — block the prompt and inject stats output as
     // the hook's reason. The script reads the active session log, so we pass
     // transcript_path through when Claude Code provides it.
-    const statsMatch = /^\/caveman(?::caveman)?-stats(?:\s+(.*))?$/.exec(prompt);
+      const statsMatch = /^\/caveman(?::caveman)?-stats(?:\s+(.*))?$/.exec(prompt);
     if (statsMatch) {
       const tailArgs = (statsMatch[1] || '').trim().split(/\s+/).filter(Boolean);
+      let context;
       try {
         const statsPath = path.join(__dirname, 'caveman-stats.js');
         const argv = [statsPath];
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
Good catch. The implementation now uses hookSpecificOutput.additionalContext instead of blocking the prompt with reason, so the comment should be updated to reflect the new behavior. The extra indentation on const statsMatch also looks unintentional and is worth cleaning up for consistency.

JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`7693a046`](https://github.com/JuliusBrussee/caveman/commit/7693a046) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

PR #608, opened 2026-07-02
fix(hooks): NL trigger misfires + one-shot independent modes

AmirF194:
## What happen

Two mode-tracker issues, both reproduced by piping prompts into the hook:

**#598 — natural-language matching misfires:**
| Prompt | Before this PR |
|---|---|
| `turn caveman mode off` (flag=ultra) | flag **overwritten to full** — deactivation re-armed caveman and downgraded the level |
| `turn caveman off` | no-op — only the contiguous phrase "turn off" matched |
| `enable caveman and stop apologizing` | flag **deleted** — explicit enable performed a disable |
| `what is caveman mode?` | activated the mode |
| `how do I exit vim normal mode` | deactivated the session mode |
| `stop\ncaveman` (multiline) | no-op — `.*` without the `s` flag |

**#599 — one-shot modes clobber the level:** `/caveman-commit` overwrote the flag with `commit` permanently — per-turn reinforcement stopped for the rest of the session and the statusline stuck at `[CAVEMAN:COMMIT]`, despite SKILL.md's "Level persist until changed or session end". Also, only compress/stats had plugin-namespaced variants: `/caveman:caveman-commit` and `-review` were unrecognized entirely.

Fixes #598, fixes #599.

## Fix

- Deactivation intent computed **first** and word-order tolerant; "normal mode" deactivates only as a command (prompt-initial, optionally with a switch-back verb) or with caveman context; questions are not commands; prompts normalized to one line; scoped brevity ("be brief in the summary section") stays a one-off instruction.
- Independent modes save the displaced prose mode to `.caveman-active.prev` (via `safeWriteFlag`, per the CLAUDE.md rule); the next ordinary prompt restores it — or deactivates if caveman wasn't active before. Chained one-shots (`/caveman-commit` → `/caveman-review`) keep the original. Deactivation clears the saved prev. Namespaced `-commit`/`-review` variants added.

## Tests

New `tests/test_mode_tracker.py` — 23 tests covering every confirmed misfire, the README-promised triggers that must keep working ("activate caveman", "turn on caveman mode", "talk like caveman", bare "caveman mode", "be brief"), slash-command switching, and the full one-shot save/restore lifecycle. All 23 pass; `verify_repo.py` (which pins `/caveman ultra`, "normal mode", and default-off behavior) passes unchanged, locally **and in a clean node:20-bookworm container**.

Issue #601, opened 2026-07-02
caveman-stats: whole-session tokens attributed to the mode active at stats time — savings inflated or zeroed on mid-session mode change


AmirF194:
**What happen**

`/caveman-stats` attributes **all** output tokens since session start to whatever mode the flag holds at the moment stats run:

- Activate caveman at turn 51 of a 50-turn verbose session → all prior verbose output is counted as caveman-compressed, then `estNormal = tokens/0.35` is back-computed on top (`src/hooks/caveman-stats.js:310-311, 142-150`) → wildly inflated savings.
- Inverse: run caveman all session, say "normal mode", then `/caveman-stats` → mode reads null → "Caveman not active this session", zero savings.

The inflated `est_saved_tokens` is appended to `.caveman-history.jsonl` (`:316-327`) and rendered in the statusline lifetime suffix — so the error compounds into the permanent lifetime number the README treats as an honest metric ("Benchmark and eval numbers must be real").

The per-mode arithmetic itself is fine (verified synthetically: 1,000 tokens, `full`, sonnet-4-5 → 2,857 est / 1,857 saved); the defect is mode attribution over time.

**Fix direction**

Track mode transitions (the tracker already writes the flag on every change — append `{ts, mode}` to a small session log) and attribute token spans to the mode active when they were generated; or at minimum only count tokens from the timestamp the current mode was set.

main @ 25d22f8.

AmirF194:
Fixed on main by e9cb843, reviewed the implementation and it covers everything this issue asked for: every flag transition site in both hooks (slash, NL activation/deactivation, one-shot restore, SessionStart, off-mode) is paired with `recordModeChange`, the log stores `{ts, mode, prev}` with a no-op guard on unchanged mode, and `attributeByMode` joins spans against message timestamps with honest fallbacks (mid-session flag write → pre-change tokens excluded rather than guessed). Both failure cases from this report now produce correct output. This can be closed.

One small follow-up I'll file separately: the new `.caveman-mode-log.jsonl` (and the other caveman state files) survive `--uninstall`, so a later reinstall inherits stale attribution rows and a stale lifetime badge.

JuliusBrussee:
[...]

PR #623, opened 2026-07-03
fix(parse): extract shared mode parser to fix three drifts between tracker and opencode plugin

### LINKED ISSUE #602: opencode plugin mode parsing drifted from Claude tracker — brevity triggers missing, bogus level overwrites flag, independent modes unreachable
**What happen**

The opencode plugin's mode parsing (`src/plugins/opencode/plugin.js:99-161`) has drifted from the Claude Code tracker (`src/hooks/caveman-mode-tracker.js:30-39`). Three confirmed divergences (driven live with an `XDG_CONFIG_HOME` sandbox):

1. **Brevity triggers missing** — the tracker activates on "less tokens" / "fewer tokens" / "be brief" / "be terse" / "shorter answers" (promised in README). The opencode plugin has no such branch: `"be brief please"` and `"use less tokens from now on"` left the flag unchanged.
2. **Bogus level overwrites the flag** — the expanded-template branch (`plugin.js:123-130`) falls through to `return getDefaultMode()` for an unknown level, while the literal branch (`:155-156`) and the Claude tracker (`caveman-mode-tracker.js:94`) deliberately return null. Confirmed: after `/caveman wenyan-ultra`, "Activate caveman mode: bogus-level" overwrote the flag to `full`.
3. **Independent modes unreachable** — opencode expands `/caveman-commit` etc. into the command body ("Generate a commit message…"), which matches no parse branch, so the flag stays at e.g. `full` and the per-turn caveman-prose reinforcement (`plugin.js:219-225`) keeps injecting during commit/review/compress turns — exactly what the tracker's `INDEPENDENT_MODES` suppression exists to prevent.

**Fix direction**

Extract the parsing into a shared CJS module (e.g. `src/hooks/caveman-parse.js`) consumed by both the tracker and the opencode plugin — the plugin already proves this pattern works in the compiled-Bun runtime by reusing `caveman-config.js` via `createRequire`.

main @ 25d22f8.

Hardik-369:
## Problem

The opencode plugin mode parsing (`src/plugins/opencode/plugin.js:99-161`) had drifted from the Claude Code tracker (`src/hooks/caveman-mode-tracker.js`). Three confirmed divergences:

1. **Brevity triggers missing** — "less tokens" / "fewer tokens" / "be brief" / "be terse" / "shorter answers" only worked in the tracker, not the plugin.
2. **Bogus level overwrites flag** — expanded-template branch fell through to `getDefaultMode()` for unknown levels, silently overwriting the flag. Tracker and slash-command branch correctly return null.
3. **Independent modes unreachable** — opencode expands `/caveman-commit` etc. into command descriptions that match no parse branch, so the flag never flipped and per-turn reinforcement kept injecting.

## Fix

Extracted the parsing into `src/hooks/caveman-parse.js` — a shared CJS module consumed by both the tracker and the opencode plugin. `bin/install.js` updated to copy `caveman-parse.{js->cjs}` alongside `caveman-config.{js->cjs}` for the opencode plugin install path.

All three bugs are fixed in the single shared `parseModeChange()`: brevity triggers are included, unknown levels return null, and independent modes are reachable via their slash commands.

## Validation

- 28 unit tests passing covering all three bugs, deactivation, activation, slash commands, and quote unwrapping
- All four changed files pass Node.js syntax check

Closes #602

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -44,6 +44,7 @@ const MCP_SHRINK_PKG = 'caveman-shrink';
 const HOOK_FILES = [
   'package.json',
   'caveman-config.js',
+  'caveman-parse.js',
   'caveman-activate.js',
   'caveman-mode-tracker.js',
   'caveman-stats.js',
@@ -641,6 +642,8 @@ function installOpencode(ctx) {
       // sibling would be loaded as ESM and break the plugin's require() bridge.
       [path.join(repoRoot, 'src', 'hooks', 'caveman-config.js'),
        path.join(pluginDir, 'caveman-config.cjs')],
+      [path.join(repoRoot, 'src', 'hooks', 'caveman-parse.js'),
+       path.join(pluginDir, 'caveman-parse.cjs')],
     ];
     for (const [src, dest] of pluginPayload) {
       if (fs.existsSync(dest) && !opts.force) {
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -19,9 +19,10 @@ process.stdin.on('end', () => {
     const data = JSON.parse(input);
     const prompt = (data.prompt || '').trim().toLowerCase();
 
-    // /caveman-stats [--share] — block the prompt and inject stats output as
-    // the hook's reason. The script reads the active session log, so we pass
-    // transcript_path through when Claude Code provides it.
+    // /caveman-stats [--share] — run the stats script and inject its output
+    // via additionalContext so both terminal and macOS app render it.
+    // The script reads the active session log, so we pass transcript_path
+    // through when Claude Code provides it.
     const statsMatch = /^\/caveman(?::caveman)?-stats(?:\s+(.*))?$/.exec(prompt);
     if (statsMatch) {
       const tailArgs = (statsMatch[1] || '').trim().split(/\s+/).filter(Boolean);
@@ -41,8 +42,9 @@ process.stdin.on('end', () => {
           'VERBATIM inside a code block. Do not summarize, reformat, or add ' +
           'commentary (this overrides any active brevity/caveman style):\n\n' + out.trim();
       } catch (e) {
-        context = 'The user ran /caveman-stats but the stats script failed. ' +
-          'Tell them to run it manually: node hooks/caveman-stats.js';
+        const detail = e.message ? ': ' + e.message : '';
+        context = 'The user ran /caveman-stats but the stats script failed' +
+          detail + '. Tell them to run it manually: node hooks/caveman-stats.js';
       }
       process.stdout.write(JSON.stringify({
         hookSpecificOutput: {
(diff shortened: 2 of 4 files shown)


rajpratham1 (changes requested):
The overall refactoring is good—moving parsing logic into a shared module reduces duplication and future drift. However, the review comments identify real regressions, especially in the opencode plugin and /caveman-stats behavior.

Blocking issues
1. /caveman-stats behavior changed (High)  Agree with Copilot

Previously:

{ "decision": "block", "reason": "..." }

Now it emits:

hookSpecificOutput.additionalContext

without blocking.

This changes externally observable behavior:

existing tests fail
documented behavior changes
users no longer immediately receive stats output

Unless the project intentionally changed the hook contract, this is a regression.

Request: restore decision: "block" or update the entire feature/tests/docs together.

2. require('./caveman-config') in caveman-parse.js (High) ✅ Agree

The installer copies

caveman-config.cjs
caveman-parse.cjs

into the plugin.

Inside the copied parser:

require('./caveman-config')

Node does not resolve:

./caveman-config.cjs

from an extensionless CommonJS require.

That means the installed plugin can fail to load.

The suggested fallback:

try {
    require("./caveman-config.cjs")
} catch {
    require("./caveman-config")
}

is a reasonable solution.

3. Plugin doesn't pass parser options (High) ✅ Agree

The parser now supports:

unwrapQuotes
expandedTpl
opencodeCommandBodies

But the plugin calls:

parseModeChange(text)

instead of

parseModeChange(text, {
    unwrapQuotes: true,
    expandedTpl: true,
    ...
})

Meaning the new parser features are effectively disabled for opencode.

This defeats part of the refactor.

Medium issue
Expanded command bodies

Copilot notes that

/caveman-review

becomes

Review the current diff...

inside opencode.

If true, the shared parser currently won't detect these commands.

Whether this is blocking depends on how opencode expands commands today, but it's worth verifying.

AmirF194:
This is the right direction, and it is basically the shared-parser design #602 was asking for, so thanks for taking it on. I want to flag a few things before it lands, because the branch looks like it was cut before some related changes merged and the extraction picked up the older behavior.

1. The tracker on main calls `recordModeChange(claudeDir, ...)` at every flag transition (the #601 timestamped transition log that caveman-stats reads for per-mode attribution). The refactored tracker here drops all of those calls and the import, so merging this would quietly revert #601.

2. Main also keeps a `.caveman-active.prev` one-shot save/restore (#608) that isn't carried through the new path.

3. The brevity regex in `caveman-parse.js` is the pre-#608 version without the negative lookahead. Main has `...shorter answers)\b(?!\s+(in|for|on|about|when|during|with)\b)` so "be brief in the summary" doesn't flip global mode; the extracted copy loses that guard.

4. On the opencode side, the parser gates quote-unwrapping and the expanded-template branch behind `options.unwrapQuotes` / `options.expandedTpl`, but the call site (`parseModeChange(part.text)`) passes no options, so both are off. That means the two opencode-specific bugs this is meant to fix (the expanded `Activate caveman mode: <level>` template and the quoted `run`-path input) aren't handled through this path. The Claude tracker calling it with no options is correct; opencode needs `{ unwrapQuotes: true, expandedTpl: true }`.

None of this is a knock on the approach, which is solid and overdue. A rebase onto current main, re-extracting from the current tracker (keeping `recordModeChange` and `.prev`) and passing the options at the opencode call site would get it there. Happy to sanity-check once it's rebased, since I filed #602.

JuliusBrussee:

PR #624, opened 2026-07-03
fix(commands): /caveman-init works for installed users

### LINKED ISSUE #603: commands/caveman-init.toml runs repo-relative src/tools path — fails for every installed Codex/Gemini user

AmirF194:
## What happen

`commands/caveman-init.toml` tells the agent to run `node src/tools/caveman-init.js {{args}}` "in the current repo". For every installed Codex / Gemini-extension user, cwd is **their** project — that path only exists inside a caveman clone, so `/caveman-init` fails with `Cannot find module` for everyone except caveman developers.

Fixes #603.

## Fix

Two parts:

1. **New `--init-only` flag in `bin/install.js`** — writes the per-repo IDE rule files and skips everything else: the provider install matrix, the interactive TTY picker, and the npx-skills auto-detect fallback. Rationale: the TOML can't just use `--with-init`, because that runs the full agent-install matrix as a side effect — a slash command named *init* must not silently install plugins.
2. **TOML now runs the installer that works from anywhere:** `npx -y github:JuliusBrussee/caveman -- --init-only {{args}}`. `--dry-run` / `--force` pass straight through, so the existing prompt guidance ("dry-run first unless the user passed --force") keeps working.

The other three TOML stubs are prompt-only (no paths) — unaffected.

## Tests

New dry-run e2e test: `--init-only --dry-run` plans rule files, and asserts **no** provider output appears (no `Claude Code detected`, no `would run: claude plugin`, no `skills add`), and nothing is written. Also verified live in a scratch git repo (with `OPENCLAW_WORKSPACE` isolated): `.cursor/`, `.windsurf/`, `.clinerules/`, `.github/`, `.opencode/`, `AGENTS.md` written; zero agent installs triggered. Dry-run suite 3/3, `node --check` clean.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -55,7 +55,7 @@ const HOOK_FILES = [
 function parseArgs(argv) {
   const opts = {
     dryRun: false, force: false, skipSkills: false,
-    withHooks: 'auto', withInit: false, withMcpShrink: false,
+    withHooks: 'auto', withInit: false, initOnly: false, withMcpShrink: false,
     all: false, minimal: false, listOnly: false, noColor: false,
     only: [], uninstall: false, nonInteractive: false,
     configDir: null, help: false,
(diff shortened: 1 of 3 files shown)


AmirF194:
Closing, main now solves this with the standalone curl|node rewrite of caveman-init.toml (nice, self-contained approach). The `--init-only` installer flag from this branch is still around on my fork if a no-curl path is ever wanted, but the user-facing bug is gone. #603 can be closed too.

PR #625, opened 2026-07-03
test(mcp-shrink): guard npm files[] against missing relative requires

AmirF194:
## What this is now

Rescoped after a884696 landed — the maintainer's commit already ships `spawn-options.js` and bumps 0.1.1, so this PR no longer touches `package.json`. What remains is the **regression guard** so this bug class can't come back silently:

every relative `require()` reachable from the shipped js files must itself be listed in `files[]` — the test fails at CI time instead of at the next `npm publish` if a future refactor adds a helper module without shipping it (exactly how #597 happened: `spawn-options.js` was added for the Windows shell fix but `files[]` wasn't updated).

Refs #597.

## Verification

- 18/18 tests pass against current main (guard included)
- Removing `spawn-options.js` from `files[]` makes the guard fail with `index.js requires ./spawn-options.js but package.json files[] does not ship it` — verified both directions

### DIFF
--- a/tests/test_mcp_shrink.js
+++ b/tests/test_mcp_shrink.js
@@ -30,6 +30,28 @@ function test(name, fn) {
 
 console.log('mcp-shrink compress tests\n');
 
+// ── Packaging guard (issue #597) ──────────────────────────────────────────
+// Every relative require() reachable from the published entry points must be
+// listed in package.json "files", or `npm publish` ships a package that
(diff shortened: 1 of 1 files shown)


AmirF194:
No rush, just flagging this is still open a couple weeks in. Happy to tweak the approach if you'd prefer something different.

AmirF194:
Closing to keep the queue clean, happy to reopen if there's interest.

PR #606, opened 2026-07-02
fix(compress): never classify build files or shebang scripts as prose

AmirF194:
## What happen

`detect.py` lists `.dockerfile` / `.makefile` in `SKIP_EXTENSIONS`, but real files are named `Dockerfile` / `Makefile` with **no extension**, so they fall through to the content heuristic — which has no shebang or known-filename check — and come back `natural_language, compress=True`. Net effect: `/caveman-compress Dockerfile` sends the file to Claude and overwrites it with caveman prose. Confirmed for `Dockerfile`, `Makefile`, and a `#!/usr/bin/env bash` script with no extension.

Fixes #600.

## Fix

Two small guards in `detect_file_type`:

1. **Basename check before any extension rule** — `KNOWN_CODE_FILENAMES` (case-insensitive): Dockerfile, Makefile, GNUmakefile, Jenkinsfile, Vagrantfile, Rakefile, Gemfile, Justfile, Procfile, Brewfile, CMakeLists.txt. It must run before extension classification because `CMakeLists.txt` would otherwise ride the compressible `.txt` rule.
2. **Shebang check** in the extensionless branch — first two bytes `#!` → `code`, never prose.

The plugin mirror copy (`plugins/caveman/skills/caveman-compress/scripts/detect.py`) is synced in the same commit: the sync workflow only triggers on `skills/**/SKILL.md` changes, so a scripts-only change has to carry its mirror along (same as bdcba4c did).

## Tests

New `tests/test_detect.py` (7 tests): Dockerfile/Makefile/CMakeLists.txt/shebang scripts classify as code and refuse compression; case-insensitivity; and two guards that extensionless prose (`NOTES`) and `README.md` **stay** compressible. Full `tests/verify_repo.py` suite and `tests/test_compress_safety.py` still pass.

PR #626, opened 2026-07-03
fix(compress): UTF-8 I/O + atomic writes, no more wiped files

### LINKED ISSUE #533: `caveman-compress` crashes on Windows with `charmap` codec error (Unicode characters)

AmirF194:
## What happen

Every file read/write in `compress.py` / `validate.py` / `benchmark.py` used the platform default encoding. On Windows that's cp1252, so any non-ASCII character (`→`, `—`, accented text, CJK) crashed with `UnicodeEncodeError` — and because `Path.write_text` **opens-and-truncates the destination before encoding the string**, the crash left the user's live file at 0 bytes. At that point the backup readback hadn't completed its job either, so there was no recovery path.

Reproduced in a clean `node:20-bookworm` container by forcing the same failure class (`LC_ALL=C PYTHONUTF8=0` → ASCII default encoding, mocked `call_claude`):

| | result | live file after |
|---|---|---|
| `main` | `UnicodeEncodeError: 'ascii' codec can't encode '\u2192'` | **0 bytes (wiped)** |
| this branch | `ok=True` | 19 bytes, `A → B done.` intact |

Fixes #533.

## Fix

- **`encoding='utf-8'`** pinned on every file read (`compress.py:249,304`, `validate.py:31`, `benchmark.py:26-27`). The repo already pins UTF-8 for the CLI subprocess and stdout — file I/O was the gap.
- **New `write_text_atomic()`** used for all four writes (backup, live write, retry write, restore): encode **first** (an encode error can no longer touch the file), write to a temp file in the same directory, fsync, `os.replace` — atomic on POSIX and Windows. Failure at any step leaves the destination untouched and cleans up the temp file.
- **Fix-retry loop guarded against `call_claude` returning `None`/empty** — previously `Path.write_text(None)` raised `TypeError` *after* truncating the live file (same data-loss class). Now it skips the write and lets the retry cap restore the original.
- `detect.py:79` deliberately left alone: reads with `errors='ignore'` never crash, it's classification-only, and PR #606 already touches that region — avoiding a cross-PR conflict.
- Plugin mirror (`plugins/caveman/skills/caveman-compress/scripts/`) synced in the same commit, since the CI sync workflow only triggers on `SKILL.md` changes.

## Tests

Four new tests in `tests/test_compress_safety.py`: UTF-8 bytes on disk; **write failure leaves destination untouched + no temp-file litter** (surrogate encode error); full unicode round-trip through `compress_file` (file + backup); `None` fix-retry output restores the original instead of crashing. Full safety suite passes locally (exit 0) and in the container.

### DIFF
--- a/plugins/caveman/skills/caveman-compress/scripts/benchmark.py
+++ b/plugins/caveman/skills/caveman-compress/scripts/benchmark.py
@@ -23,8 +23,8 @@ def count_tokens(text):
 
 
 def benchmark_pair(orig_path: Path, comp_path: Path):
-    orig_text = orig_path.read_text()
-    comp_text = comp_path.read_text()
+    orig_text = orig_path.read_text(encoding="utf-8", errors="ignore")
+    comp_text = comp_path.read_text(encoding="utf-8", errors="ignore")
 
     orig_tokens = count_tokens(orig_text)
     comp_tokens = count_tokens(comp_text)
--- a/plugins/caveman/skills/caveman-compress/scripts/compress.py
+++ b/plugins/caveman/skills/caveman-compress/scripts/compress.py
@@ -11,6 +11,7 @@ import re
 import shutil
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 from typing import List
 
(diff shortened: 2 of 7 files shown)


AmirF194:
No rush, just checking in. Happy to rebase or split the fix if that makes review easier.

AmirF194:
Closing to keep the queue clean, happy to reopen if there's interest.

PR #262, opened 2026-04-22
fix(security): add SHA-256 integrity verification for downloaded hook files

xiaolai:
> **Automated audit**: This PR was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the diff on its merits.

## Bug

`hooks/install.sh` downloads five hook JS files via `curl` to `~/.claude/hooks/` with no integrity verification:

```bash
curl -fsSL "$REPO_URL/$hook" -o "$HOOKS_DIR/$hook"
```

If an attacker performs a network-level MITM attack or temporarily compromises a CDN serving the raw GitHub content, the modified files would be silently installed and executed as hooks in every future Claude Code session.

**Why it matters**: The hook files (`caveman-activate.js`, `caveman-mode-tracker.js`, `caveman-config.js`) run in every Claude Code session via `SessionStart` and `UserPromptSubmit` hooks. They have read/write access to `~/.claude/`. A tampered file at install time gives persistent access. The existing code has excellent runtime security hygiene (`O_NOFOLLOW`, VALID_MODES whitelist, ANSI-stripping) — this PR adds the missing install-time protection.

## Fix

**New file**: `hooks/checksums.sha256` — SHA-256 hashes of the five hook files at v1.6.0.

**Modified**: `hooks/install.sh` — for remote installs (curl-based), downloads `checksums.sha256` first, then verifies each file before writing it to `~/.claude/hooks/`:

- Supports both `sha256sum` (Linux) and `shasum -a 256` (macOS)
- Falls back silently if neither tool is available (preserves compatibility)
- On mismatch: removes the file, prints a clear error message, exits 1
- Local installs (from `SCRIPT_DIR`) are unaffected
- Checksums file is cleaned up after use

**Maintenance**: Update `hooks/checksums.sha256` alongside the hook files on each release. A CI step that runs `sha256sum hooks/*.js hooks/package.json hooks/*.sh > hooks/checksums.sha256` before tagging would automate this.

**Files changed**: `hooks/checksums.sha256` (new), `hooks/install.sh`

**Security finding**: Medium (network fetch + file write without integrity check — NLPM security scan finding #1)

PR #627, opened 2026-07-03
fix(install): verify caveman-init.js integrity before executing (#262)

kapil971390:
While reviewing recent commits, I noticed that the SHA-256 integrity check added in commit f0dd780 (issue #262) covers the seven hook files in `HOOK_FILES` but leaves `caveman-init.js` unprotected.

**Root cause**

In the curl-pipe fallback path (detached install, no local clone), `runInit()` downloads `src/tools/caveman-init.js` from the pinned release ref and immediately executes it — with no checksum verification:

```js
// bin/install.js lines 973-975 (before)
const tmp = path.join(os.tmpdir(), `caveman-init-${process.pid}.js`);
await downloadTo(INIT_SCRIPT_URL, tmp);
const r = child_process.spawnSync(absoluteNodePath(), [tmp, ...args], { stdio: 'inherit' });
```

The hook-file loop (lines 832-843) fetches `checksums.sha256`, computes `sha256File(dest)`, and aborts on mismatch. `runInit()` has no equivalent — a tampered download would be executed.

**Fix**

Two changes:

1. `src/hooks/checksums.sha256` — add the SHA-256 entry for `caveman-init.js` so the remote manifest covers it:

```
fa5cfbee948801ce8ea280af2a39e79807438fce729f653740a40f0a5cb9bf2a  caveman-init.js
```

2. `bin/install.js` `runInit()` — call `loadRemoteHookChecksums()` + `sha256File()` after the download and before `spawnSync`, mirroring the pattern used for hook files:

```js
// After
await downloadTo(INIT_SCRIPT_URL, tmp);
const checksums = await loadRemoteHookChecksums();
if (checksums) {
  const want = checksums.get('caveman-init.js');
  const got  = sha256File(tmp);
  if (!want || want !== got) {
    try { fs.unlinkSync(tmp); } catch (_) {}
    warn(`  integrity check failed for caveman-init.js ...`);
    return false;
  }
} else {
  warn(`  note: no integrity manifest at ${PINNED_REF} — caveman-init.js executed unverified`);
}
const r = child_process.spawnSync(absoluteNodePath(), [tmp, ...args], { stdio: 'inherit' });
```

When the manifest is unavailable (pinned ref predates #262), falls through with a warning — same back-compat behaviour as hook files.

The local-clone fast path (`repoRoot` set, `src/tools/caveman-init.js` exists) is unaffected — it runs from a package-verified copy and never downloads.

**Reproduction / verification**

Environment: macOS, Node v23.11.0

```
cd /tmp/caveman
node -e "
const fs = require('fs'); const crypto = require('crypto'); const path = require('path');
function parseChecksums(txt) {
  const map = new Map();
  for (const line of txt.split('\n')) {
    const m = line.trim().match(/^([0-9a-fA-F]{64})\s+\*?(.+)$/);
    if (m) map.set(path.basename(m[2].trim()), m[1].toLowerCase());
  }
  return map;
}
const cs = parseChecksums(fs.readFileSync('src/hooks/checksums.sha256','utf8'));
const want = cs.get('caveman-init.js');
const got = crypto.createHash('sha256').update(fs.readFileSync('src/tools/caveman-init.js')).digest('hex');
console.log('entry present:', cs.has('caveman-init.js'));
console.log('hashes match:', want === got);
"
```

Edge cases verified:

| Case | Input | Before | After |
|---|---|---|---|
| Valid download | matching hash | executed without check | hash verified, executed |
| Tampered download | hash mismatch | executed anyway | blocked, tmp deleted, returns false |
| Local clone present | repoRoot set | uses local copy (no download) | unchanged |
| Manifest unavailable | old pinned ref | executed without check | executed with warning (back-compat) |
| Hook files | caveman-activate.js etc | verified via #262 | still verified, unchanged |
| checksums.sha256 entry | caveman-init.js | absent | present |

`npm test` — 69 pass, 0 fail (unchanged).

**Out of scope**

This PR does not change behaviour when `CAVEMAN_REF` is set to an arbitrary branch — that path already bypasses the pinning guarantee and is documented as a testing override. The manifest-unavailable warning covers that case.


src/hooks/checksums.sha256
```diff
@@ -5,3 +5,4 @@ c0b77891d1f8aaef3a7fbf76533ba63a42eaa96431fb7ece611ccff7474de2ba  caveman-config
 be1129b422b7d3edfa8c97df3d8a81f1d1c94cf4e1fd7d6f3a253afebd93fed7  caveman-stats.js
 d2deff457d0a5d8e1848193e6af6a68a0ebdba4fbdf250889400d5ea231e088f  caveman-statusline.sh
 1690c639f05940cbff39e0383a27053898b30c224aa651043db29b2842cb524a  caveman-statusline.ps1
+fa5cfbee948801ce8ea280af2a39e79807438fce729f653740a40f0a5cb9bf2a  caveman-init.js
```
AmirF194: I checked this against current main. The checksum you added for caveman-init.js does not match what is on main right now:

```
sha256sum of main's src/tools/caveman-init.js: 1fd06cee3973934793120314fc4bbb635153dc37592d17f521a8faa49729931e
entry added here:                              fa5cfbee948801ce8ea280af2a39e79807438fce729f653740a40f0a5cb9bf2a
```

Both files are fetched from the same pinned release ref at install time, so merging this as is would ship a manifest that fails runInit's own check for everyone, since the check refuses to execute on a mismatch.

Also, runInit on main has moved past the tmp-path shape this patch assumes. It now tries a local copy first and downloads through privateTmpDir() for the curl fallback, added because the old predictable path could be pre-planted as a symlink and then executed. Wiring the integrity check into that shape, not just refreshing the hash, would keep both fixes.

kapil971390:
@AmirF194 thanks for catching this — both fixed:

1. Recomputed the checksum against the actual current `caveman-init.js` (`1fd06cee...`, matches both `origin/main` and the `v2.6.0` pinned tag it's fetched from at install time — verified they're identical).

2. Rewired the check into the current `privateTmpDir()`-based shape instead of the old predictable-path assumption, reusing the existing `loadRemoteHookChecksums()` / `sha256File()` helpers.

While re-verifying end-to-end against the real pinned `v2.6.0` manifest, found one more thing worth flagging: the manifest at that ref exists but has no `caveman-init.js` entry yet (true for every release up to and including this one). If I'd mirrored the `HOOK_FILES` loop's exact logic, that would hard-fail every curl-pipe install as "not in manifest" until a new release ships the updated manifest. Added a distinction for that case — present-but-missing entry now warns and proceeds (same as no manifest at all), and only a present-but-*wrong* entry blocks execution. Verified the warn-and-proceed path for real (a detached copy of `bin/install.js` run outside the repo, so it takes the curl-pipe path against the live `v2.6.0` manifest), and the correct-hash/tampered-hash cases with isolated checks.

`npm test`: 245 tests, 241 pass, 0 fail, 4 skipped (same as main).

AmirF194:
Confirmed, the recomputed checksum and the privateTmpDir() rewiring both match main. Good catch on the present-but-missing manifest entry too.

PR #632, opened 2026-07-03
fix(hooks): deliver /caveman-stats via additionalContext for macOS app compatibility

udbhav968-creator:
### Summary`nFix `/caveman-stats` output in Claude Code macOS by sending stats through `hookSpecificOutput.additionalContext` instead of `decision: "block"` + `reason`.`n`n### What changed`n- Updated `src/hooks/caveman-mode-tracker.js``n- `/caveman-stats` now emits `additionalContext``n- Added explicit instruction to print stats verbatim in a code block``n- Preserved fallback behavior when the stats script fails`n`n### Why`n- Claude Code macOS honors blocked prompts but does not render `reason``n- The terminal worked before, but macOS app swallowed the stats output``n- This fix makes the command visible in both clients`n`n### Related`n- Fixes issue `#618``n- Complements PR `#505`

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -70,9 +70,10 @@ process.stdin.on('end', () => {
       }
     }
 
-    // /caveman-stats [--share] — block the prompt and inject stats output as
-    // the hook's reason. The script reads the active session log, so we pass
-    // transcript_path through when Claude Code provides it.
+    // /caveman-stats [--share] — deliver stats as hook-specific additional
+    // context so the Claude Code macOS app renders the output, too.
+    // The script reads the active session log, so we pass transcript_path
+    // through when Claude Code provides it.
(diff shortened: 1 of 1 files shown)


AmirF194:
The mechanism is right, and it matches how the tracker already delivers per-turn reinforcement (`hookSpecificOutput.additionalContext`), so it should render in the macOS app where `decision` + `reason` doesn't.

Two things. This overlaps #622, which makes the same change for the same issue; #622 restructures so both the success and failure branches flow through one `additionalContext` write, whereas here the pre-existing outer `catch` still emits `decision: 'block'` on a non-exec error, which is the exact path that won't render in the app. Narrow, but it is the case you're fixing. Might be worth you, #622's author, and the maintainer settling on one.

Separately, the PR description is showing literal `` `n `` sequences instead of line breaks (looks like PowerShell backtick-n leaked in); a quick edit would make it readable.

JuliusBrussee:

PR #634, opened 2026-07-03
fix(hooks): respect repo-local defaultMode=off in per-turn reinforcement

AmanKRoy:
## Problem

The UserPromptSubmit flag file (`$CLAUDE_CONFIG_DIR/.caveman-active`) is global, shared across all concurrent Claude Code sessions on a machine. If session A (in some directory) activates caveman mode, it writes that flag. A concurrent session B sitting in a repo whose `.caveman.json` sets `defaultMode: "off"` would still see B's stale/foreign flag and re-inject the per-turn "CAVEMAN MODE ACTIVE" reinforcement context on every prompt — silently defeating the repo-local off config.

Separately, `getDefaultMode()` always resolved the repo-local config walk from `process.cwd()` (the hook process's spawn cwd) rather than the cwd of the prompt actually being handled, so directory-scoped config could miss mid-session directory changes.

## Fix

- `src/hooks/caveman-config.js`: `getDefaultMode()` gains an optional `startDir` param, used as the start of the repo-local config walk (`findRepoConfigPath(startDir || process.cwd())`). Fully backward compatible — existing callers passing nothing behave identically.
- `src/hooks/caveman-mode-tracker.js`:
  - Natural-language activation now calls `getDefaultMode(data.cwd)`, using the per-prompt `cwd` from the hook's stdin JSON instead of the hook process's spawn cwd.
  - The per-turn reinforcement emit is additionally gated on `getDefaultMode(data.cwd) !== 'off'`, so a foreign/stale global flag can no longer override a directory's repo-local `off` config.

## Verification

Ran the full existing suite against the patched clone — all pass, no regressions:
- `npm test` — 112/112 pass
- `python3 -m unittest tests.test_compress_safety` — OK
- `node tests/test_caveman_init.js` — 8/8 pass
- `node tests/test_symlink_flag.js` — 12/12 pass

Additionally hand-verified the exact scenario this fixes, using a scratch `CLAUDE_CONFIG_DIR` and a test dir with `.caveman.json` = `{"defaultMode":"off"}`, piping JSON directly into `caveman-mode-tracker.js`:

| # | stdin | flag before | Expected | Result |
|---|---|---|---|---|
| A | `{"prompt":"hello there","cwd":"<off-dir>"}` | `full` | no output | ✅ no output |
| B | `{"prompt":"hello there","cwd":"<normal-dir>"}` | `full` | JSON `additionalContext` emitted | ✅ emitted |
| C | `{"prompt":"be brief please","cwd":"<off-dir>"}` | none | flag NOT created | ✅ not created |
| D | `{"prompt":"be brief please","cwd":"<normal-dir>"}` | none | flag created, contents `full` | ✅ created, `full` |

Also smoke-tested `caveman-activate.js` (SessionStart hook, unaffected by this change) run from the off-dir vs. a normal dir to confirm no regression: off-dir → no flag written, mode `off`; normal dir → flag written, `full`.

## Backward compatibility

`getDefaultMode()` is called elsewhere (e.g. `caveman-activate.js`) with no argument — `startDir` is optional and falls back to `process.cwd()`, so those call sites are unaffected.

### DIFF
--- a/src/hooks/caveman-config.js
+++ b/src/hooks/caveman-config.js
@@ -87,7 +87,11 @@ function readModeFromConfigFile(configPath) {
   return null;
 }
 
-function getDefaultMode() {
+// `startDir` (optional) overrides where the repo-local config walk begins.
+// Hooks that receive a per-event `cwd` on stdin (e.g. UserPromptSubmit) should
+// pass it so directory-scoped config tracks mid-session directory changes
+// instead of the hook process's spawn cwd.
+function getDefaultMode(startDir) {
   // 1. Environment variable (highest priority)
   const envMode = process.env.CAVEMAN_DEFAULT_MODE;
   if (envMode && VALID_MODES.includes(envMode.toLowerCase())) {
@@ -95,7 +99,7 @@ function getDefaultMode() {
   }
 
   // 2. Repo-local config (checked-in, per-project default)
-  const repoConfigPath = findRepoConfigPath(process.cwd());
+  const repoConfigPath = findRepoConfigPath(startDir || process.cwd());
   if (repoConfigPath) {
     const repoMode = readModeFromConfigFile(repoConfigPath);
     if (repoMode) return repoMode;
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -62,7 +62,7 @@ process.stdin.on('end', () => {
           /\bcaveman\s+mode\s+(on|please|now)\b/.test(prompt) ||
           /^caveman(\s+mode)?\s*[.!]*$/.test(prompt) ||
           /\b(less tokens|fewer tokens|be brief|be terse|shorter answers)\b(?!\s+(in|for|on|about|when|during|with)\b)/.test(prompt)) {
-        const mode = getDefaultMode();
+        const mode = getDefaultMode(data.cwd);
         if (mode !== 'off') {
           recordModeChange(claudeDir, mode); // #601: timestamped transition log
           safeWriteFlag(flagPath, mode);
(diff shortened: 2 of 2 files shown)


AmirF194:
Nicely verified, and the scenario table makes it easy to follow. The optional `startDir` on `getDefaultMode()` is clean and backward compatible.

One thing I'd want the maintainer's read on before it lands: gating the reinforcement emit on `getDefaultMode(data.cwd) !== 'off'` also suppresses reinforcement for an explicit in-session activation. If someone runs `/caveman ultra` inside a repo whose config says `off`, the flag still gets set to ultra, but the per-turn anchor is dropped every turn, so a competing plugin can drift the style back with nothing re-anchoring it. Is repo-default-off meant to override an explicit activation like that? If not, this may want to key off the flag's origin rather than the repo default.

Also worth noting #542 is going after the root cause you describe in the problem statement (the global flag shared across sessions) by isolating the flag per session; if that lands, the concurrent-foreign-session motivation here mostly goes away, though the cwd-tracking half stands on its own. #532 is in the same area.

AmanKRoy:
Thanks for the review — appreciate you digging through the overlap with #542/#532, saved me from missing it.

On the explicit-activation question: good catch, and no — repo-off shouldn't silently swallow an explicit `/caveman ultra` in-session. Current gate (`getDefaultMode(data.cwd) !== 'off'`) can't tell "stale/foreign flag from another session" apart from "user just activated it here," since both look like `activeMode` set + repo default off. Agree this needs to key off the flag's origin, not just repo default — will push a follow-up that tags the flag with how it was set (explicit command vs. natural-language/inherited) so the gate only suppresses the latter.

On #542/#532 overlap: this PR is scoped narrower than either — it's fixing `getDefaultMode()` resolving repo-local config from the hook's spawn `cwd` instead of the per-event `cwd` on stdin, which is a bug independent of how the flag file itself is stored. #542's per-session flag files remove the concurrent-foreign-session motivation from the problem statement, but its current diff doesn't carry any off-gate into the reinforcement check at all (looks like a rebase gap, not intentional) — the reinforcement block after #542 lands is otherwise unguarded again either way. #532 and this PR touch the identical line in `caveman-mode-tracker.js`, so those two will conflict directly regardless of order.

Given the overlap, happy to rebase this on top of whichever of #542/#532 lands first — just want maintainer's call on merge order before I do, since re-deriving `getDefaultMode(startDir)` on top of #542's per-session flagPath changes is a different diff than rebasing on current main.

JuliusBrussee:

PR #636, opened 2026-07-03
fix(uninstall): remove hook/stats state files from config dir

### LINKED ISSUE #635: uninstall leaves all caveman state files in $CLAUDE_CONFIG_DIR — stale lifetime badge and mode-log attribution on reinstall
**What happen**

`--uninstall` strips settings.json wiring and deletes hook files, but never removes the state files the hooks/stats own in $CLAUDE_CONFIG_DIR. Verified on main (0d95a81): seeded all five, ran `node bin/install.js --uninstall --non-interactive --config-dir <tmp>`, all five remain:

```
.caveman-active
.caveman-active.prev
.caveman-history.jsonl
.caveman-mode-log.jsonl
.caveman-statusline-suffix
```

(`uninstall()` in bin/install.js removes hooks + settings entries only; the opencode section does remove *its* flag at ~/.config/opencode/.caveman-active, so the Claude one looks like an oversight rather than a policy.)

**Why it matters**

- Reinstall later → `.caveman-mode-log.jsonl` still holds pre-uninstall transition rows, which the new #601 attribution joins against fresh sessions; `.caveman-statusline-suffix` immediately renders the old lifetime number as if it were current.
- `.caveman-active` left set means caveman reads as armed the moment hooks come back, even if the user had turned it off before uninstalling.
- Plain hygiene: uninstall should not leave five dotfiles behind.

**Expected**

Uninstall removes the ephemeral state files it owns (`.caveman-active`, `.caveman-active.prev`, `.caveman-mode-log.jsonl`, `.caveman-statusline-suffix`). `.caveman-history.jsonl` is arguably user data (lifetime stats) — keeping it with a printed note, or removing it too, both defensible; PR incoming proposes keep + note.

**Platform**

All (macOS/Linux/Windows — same code path).

AmirF194:
## What happen

`--uninstall` removes `.caveman-active` but leaves the newer state files behind in `$CLAUDE_CONFIG_DIR`:

- `.caveman-active.prev` — one-shot mode restore (from #608)
- `.caveman-mode-log.jsonl` — the #601 mode-transition log (e9cb843)
- `.caveman-statusline-suffix` — pre-rendered lifetime badge

On a later reinstall, the stale mode-log feeds transition rows from the previous install into the #601 stats attribution, and the suffix resurrects the old lifetime number immediately. Verified in a clean `node:20-bookworm` container (details and an honest correction of my original repro in #635).

Fixes #635.

## Fix

Consolidates state-file removal into one block in `uninstall()` (replacing the lone `// Flag file` removal at the tail): `.caveman-active`, `.caveman-active.prev`, `.caveman-mode-log.jsonl`, `.caveman-statusline-suffix` are removed; `.caveman-history.jsonl` is **deliberately kept** — it's the user's lifetime stats record — with a printed `kept … delete manually if unwanted` note so the choice is visible rather than silent. Happy to flip that to remove-by-default if you'd rather uninstall leave zero trace.

## Tests

New e2e test (runs without the `claude` CLI — empty PATH, so it exercises everywhere): seeds all five files + a hook file, runs `--uninstall`, asserts the four ephemeral files and the hook are gone, history kept, and the note printed. Container run: 51 tests, 47 pass, 4 skipped (claude-CLI-dependent), 0 fail; live container round-trip leaves only `.caveman-history.jsonl`.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -1182,6 +1182,24 @@ function uninstall(ctx) {
     // Don't rmdir hooksDir — other plugins may use it.
   }
 
+  // State files the hooks/stats own (issue #635). Left behind, they poison a
+  // later reinstall: .caveman-mode-log.jsonl feeds stale transition rows into
+  // the #601 attribution, .caveman-statusline-suffix resurrects the old
+  // lifetime badge, and a set .caveman-active re-arms caveman immediately.
+  // .caveman-history.jsonl stays — it's the user's lifetime stats record —
+  // with a note so the choice is visible.
+  const STATE_FILES = ['.caveman-active', '.caveman-active.prev', '.caveman-mode-log.jsonl', '.caveman-statusline-suffix'];
+  for (const f of STATE_FILES) {
+    const p = path.join(configDir, f);
+    if (!fs.existsSync(p)) continue;
+    if (!opts.dryRun) { try { fs.unlinkSync(p); } catch (_) {} }
+    note(`  removed ${p}`);
+  }
+  const historyPath = path.join(configDir, '.caveman-history.jsonl');
+  if (fs.existsSync(historyPath)) {
+    note(`  kept ${historyPath} (lifetime stats) — delete manually if unwanted`);
+  }
+
   // Plugin uninstall on Claude. Probe `plugin list` first so a re-run on a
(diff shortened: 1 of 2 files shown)


AmirF194:

PR #639, opened 2026-07-03
feat(compress): --undo restores the pre-compression backup

AmirF194:
## What

`caveman-compress` keeps a backup, but restoring it means manually copying a file out of a platform-specific data dir most users will never find. New command:

```
python3 -m scripts --undo <file>
```

Finds the backup (out-of-tree #420 layout first, legacy in-tree `<file>.original.md` sibling as fallback), restores **byte-exact** via atomic temp+`os.replace`, verifies the readback, then removes the backup — `compress_file` refuses to run while a backup exists, so keeping it would block the next compression of the same file. Clear error (file untouched) when there's nothing to undo. SKILL.md documents the trigger ("undo/restore/revert a compression"); mirror synced same-commit.

Deliberately implemented entirely in `cli.py` so it can't conflict with #626 (which touches `compress.py`).

## Tests

4 new tests: out-of-tree restore + backup removal, legacy sibling fallback, no-backup error path leaves the file alone, missing file errors. Validated in a clean `node:20-bookworm` container: 9/9 (undo + compress-safety suites), plus a live round-trip — unicode content restored byte-exact.

### DIFF
--- a/plugins/caveman/skills/caveman-compress/SKILL.md
+++ b/plugins/caveman/skills/caveman-compress/SKILL.md
@@ -17,6 +17,10 @@ Compress natural language files (CLAUDE.md, todos, preferences) into caveman-spe
 
 `/caveman-compress <filepath>` or when user asks to compress a memory file.
 
+## Undo
+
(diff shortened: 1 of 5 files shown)


AmirF194:

Issue #638, opened 2026-07-03
proposal: caveman-lint — make wasted tokens visible and measurable


AmirF194:
## Idea

A read-only CLI that reports where an AI answer (or any prose file) wastes tokens — the measurement counterpart to caveman's compression:

```
caveman-lint: answer.md
────────────────────────────────────────
Total tokens:   412  (tiktoken cl100k_base)
Wasted tokens:  118  (~29% of document)
────────────────────────────────────────
  greeting               2×  ~21 tokens
      · Sure! I'd be happy to help you with that
  hedging                3×  ~24 tokens
      · It's important to note that
  boilerplate            2×  ~19 tokens
  filler-word           11×  ~11 tokens
  duplicate-sentence     1×  ~43 tokens
────────────────────────────────────────
  advisory: 1 code block over 80 lines — prefer a diff unless the
  full file was requested. Not counted as waste.
```

It scans for the same waste families SKILL.md tells models to drop, so it doubles as a regression check on caveman's own outputs (evals integration could come later).

## Honesty by construction (project policy baked in)

- fenced/inline code is never counted as waste
- overlapping matches are span-claimed — nothing counted twice
- token counts via tiktoken when installed, else chars/4 — the report says which was used
- oversized code blocks are advisory only, excluded from the waste total (the user may have asked for the full file)

## Status

Working implementation with 7 tests, validated in a clean container — PR incoming referencing this issue. Placement: `src/tools/caveman-lint.py` (pure stdlib, tiktoken optional), same spot as caveman-init. Zero new dependencies.

Issue #637, opened 2026-07-03
proposal: diff-first output rule — never reprint a full file for a local change


AmirF194:
## Idea

The biggest remaining token waste isn't wording — it's assistants reprinting a 300-line file to change one line. Caveman's SKILL.md already bans decorative fluff and raw log dumps; the same principle extends naturally to code output:

**Proposed addition to `skills/caveman/SKILL.md` Rules section** (exact wording for your edit/veto):

> Code changes: show smallest diff, not whole file. Patch block, changed function, or `- old` / `+ new` lines with 1-2 context lines. Full file only when user ask, file is new, or most of file change anyway.
>
> Follow-up answers: only the delta. No restating what previous answer already establish.

## Where the savings land (honest scoping)

- **Chat-style assistants** (Copilot chat, web UIs, opencode ask-mode): large — reprinting files is their default failure mode.
- **Agentic tools** (Claude Code): smaller — edits go through tool calls that never hit chat output. The rule still helps for explanations and review commentary.

Worth stating that split in README/docs if adopted — fits the docs/HONEST-NUMBERS.md ethos.

## Why an issue instead of a PR

SKILL.md is the product; wording changes belong to you. Happy to turn the blessed wording into a PR (SKILL.md + mirror sync + eval prompt exercising the rule) — or you just commit it directly, also fine.

Related: the auto-clarity rule already covers when NOT to compress; this covers what never needed printing at all.

PR #640, opened 2026-07-03
feat(tools): caveman-lint — report wasted tokens in AI prose

AmirF194:
Implements #638 — see the issue for the full pitch and honesty rules.

## What

`python3 src/tools/caveman-lint.py <file...> [--json] [--max-code-lines N]` — read-only report of wasted tokens per category (greeting, apology, hedging, boilerplate, filler-word, duplicate-heading, duplicate-sentence), the same waste families SKILL.md tells models to drop.

Design points worth review attention:

- **Span-claiming**: overlapping matches (`Sure! I'd be happy to help you with that` also contains `I'd be happy to` and `Sure!`) are counted once — greedy left-to-right, longest-first. Without this the waste total overstates ~30% on typical AI prose, which would violate the honest-numbers policy.
- **Fence-aware**: code blocks and inline code are stripped before matching; a fenced block full of trigger phrases produces zero findings (tested).
- **Token counting**: tiktoken (cl100k_base) when installed, chars/4 otherwise — the report labels which was used.
- **Oversized code blocks** (default >80 lines) are an *advisory* line, excluded from the waste total — the user may have asked for the full file. (Pairs with the diff-first proposal in #637.)

Pure stdlib, tiktoken optional, no new dependencies. Placement mirrors caveman-init (`src/tools/`).

## Tests

7 tests: category detection on verbose prose, clean caveman prose yields zero waste, fence trap, advisory-not-waste for big code blocks, short-repeat exemption, `--json` CLI contract, missing-file exit code. Green locally and in a clean container.

### DIFF
--- a/src/tools/caveman-lint.py
+++ b/src/tools/caveman-lint.py
@@ -1,0 +1,250 @@
+#!/usr/bin/env python3
+"""caveman-lint — report wasted tokens in AI-generated (or any) prose.
+
+Usage:
+    python3 src/tools/caveman-lint.py <file> [<file> ...] [--json] [--max-code-lines N]
+
+Scans prose for the waste patterns caveman exists to eliminate (greetings,
+apologies, hedging, filler, boilerplate transitions, duplicated content) and
+reports estimated token waste per category. Makes waste visible and measurable
+— it never edits the file.
+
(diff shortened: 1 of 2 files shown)


AmirF194:

PR #641, opened 2026-07-03
feat(cli): caveman config show/get/set/unset

AmirF194:
## What

The repo-local and user config files already drive the default mode (the `getDefaultMode` resolution chain), but changing them means hand-editing JSON in a platform-specific path (`~/.config/caveman/`, `%APPDATA%\caveman\`, `.caveman/config.json`). CLI sugar over the exact same files:

```
npx -y github:JuliusBrussee/caveman -- config show              # resolution chain + effective mode
npx -y github:JuliusBrussee/caveman -- config set intensity ultra
npx -y github:JuliusBrussee/caveman -- config set mode full --project
npx -y github:JuliusBrussee/caveman -- config unset mode
```

- Canonical key is `defaultMode` (what the hooks read); `mode` and `intensity` are accepted aliases since the README speaks in intensities.
- Values validated against `VALID_MODES` — a typo exits 2 and writes nothing, so a bad value can never silently break activation.
- `--project` updates the **nearest** checked-in config via the existing walk-up (`findRepoConfigPath`) instead of sprinkling new `.caveman/` dirs in subdirectories.
- Reuses the JSONC-tolerant atomic read/write from `bin/lib/settings.js`; resolution logic is required from `src/hooks/caveman-config.js`, not duplicated.

## Tests

4 e2e tests, the key one being the full loop: `config set intensity ultra` → `config get mode` returns `ultra` → **the real SessionStart hook writes `ultra` to the flag file**. Plus: subdir `--project` updates the same parent config (no second file created), invalid mode/unknown key reject without writing, unset + show. Green locally and in a clean `node:20-bookworm` container (live round-trip included).

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -1407,13 +1407,118 @@ EXAMPLES
   npx -y github:JuliusBrussee/caveman -- --all               # all the trimmings
   npx -y github:JuliusBrussee/caveman -- --only claude --no-mcp-shrink
   npx -y github:JuliusBrussee/caveman -- --uninstall
+  npx -y github:JuliusBrussee/caveman -- config show     # mode resolution
+  npx -y github:JuliusBrussee/caveman -- config set intensity ultra
+  npx -y github:JuliusBrussee/caveman -- config set mode full --project
 
   Issues: https://github.com/${REPO}/issues
 `);
 }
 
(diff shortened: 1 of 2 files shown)


AmirF194:

PR #646, opened 2026-07-04
fix(install): guard non-array hook events in rewrite

SEPURI-SAI-KRISHNA:
## What

Installer crash mid-run on malformed `settings.json`. `installHooks` call `rewriteLegacyManagedHookCommands` **before** `validateHookFields`; a hook event that is object/string instead of array survive JSONC parse, then `for (const entry of settings.hooks[ev])` throw:

TypeError: settings.hooks[ev] is not iterable

Whole installer die unhandled — hook files copied, settings.json never wired, no summary. Exactly the malformation `validateHookFields` exist to clean.

 ## Fix

`Array.isArray` guard, skip malformed event, validation still clean it before write. Same guard `removeCavemanHooks` already got for identical crash (see existing test at `unit.settings.test.mjs:137`).

## Tests

New case in `tests/installer/unit.settings.test.mjs`: malformed events not throw, well-formed sibling still rewritten. `npm test`: 113/113 pass.

### DIFF
--- a/bin/lib/settings.js
+++ b/bin/lib/settings.js
@@ -253,6 +253,10 @@ function rewriteLegacyManagedHookCommands(settings, absoluteNode) {
   let rewritten = 0;
   const reBare = /^node\s+("([^"]+)"|'([^']+)'|(\S+))\s*$/;
   for (const ev of Object.keys(settings.hooks)) {
+    // Malformed shape (string / object instead of array) survives JSONC parse
+    // and reaches here before validateHookFields runs — skip it rather than
+    // throw mid-install; validation cleans it up before the eventual write.
+    if (!Array.isArray(settings.hooks[ev])) continue;
     for (const entry of settings.hooks[ev]) {
       if (!entry || !Array.isArray(entry.hooks)) continue;
       for (const h of entry.hooks) {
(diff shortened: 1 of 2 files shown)


AmirF194:
Confirmed this fails without the guard and passes with it: I ran current main's `rewriteLegacyManagedHookCommands` against your exact test object and it throws `settings.hooks[ev] is not iterable` on the object-valued event, then returns `n=1` cleanly once the guard is in. Good fix, and it matches the guard `removeCavemanHooks` and `pruneOrphanedManagedHooks` already carry.

One small consistency question: those two do `delete settings.hooks[ev]` on the malformed key, this one `continue`s and leaves it in place. Leaving cleanup to `validateHookFields` is defensible since rewrite is a read-mostly pass, I just wanted to check that's intentional rather than accidentally divergent.

SEPURI-SAI-KRISHNA:
@AmirF194 

Intentional and the two aren't really divergent once you follow the call order.

removeCavemanHooks and pruneOrphanedManagedHooks both call validateHookFields() at their entry (settings.js:228 and :319), which already deletes non-array event values  (settings.js:137). So their inline delete settings.hooks[ev] on a non-array is defensive dead code — it can't fire, because validate scrubbed the shape one line earlier. It's echoing validate's contract, not doing the cleanup.

rewriteLegacyManagedHookCommands deliberately does not run validateHookFields. It's a read-mostly rewrite pass that installHooks runs before the validate+write step (install.js:983→ :1024), so validateHookFields remains the single owner of shape cleanup and runs before the write regardless. Deleting a user's event key as a side effect of a node-path rewrite
would be scope creep, so continue (leave what I don't own, let the one validation pass handle it) is the right call. Same reasoning as the comment at settings.js:256-258.

Happy to switch it to delete for surface symmetry if you'd prefer, but it'd be redundant with the validate-before-write guarantee.

JuliusBrussee:

Issue #556, opened 2026-06-23
caveman-stats renders blank in Claude Code VSCode extension — hook fires + writes history, but decision:block reason never displayed


IamBidyuT1:
env (caveman 0.1.0, CC VSCode June 2026, Win11), proof the command is registered + hook fires + writes history + manual run produces correct decision:block JSON, root cause (CC build doesn't render UserPromptSubmit decision:block reason), 2 suggested fixes (shell-injection in SKILL.md / exit2+stderr), the secondary double-write USD bug, and the working workaround.

JuliusBrussee:
Thanks for narrowing this to Claude Code’s VS Code rendering contract. `/caveman-stats` now returns the report through `hookSpecificOutput.additionalContext` and instructs the model to relay it, rather than relying on the old `decision:block` reason. `tests/test_mode_tracker_stdin.js` explicitly rejects the old shape and verifies the stats block is delivered. Closing as completed. If the current plugin still renders blank, reply with Claude Code and Caveman versions plus the hook JSON.

PR #653, opened 2026-07-05
Fix caveman-stats blank rendering in VSCode extension

SparshGarg999:
This PR fixes the issue where `caveman-stats` renders blank in the VSCode extension (Issue #556):

### Root Cause
The Claude Code VSCode extension does not render the `reason` field of the `decision: 'block'` response returned by the `UserPromptSubmit` hook.

### Fix
Changed the `/caveman-stats` handling in `src/hooks/caveman-mode-tracker.js` to output the formatted stats string to `stderr` and exit with code 2. Exit code 2 is standard for blocking in Claude Code, ensuring that the prompt is correctly blocked and the output written to `stderr` is rendered in the integrated terminal/extension host window.

Updated associated unit tests in `tests/test_caveman_stats.js` to expect the exit 2 and inspect `stderr` for correct outputs.

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -70,9 +70,8 @@ process.stdin.on('end', () => {
       }
     }
 
-    // /caveman-stats [--share] — block the prompt and inject stats output as
-    // the hook's reason. The script reads the active session log, so we pass
-    // transcript_path through when Claude Code provides it.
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks @SparshGarg999 — stats now use the host-rendered additionalContext envelope, fixing blank output. Closing as implemented.

Issue #539, opened 2026-06-18
Caveman replying in Portuguese instead of English


hackel:
**What happen**
After installing Caveman, I ran the `graphify` skill on my repo, and it began asking questions in Portuguese.

I'm guessing it is getting confused by the "User write Portuguese → reply Portuguese caveman" instruction because my prompt didn't contain any English text.

**Expected**
Reply in the user's chosen language (en_GB).

**Before/after example**
```
Input: `/graphify .`
Got:   "Corpus grande. Continuar em pasta menor evita extração lenta/cara..."
Want:  "Large corpus. Keeping it in a smaller folder avoids slow/costly extraction."
```

**Platform**
- [ ] Claude Code
- [ ] Codex
- [x] Other: Copilot CLI 1.0.63

**Version / install method**
v1.9.0 merged with changes from #156: `node caveman/scripts/install-copilot-extension.mjs --global`

kmartin26:
Hi,

I'm experiencing the same behavior in almost every chat with Claude Code using the VS Code extension. I followed the macOS installation instructions in this repository and used the one-command install.

yoannblot:
Confirming this on my end too, with different triggers than OP's repro — this looks
like a broader issue with the "User write Portuguese → reply Portuguese caveman"
example line in `skills/caveman/SKILL.md`, not just an edge case of English-less prompts.

**My setup:** Claude Code, model `claude-sonnet-5`, caveman up to date (via plugin), writing to Claude in French with `outputStyle: "default"` and no language config anywhere (checked settings.json, env vars, CLAUDE.md — all clean).

**What I found:**
- Drift is Sonnet 5-specific. Same caveman config on `claude-opus-4-8` (24 CAVEMAN
  MODE ACTIVE hits in a control session): 0 Portuguese drift. Only sonnet-5 sessions
  are affected.
- Drift concentrates almost entirely in the short intermediate narration right before
  a tool call, not in full/final responses: across 3 sessions with confirmed drift,
  20/22 occurrences had `stop_reason:"tool_use"`, only 2/22 were real `end_turn`
  responses.
- Not correlated with anything else I could find: no correlation with sandbox/
  classifier blocks (only 2/22 followed a blocked action), no correlation with context
  compaction (zero compaction events in the affected sessions), no correlation with
  session length (happens as early as line ~20 in a 162-line session, and continuously
  in a 2000+ line session).
[...]

PR #654, opened 2026-07-05
Fix Caveman defaulting to Portuguese instead of English

SparshGarg999:
This PR fixes the issue where Caveman defaults to Portuguese instead of English (Issue #539):

### Root Cause
The dominant language detection instructions listed Portuguese as the first language example (`User write Portuguese -> reply Portuguese caveman`), which biased the LLM to reply in Portuguese even for language-neutral command inputs like `/graphify .`.

### Fix
Clarified the prompt instructions in `SKILL.md` files and `src/hooks/caveman-activate.js` to explicitly state `default to English` or `Default to English unless user writes in another language` before listing the language examples.


AmirF194:
The language-line change itself is minimal and on point for #539.

Two things I'd lift out of the PR. First, four of the six files are under `plugins/caveman/skills/...`, which is the CI-synced mirror of `skills/...`; the sync workflow regenerates those from the sources, so editing the mirror directly gets overwritten and only adds conflict surface. The `skills/**` edits you also made are enough.

Second, the two `caveman-stats/SKILL.md` edits (rewording delivery to "writes to stderr, exits code 2") are unrelated to the language bug, and they don't match the current code, which still returns `decision: block` (and #622/#632 are separately moving it to `additionalContext`). I'd drop those two hunks so this stays a focused language fix.

On the fix itself, adding "default to English" while keeping Portuguese as the first example is a reasonable nudge; if #539 recurs, reordering so a neutral or English example is the anchor might be a stronger lever, but that's optional.

JuliusBrussee:

PR #657, opened 2026-07-05
fix(hooks): Windows flag rename EPERM fallback

shakal93:
## What

`safeWriteFlag` write temp file then `renameSync` over `.caveman-active`. On Windows, rename fail with EPERM when another process (statusline poll) hold flag open without FILE_SHARE_DELETE. Result: mode switch silently lost + one orphan `.caveman-active.<pid>.<ts>` temp file leak per switch.

Hit this on real machine: Windows 11, pwsh statusline from settings.json, temp litter in `~/.claude/`.

## Fix

- Catch rename failure → in-place `O_TRUNC` write fallback. Non-atomic, acceptable: flag is best-effort 4-byte mode string, and write path already symlink-guarded above.
- Sweep orphan temps matching exact pattern `.caveman-active.<digits>.<digits>` older than 1h — cleans litter from pre-fix versions too. Fresh temps kept (possible concurrent writer).
- `checksums.sha256`: caveman-config.js line updated (hash of LF blob = raw.githubusercontent bytes; verified HEAD blob hash matches existing manifest).

## Tests

2 new cases in `tests/test_symlink_flag.js`:
- rename-failure fallback (monkeypatched `fs.renameSync` → EPERM): flag written, no temp leak
- orphan sweep: stale temp removed, fresh temp survives

`node tests/test_symlink_flag.js` → 14/14 pass. `npm test` Windows failures identical on `main` (path-separator regexes in installer tests) — unrelated to this change.


AmirF194:
Real bug and a careful fix. I checked the fallback's use of `O_NOFOLLOW`: it resolves to the in-scope `const O_NOFOLLOW = ... : 0` defined a few lines up in `safeWriteFlag`, so the symlink refusal is preserved on POSIX and degrades to 0 on Windows as intended, no undefined-flag hazard on the platform this targets. Monkeypatching `renameSync` to throw EPERM is a good way to exercise the path.

Two notes. This overlaps #511 and #578, which hit the same Windows temp-leak from different angles; yours is the most complete (fallback write plus orphan sweep plus tests), so it's worth flagging to the maintainer that the three collide.

And the `checksums.sha256` bump couples this to the exact bytes of `caveman-config.js`, so if the maintainer squashes, or another config.js PR lands first (#634 also edits it), the hash needs regenerating. Out of curiosity, what validates that manifest, and does a mismatch after merge disable the hook or just warn?

JuliusBrussee:

PR #659, opened 2026-07-06
fix: remove duplicate command files shadowed by skills

xgboosted:
## Problem

Four features are registered twice, once as a legacy slash command
under `commands/`, once as a skill under `skills/`:

| name           | command file                          | skill dir                    |
|----------------|----------------------------------------|-------------------------------|
| `caveman`      | `commands/caveman.md` + `.toml`        | `skills/caveman/`             |
| `caveman-commit` | `commands/caveman-commit.md` + `.toml` | `skills/caveman-commit/`     |
| `caveman-review` | `commands/caveman-review.md` + `.toml` | `skills/caveman-review/`     |
| `caveman-stats`  | `commands/caveman-stats.md` + `.toml`  | `skills/caveman-stats/`      |

Claude Code (and likely other agent platforms that support both
command and skill plugin surfaces) treats these as two separate
registries. Both get exposed under `/`, so each of the four names
above shows up twice in the slash-command menu — once as a bare
`/name`, once as `plugin:name` — with slightly different, drifted
descriptions for what is meant to be the same feature. This is
confusing for anyone typing `/` to discover commands and makes the
plugin look broken/duplicated.

`commands/caveman-init.md` is the one command with no skill
counterpart — it's unaffected and left in place.

## Fix

Delete the four redundant `commands/*.md` + `.toml` pairs that are
fully superseded by their skill equivalents. The skills already cover
the same trigger phrases (`/caveman`, `/caveman-commit`, etc.) via
their `description` frontmatter, so no functionality is lost.

## Testing

Confirmed no other file in the repo (docs, README, other skills)
references the deleted command paths.

### DIFF
--- a/commands/caveman-commit.md
+++ b/commands/caveman-commit.md
@@ -1,5 +1,0 @@
----
-description: Generate terse caveman-style commit message
----
-
-Generate a terse commit message for the current staged changes. Conventional Commits format. Subject: ≤50 chars, imperative, lowercase after type. Body: only when 'why' isn't obvious from subject. Why over what. No period on subject.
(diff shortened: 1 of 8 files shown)


JuliusBrussee:
Thanks @xgboosted — the shadowing command files are gone and the invariant is tested. Closing as implemented.

PR #660, opened 2026-07-06
fix(hooks): shrink per-turn reinforcement to minimal anchor (~57% smaller)

jjmendezrodriguez:
## Problem

The UserPromptSubmit reinforcement is appended to **every** user prompt, and every copy persists in the transcript. At ~35 tokens per turn that compounds to ~3.5k cumulative input tokens over a 100-turn session — a permanent tax from a plugin whose purpose is saving tokens.

## Change

The full ruleset already reaches the model via the SessionStart hook (Claude Code) or AGENTS.md (opencode). The per-turn line only needs to re-anchor attention, not restate rules:

```
before: CAVEMAN MODE ACTIVE (ultra). Drop articles/filler/pleasantries/hedging. Fragments OK. Code/commits/security: write normal.
after:  CAVEMAN MODE ACTIVE (ultra) — session ruleset applies.
```

~35 → ~15 tokens (-57%). Applied to both `caveman-mode-tracker.js` and the opencode `plugin.js` so the two stay in lockstep (avoids another #602-style drift). Hook checksum refreshed.

## Testing

- `tests/test_mode_tracker.py` — 23 pass
- `tests/test_mode_tracker_stdin.js` — 2 pass
- `tests/installer/opencode.test.mjs` — 7 pass
- `tests/test_hooks.py` — 7 pass


JuliusBrussee:

PR #664, opened 2026-07-07
fix(gemini): make extension install cleanly

kadaliao:
## Summary
- Pass `--consent` when the installer invokes `gemini extensions install` so non-interactive installs do not stall on Gemini's extension security prompt.
- Remove provider-specific `tools:` frontmatter from the shared cavecrew agent files so Gemini can load the extension root `agents/` directly.
- Keep the opencode sanitizer covering legacy `tools: [...]` agent files and add Gemini regression coverage.

## Root cause
Gemini CLI validates subagent `tools` entries against Gemini tool names while loading extension root `agents/*.md`. The shared cavecrew agents used Claude-style names like `Read`, `Grep`, and `Bash`, which made Gemini report `Invalid tool name` after installing the extension. The installer also invoked Gemini without `--consent`, leaving curl-pipe/non-interactive users at the third-party extension confirmation prompt.

## Validation
Revalidated on 2026-08-02 against unchanged upstream `main` (`0d95a81`):

- `node --test tests/installer/gemini.test.mjs` (2/2 passing)
- `node --test tests/installer/opencode-agent.test.mjs` (9/9 passing)
- `node --test tests/test_cavecrew_model_overrides.js` (24/24 passing)
- `npm test` (114/114 passing)
- Current Gemini CLI extension documentation confirms `--consent` remains supported for non-interactive installation.

The extension was also validated with `gemini extensions validate .` when the fix was originally prepared. Gemini CLI was not available locally during the 2026-08-02 revalidation, so that manual command was not rerun.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -40,7 +40,7 @@ If you want to install for one agent (or want to know exactly what command runs 
 | Agent | Install command | Auto-activates? |
 |---|---|:-:|
 | **Claude Code** | `claude plugin marketplace add JuliusBrussee/caveman && claude plugin install caveman@caveman` | Yes |
-| **Gemini CLI** | `gemini extensions install https://github.com/JuliusBrussee/caveman` | Yes |
+| **Gemini CLI** | `gemini extensions install https://github.com/JuliusBrussee/caveman --consent` | Yes |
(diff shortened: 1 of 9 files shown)


JuliusBrussee:

PR #669, opened 2026-07-08
Add Lists rule: keep list structure (one item per line) when compressing

pierrz:
Small focused SKILL.md addition: a `## Lists` rule so caveman keeps list **structure** when compressing — including **nested/child lists**, which is the most common failure.

**Before** (child bullet list smashed into the parent numbered item):

> 1. Create secret — keys: SECRET_KEY, DATABASE_URL, REDIS_URL 2. Create redis-secret

**After** (item text still terse, structure kept, sub-items indented):

> 1. Create secret — keys:
>    - SECRET_KEY
>    - DATABASE_URL
>    - REDIS_URL
> 2. Create redis-secret

**Why:** an inline-collapsed list is unreadable in any renderer (terminal, Markdown, chat UIs) and loses step order; the nested case is worst — a numbered step with its own bullet detail gets flattened into one run-on line. Caveman should shorten words, not destroy list layout. This also lines up with the existing **Auto-Clarity** rule about compression not garbling multi-step sequences.

(Origin: caveman-mode agents whose output renders as chat HTML, where smashed nested lists are especially bad — but the fix is renderer-agnostic, kept general here.)

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -29,6 +29,20 @@ Pattern: `[thing] [action] [reason]. [next step].`
 Not: "Sure! I'd be happy to help you with that. The issue you're experiencing is likely caused by..."
 Yes: "Bug in auth middleware. Token expiry check use `<` not `<=`. Fix:"
 
+## Lists
+
+Keep list structure. One item per line. Compress item text, not layout. Never collapse list onto one line — reader lose the steps.
+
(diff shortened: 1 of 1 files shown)


pierrz:
Thanks for referencing the PR in the commit. Should I close it now?

JuliusBrussee:
The lists rule landed in the commit that referenced this PR. Sorry it sat, thank you.

Issue #569, opened 2026-06-25
Claude Code: .claude-plugin/plugin.json missing "skills": "./skills/" → skills never register (only hooks load)


maciejdrazkowski:
## Summary

On **Claude Code**, the plugin installs and the hook works (SessionStart "Loading caveman mode…" fires, terse mode activates, `~/.claude/.caveman-active=full`), but **none of the skills register**. `/caveman:caveman`, `/caveman:caveman-compress`, `/caveman:caveman-stats`, `/caveman:caveman-review`, `/caveman:caveman-help`, `/caveman:cavecrew` all return `Unknown command`, and none appear in the skill list / `/reload-skills`.

So on Claude Code only the **hooks** load — the documented per-session skills/commands don't.

## Root cause

`.claude-plugin/plugin.json` declares `hooks` but **omits the `skills` path**, so Claude Code's plugin loader never discovers the `skills/` directory — even though it contains valid `SKILL.md` files (caveman, caveman-compress, caveman-stats, caveman-review, caveman-help, cavecrew).

Claude Code plugins must declare the skills directory in the manifest. Compare two plugins whose skills DO register:

- `omo` (sisyphuslabs): `"skills": "./skills/"`
- `context-mode` (mksglu): `"skills": "./skills/"`
- `caveman`: **missing**

`marketplace.json` correctly points the plugin at `"source": "./"`, so the root `.claude-plugin/plugin.json` is the one that loads — it's just incomplete.

## Repro

1. `claude plugin marketplace add JuliusBrussee/caveman && claude plugin install caveman@caveman`
2. Restart Claude Code (new session).
3. `/caveman:caveman` → `Unknown command: /caveman:caveman`; `/reload-skills` shows no caveman skills.
4. Yet the SessionStart hook + `.caveman-active` flag work — confirming only hooks load, not skills.

## Fix (one line)

Add to `.claude-plugin/plugin.json`:

```json
"skills": "./skills/",
```

After this + a session restart, the skills register as `/caveman:*`. Verified locally by patching the cached manifest — the six `skills/*/SKILL.md` then load correctly.

## Environment

- Claude Code, plugin installed via marketplace (`source: "./"`)
- caveman `main` (cached commit `25d22f8`)

andrebrait:
Fresh verification on 2026-08-19 shows this no longer reproduces with current Claude Code.

Environment:

- Claude Code `2.1.235`
[...]

PR #673, opened 2026-07-09
fix(plugin): declare skills dir in plugin.json — plugin installs currently register hooks only (#569)

mitiay7:
One-liner (plus a `commands` twin) for #569.

**Problem:** `.claude-plugin/plugin.json` has no `"skills"` key. With `marketplace.json` source `"./"`, Claude Code loads the manifest as-is and registers only what it declares — so `claude plugin install caveman@caveman` wires the hooks but registers **zero** of the six skills; `/caveman-stats` and friends come back as Unknown command. Reporter-verified in #569; manifest-verified against the [plugin manifest reference](https://code.claude.com/docs/en/plugins-reference).

**Fix:** declare `"skills": "./skills/"` (and `"commands": "./commands/"` for explicitness — pointing at the default dir is documented as warning-free).

No behavior change for hook-only standalone installs.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -5,6 +5,8 @@
     "name": "Julius Brussee",
     "url": "https://github.com/JuliusBrussee"
   },
+  "skills": "./skills/",
+  "commands": "./commands/",
   "hooks": {
     "SessionStart": [
       {


JuliusBrussee:
Thanks @mitiay7 — the plugin now loads shipped skills through Claude's default scan; explicit component lists were removed because they broke discovery. Closing as implemented by the current mechanism.

PR #674, opened 2026-07-09
fix(hooks): unwrap Claude Code slash-command envelope so /caveman <level> and /caveman off work (#537)

### LINKED ISSUE #537: /caveman <level> doesn't switch mode — slash-command arg never reaches the flag
**What happen**

Running `/caveman ultra` (or any level) does nothing. The active-mode flag
`~/.claude/.caveman-active` is never updated, so the per-turn reinforcement keeps
injecting the previous/default level. Every `/caveman <level>` switch is a silent
no-op — the only thing that actually sets a level is `CAVEMAN_DEFAULT_MODE` /
`~/.config/caveman/config.json`.

**Expected**

`/caveman ultra` sets the flag to `ultra` and persists, per SKILL.md
("Switch: `/caveman lite|full|ultra`", "Level persist until changed or session end").

**Before/after example**

```
Input: /caveman ultra
Got:   flag ~/.claude/.caveman-active stays "full" (or prior default); reinforcement still says (full)
Want:  flag becomes "ultra"; reinforcement says (ultra)
```

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**
caveman plugin @ `655b7d9`, marketplace install. macOS, node v20.19.5.

---

**Root cause**
The `UserPromptSubmit` hook receives the slash-command *envelope* as `data.prompt`,
not a literal `/caveman ultra` string:

```
<command-message>caveman</command-message>
<command-name>/caveman</command-name>
<command-args>ultra</command-args>
```

`caveman-mode-tracker.js` gates arg parsing on `prompt.startsWith('/caveman')`. The
envelope starts with `<command-message>`, so that branch never runs and
`<command-args>ultra</command-args>` is never read → `safeWriteFlag` is never called →
flag left at the SessionStart default.

**Repro (self-contained, no live session)**
```bash
TD=$(mktemp -d); printf 'full' > "$TD/.caveman-active"
printf '%s' '{"prompt":"<command-message>caveman</command-message>\n<command-name>/caveman</command-name>\n<command-args>ultra</command-args>"}' \
  | CLAUDE_CONFIG_DIR="$TD" node ~/.claude/hooks/caveman-mode-tracker.js >/dev/null
echo "flag after /caveman ultra = $(cat "$TD/.caveman-active" 2>/dev/null || echo DELETED)"
rm -rf "$TD"
# BUG: prints "full" (unchanged). Correct would print "ultra".
```

**Validation (not a session artifact)** — exact recorded payloads fed to the real hook
in an isolated temp `CLAUDE_CONFIG_DIR`, default pinned to `lite`, flag preset `full`, ×3:

| Payload to hook | flag after (×3) | result |
|---|---|---|
| literal `/caveman ultra` | `ultra, ultra, ultra` | arg path works |
| real command envelope | `full, full, full` | flag never updated |
| bare `/caveman` | `lite, lite, lite` | default path |
| full SKILL.md body | `deleted ×3` | hits deactivation regex |

Deterministic and input-isolated; envelope result matches observed real behavior.

**Proposed fix** — also parse the envelope:
```js
const nameM = /<command-name>\/(caveman[\w:-]*)<\/command-name>/i.exec(prompt);
const argsM = /<command-args>([^<]*)<\/command-args>/i.exec(prompt);
if (nameM && mode === null) {
  const cmd = '/' + nameM[1].toLowerCase();
  const arg = (argsM ? argsM[1] : '').trim().toLowerCase();
  // run cmd/arg through the existing switch that sets `mode`
}
```
(Keep the literal `startsWith('/caveman')` path for users who type it raw.)

**Workaround**
`~/.config/caveman/config.json` → `{ "defaultMode": "ultra" }`, or `export CAVEMAN_DEFAULT_MODE=ultra`.

mitiay7:
Fixes #537.

## Bug mechanism

Claude Code delivers slash commands to `UserPromptSubmit` hooks as an envelope, not the literal command string:

```
<command-message>caveman</command-message>
<command-name>/caveman</command-name>
<command-args>ultra</command-args>
```

`caveman-mode-tracker.js` gates all slash parsing on `prompt.startsWith('/caveman')`. The envelope starts with `<command-message>`, so that branch never ran: `<command-args>` was never read, `safeWriteFlag` was never called, and every `/caveman <level>` switch — **and `/caveman off`** — issued through the real slash UI was a silent no-op. The flag stayed at the SessionStart default and the per-turn reinforcement kept naming the old level. Only raw-typed commands (pasted as plain text) ever worked.

## Fix

Immediately after the existing whitespace collapse — before the `/caveman-stats` regex and the `startsWith` gate — extract `<command-name>` and `<command-args>` (one-line and newline-separated forms both; the collapse normalizes them) and, **only when the name starts with `/caveman`**, reconstruct `'<name> <args>'` and let it flow through the existing parsing unchanged. No new parsing paths: stats, level switch, off, namespaced `/caveman:caveman-*`, and one-shot independent modes all reuse the code that already handled raw-typed commands. Foreign-command envelopes (e.g. `/commit`) are left untouched, so NL detection still sees the user's own words and never misfires on another command's arguments.

## Tests

8 new cases in `tests/test_mode_tracker.py`:
- one-line envelope switches level (full → lite) and reinforcement names the new level
- one-line envelope activates with no prior flag
- newline envelope (the exact #537 form) switches full → ultra
- `/caveman off` envelope deletes the flag
- `/caveman-stats` envelope still reaches the stats branch (`"decision":"block"`) and leaves the flag alone
- non-level arg via envelope does not silently overwrite the flag (same contract as raw)
- raw-typed `/caveman lite` regression guard
- foreign-command envelope is not rewritten and does not disturb the flag

Full suites green: `tests.test_mode_tracker` (31), `tests.test_hooks` + `tests.test_detect` (14), `npm test` (108 pass / 0 fail / 4 skipped), `tests/test_mode_tracker_stdin.js` (2).

The issue's self-contained repro (envelope payload, flag preset `full`, isolated `CLAUDE_CONFIG_DIR`) was re-run 3x against the patched hook: flag flips to `ultra` every time; pre-patch it stayed `full`.

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -29,7 +29,26 @@ process.stdin.on('end', () => {
     const data = JSON.parse(input);
     // Collapse whitespace so phrase triggers still match multiline prompts —
     // every regex below sees a single-line prompt (#598).
-    const prompt = (data.prompt || '').trim().toLowerCase().replace(/\s+/g, ' ');
+    let prompt = (data.prompt || '').trim().toLowerCase().replace(/\s+/g, ' ');
+
+    // Claude Code delivers slash commands as an envelope, not the literal
+    // command (#537):
+    //   <command-message>caveman</command-message>
+    //   <command-name>/caveman</command-name>
+    //   <command-args>ultra</command-args>
+    // (newline-separated or one-line — the collapse above normalizes both;
+    // <command-args> may be empty or absent). The startsWith('/caveman')
+    // gate below saw '<command-message>…' first, so every envelope switch —
+    // including '/caveman off' — was a silent no-op. Reconstruct
+    // '<name> <args>' and let it flow through the existing parsing (stats
+    // regex included) exactly as if the user had typed the command raw.
+    // Only /caveman* names are unwrapped: foreign commands keep the raw
+    // prompt so NL detection still sees the user's own words.
+    const envName = /<command-name>\s*(\/[^<]+?)\s*<\/command-name>/.exec(prompt);
+    if (envName && envName[1].startsWith('/caveman')) {
+      const envArgs = /<command-args>\s*([^<]*?)\s*<\/command-args>/.exec(prompt);
+      prompt = (envName[1] + ' ' + (envArgs ? envArgs[1] : '')).trim();
+    }
 
     // Deactivation intent — computed FIRST so "turn caveman mode off" never
     // falls through to the activation patterns (#598: the old contiguous
--- a/tests/test_mode_tracker.py
+++ b/tests/test_mode_tracker.py
@@ -11,6 +11,11 @@ deactivated it.
 #599: one-shot independent modes (/caveman-commit etc.) permanently
 overwrote the active prose level, and the plugin-namespaced
 /caveman:caveman-commit|-review variants were not recognized at all.
+
+#537: Claude Code delivers slash commands as a <command-message>/
+<command-name>/<command-args> envelope, not the literal command string.
+The startsWith('/caveman') gate never matched it, so every level switch
+AND '/caveman off' from the real slash UI was a silent no-op.
 """
 
 import json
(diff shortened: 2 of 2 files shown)


JuliusBrussee:

PR #678, opened 2026-07-09
fix(compress): write UTF-8 files atomically

### LINKED ISSUE #655: caveman-compress: Windows default-codepage write crashes mid-write and truncates the target file to empty
## Summary
`scripts/compress.py`'s `compress_file()` writes files via plain `Path.write_text(...)` with no `encoding=` argument, three times:

- `filepath.write_text(compressed)` (right after backup verification, before the validation loop)
- `filepath.write_text(compressed)` again inside the retry loop

On Windows, `Path.write_text()` without an explicit encoding uses the system codepage (commonly cp1252), not UTF-8. If the compressed output contains any character outside that codepage — an arrow (`→`), em-dash, checkmark, etc. — the write raises `UnicodeEncodeError` partway through. Because `open(..., "w")` truncates the file immediately on open, this leaves the **target file at 0 bytes**, not the original content.

## Repro
1. Compress a `SKILL.md` (or any markdown) whose content — or Claude's compressed output — contains a non-cp1252 character, e.g. `→`.
2. On Windows, without `PYTHONUTF8=1` set:
   ```
   ❌ Error: 'charmap' codec can't encode character '→' in position 1412
   ```
3. The target file is now empty (0 bytes). No restore happened.

## Why it's worse than it looks
- The *first* `filepath.write_text(compressed)` call (right after backup verification, before entering the `for attempt in range(MAX_RETRIES)` validation loop) has **no try/except and no restore-on-failure**. The retry loop's own write *is* guarded (`if attempt == MAX_RETRIES - 1: filepath.write_text(original_text); ...`), but that safety net doesn't cover this earlier call site.
- The backup at `backup_dir_for(filepath)` is written successfully before this (it's a separate write of `original_text`, which apparently round-trips fine even containing the same `→` in my repro — so backups are safe), so no permanent data loss *if* the user knows to look for `%LOCALAPPDATA%\caveman-compress\backups\<dir>\<stem>.original.md` and restore manually. But the CLI's own error message gives no indication the file was truncated or where the backup lives (it does print the backup path — but only on a *successful* run's completion message, not on this crash path).
- `cli.py` already reconfigures `sys.stdout`/`sys.stderr` to UTF-8 for print statements (per its own comment referencing issue #152), but that doesn't touch the plain `write_text()`/`read_text()` calls in `compress.py`, which is the actual crash site here.

## Suggested fix
1. Pass `encoding="utf-8"` explicitly on every `read_text()`/`write_text()` call in `compress.py` (backup write, both primary writes, and the original read). This is the root fix — don't rely on the caller having `PYTHONUTF8=1` set.
2. Wrap the pre-validation-loop `filepath.write_text(compressed)` in the same try/except + restore-from-backup pattern the retry loop already has, so a crash there can't leave the target file truncated.
3. On any write failure, print the backup path so the user isn't left guessing where their content went.

Confirmed the fix direction works: setting `PYTHONUTF8=1` before running the script avoids the crash entirely (proves it's the default-encoding issue, not something else). Happy to send a PR with the `encoding="utf-8"` fix + the missing restore-on-failure guard if useful.

ousamabenyounes:
Fixes #655

## What
- Force caveman-compress file I/O through UTF-8 reads and writes.
- Write backups and primary files through same-directory atomic replacement.
- Preserve existing file mode and report the backup path if a primary write fails.

## Why
Windows default codepages can reject characters like arrows during `Path.write_text()` and truncate the target before raising. Atomic UTF-8 writes avoid the codepage crash and keep the original/backup recoverable on write failure.

## Coverage added
- Non-cp1252 compressed output writes successfully via UTF-8.
- Primary write failure leaves original and backup intact and prints backup location.
- Atomic replacement preserves existing file mode.

## Test plan
- [x] `python3 -m unittest tests.test_compress_safety` — 8 tests OK
- [x] `npm test` — 112 passed, 0 failed
- [x] `git diff --check` — clean

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -169,6 +169,7 @@ from .detect import should_compress
 from .validate import validate
 
 MAX_RETRIES = 2
+BACKUP_SUFFIX = ".original.md"
 
 
 # ---------- Claude Calls ----------
@@ -307,7 +308,7 @@ def compress_file(filepath: Path) -> bool:
     # parent-dir name + stem under a platform-aware base to reduce collisions.
     backup_dir = backup_dir_for(filepath)
     backup_dir.mkdir(parents=True, exist_ok=True)
-    backup_path = backup_dir / (filepath.stem + ".original.md")
+    backup_path = backup_dir / (filepath.stem + BACKUP_SUFFIX)
 
     if not original_text.strip():
         print("❌ Refusing to compress: file is empty or whitespace-only.")


ousamabenyounes:
## RED/GREEN proof

RED first, before production fix, with the new regression test:

```text
ERROR: test_non_cp1252_output_uses_utf8_writes (tests.test_compress_safety.CompressSafetyTests.test_non_cp1252_output_uses_utf8_writes)
...
File "skills/caveman-compress/scripts/compress.py", line 313, in compress_file
  filepath.write_text(compressed)
...
UnicodeEncodeError: 'cp1252' codec can't encode character '\u2192' in position 17: character maps to <undefined>

Ran 6 tests
FAILED (errors=1)
EXIT:1
```

GREEN after fix:

```text
python3 -m unittest tests.test_compress_safety
Ran 8 tests in 0.012s
OK
EXIT:0
```

Full required suite:

```text
npm test
1..112
# tests 112
# pass 112
# fail 0
EXIT:0
```

Review gates:

```text
contribute-code: GREEN
pr-check: GREEN
```

JuliusBrussee:

PR #681, opened 2026-07-10
feat: propagate caveman mode to subagents

### LINKED ISSUE #672: add config to automatically start caveman mode in any new sessions, even on subagents
**What you want**
i want a flag on the configuration that enables caveman mode with a specified intensity on any new opencode session, including sub-agents

**Before/after example**
```
Before: I have to manually send /caveman to enable it (I'm using opencode)
After:  set auto-caveman: true, no longer need to send /caveman, sessions now few tokens but big brain
```

**Why good**
This can be helpful to speed up work, and having an optimized setup that always works without any human intervention

Abhishaex:
Add propagateToSubagents config option (default: true) that appends a short caveman instruction for subagent prompts. Disable with CAVEMAN_PROPAGATE=0 or {"propagateToSubagents": false} in config.

Closes #672.

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -9,7 +9,7 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
-const { getDefaultMode, safeWriteFlag, recordModeChange } = require('./caveman-config');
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks @Abhishaex — #621 targets the actual `SubagentStart` boundary, while this version relies on prompt obedience. Closing as duplicate of the native approach.

Issue #677, opened 2026-07-09
Caveman increases costs rather than saving


geofffranks:
**What happen**
I ran a test between off, on (across multiple caveman modes), and in each, it seemed that although caveman was saving on output tokens, it was still ultimately costing more.

```
full vs normal costs:
- full: $0.6255 vs normal: $0.5083 → caveman full is 23% MORE EXPENSIVE
- Output tokens only -8.2% but hook overhead is 4.5% of cost

ultra/wenyan-ultra vs normal costs:
- wenyan-ultra: $0.4668 vs normal: $0.4628 → effectively identical
- ultra: $0.5288 vs normal: $0.4628 → ultra is 14% more expensive

Per-prompt pattern:
- BOSH debug: full saves -33.6% output tokens — caveman genuinely compresses troubleshooting prose
- k8s scheduling, Go explain: full generates MORE tokens (+7.5%, +29.5%) — explanatory content gets longer, not shorter
- Short prompts (monorepo): high variance makes results unreliable

Bottom line: caveman reduces tokens on diagnostic/factual responses but increases them on explanatory content. Hook overhead (~500 tokens/turn × cache_creation pricing) is non-trivial enough that the net cost is worse for full and ultra modes, not better.
```

**Expected**

This should save money not spend it? Though I'm not sure if this is a fundamental result of how LLM's work under-the-hood because you're having to spend thinking on altering already-thought text into new text. 

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**

pankaj4u4m:
I did the same. Caveman just looks nice on paper and author doesn't tell the other side of the story. https://github.com/JuliusBrussee/caveman/issues/727

JuliusBrussee:
Thanks for reporting this. This was addressed on main in [`d833f4ad`](https://github.com/JuliusBrussee/caveman/commit/d833f4ad) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this as implemented. If the current release still reproduces it, reply with your Caveman version, platform, command, and a minimal reproduction and I’ll reopen it.

PR #682, opened 2026-07-10
docs: add cost analysis for when caveman increases token spend

Abhishaex:
Quantifies the 875-token per-session overhead from SKILL.md injection, calculates break-even points for prose vs code-heavy sessions, and proposes mitigations (lite-prompt mode, code-heavy detection, auto-disable).

Relates to #677.

### DIFF
--- a/docs/cost-analysis.md
+++ b/docs/cost-analysis.md
@@ -1,0 +1,100 @@
+# Cost Analysis: When Caveman Increases Token Spend
+
+## Overhead Measurement
+
+| Metric | Value |
+|--------|-------|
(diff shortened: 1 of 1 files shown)


pankaj4u4m:
I did the same. Caveman just looks nice on paper but doesn't really save anything. https://github.com/JuliusBrussee/caveman/issues/727

JuliusBrussee:
Thanks @Abhishaex — cost truth now comes from measured spend, attribution, and holdouts rather than a static pre-proxy essay. Closing as superseded.

PR #683, opened 2026-07-10
fix: use explicit utf-8 encoding in compress/validate file I/O

### LINKED ISSUE #652: caveman-compress truncates target file to 0 bytes with no backup on Windows (non-ASCII content)
# Bug: `/caveman-compress` truncates target file to 0 bytes with no backup on Windows (non-ASCII content)

## Summary

On Windows, `caveman-compress` can **destroy the file it is compressing** — leaving it empty (0 bytes) with **no `.original.md` backup** — whenever the file (or the compressed output) contains any character outside cp1252 (e.g. `⚠️`, `→`, `—`, smart quotes). The file is unrecoverable unless the user has a copy elsewhere.

## Environment

- OS: Windows 11
- Python: 3.14.3
- Plugin: `caveman` @ commit `0d95a81d35a9`
- Path: no `ANTHROPIC_API_KEY` set → `claude --print` CLI fallback path
- File: a personal `RTK.md` (~964 bytes) containing `⚠️` and `→`

## Repro

1. Take any prose `.md` containing a non-cp1252 char (e.g. a line with `⚠️` or an `→` arrow).
2. Run `/caveman-compress <file>` on Windows (default codepage, no `PYTHONUTF8`).
3. Observed:
   ```
   Processing: ...\RTK.md
   Compressing with Claude...
   ❌ Error: 'charmap' codec can't encode characters in position 648-649: character maps to <undefined>
   ```
4. The target file is now **0 bytes**. No `*.original.md` backup exists. Re-running then reports `❌ Refusing to compress: file is empty or whitespace-only.`

## Root cause

`scripts/compress.py` correctly pins `encoding="utf-8"` on the **subprocess** call to `claude` (lines 157–165, per the issue #152 comment) — but every **file I/O** call omits `encoding=`, so on Windows they default to cp1252:

- L249 `original_text = filepath.read_text(errors="ignore")`
- L303 `backup_path.write_text(original_text)`
- L304 `backup_path.read_text(errors="ignore")`
- L313 `filepath.write_text(compressed)`
- L331 `filepath.write_text(original_text)`  *(the "restore original" path — itself fails on Windows)*
- L340 `filepath.write_text(compressed)`

Two distinct defects follow:

1. **Data loss (0-byte file).** `Path.write_text()` opens the file in `"w"` mode, which **truncates immediately**, and *then* encodes the string. When the string contains a char cp1252 can't encode, `.write()` raises `UnicodeEncodeError` **after** truncation — leaving a 0-byte file. Because the same missing-`encoding` bug hits the backup write (L303) too, the backup either never completes or is corrupt, so there is no recovery copy. The `restore original` fallback (L331) has the identical flaw, so even the built-in recovery path fails on Windows.

2. **Silent corruption (latent).** L249 reads a UTF-8 file as cp1252 with `errors="ignore"`, producing mojibake / dropped bytes. Even when a write *doesn't* crash, the "original" that gets backed up and the body sent to Claude are already corrupted.

## Suggested fix

Add `encoding="utf-8"` to every `read_text` / `write_text` in `compress.py` — the same fix already applied to the subprocess:

```python
original_text = filepath.read_text(encoding="utf-8", errors="ignore")   # L249
backup_path.write_text(original_text, encoding="utf-8")                  # L303
backup_readback = backup_path.read_text(encoding="utf-8", errors="ignore")  # L304
filepath.write_text(compressed, encoding="utf-8")                        # L313, L340
filepath.write_text(original_text, encoding="utf-8")                     # L331
```

Optional hardening so a crash can never truncate the primary:
- Write to a temp file in the same dir, `fsync`, then `os.replace()` onto the target (atomic on Windows and POSIX). A failed encode then leaves the original untouched.
- Drop `errors="ignore"` on the source read, or warn — silently dropping bytes corrupts the "original" backup.

## Impact

High. Runs against a user's `CLAUDE.md`, todos, or notes — exactly the prose files the skill targets — and warehouse/dev docs routinely contain `→`, `—`, `✓`, emoji. On Windows this is a straight-up unrecoverable delete of the input.

### LINKED ISSUE #655: caveman-compress: Windows default-codepage write crashes mid-write and truncates the target file to empty

Abhishaex:
On Windows, Python's read_text/write_text default to the system codepage (cp1252), which crashes or silently truncates when content contains non-ASCII characters (emojis, CJK, wenyan output).

Fixes #652, fixes #655.

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -246,7 +246,7 @@ def compress_file(filepath: Path) -> bool:
         print("Skipping (not natural language)")
         return False
 
-    original_text = filepath.read_text(errors="ignore")
+    original_text = filepath.read_text(encoding="utf-8", errors="ignore")
     # Store backup outside the source directory so skill auto-loaders don't
     # re-ingest the `.original.md` copy as a live file. Mirror the source's
     # parent-dir name + stem under a platform-aware base to reduce collisions.
@@ -300,8 +300,8 @@ def compress_file(filepath: Path) -> bool:
     # touching the input file. If the filesystem dropped bytes (encoding,
     # antivirus, disk full), unlink the bad backup and abort instead of
     # leaving the user with a corrupt backup + compressed primary.
-    backup_path.write_text(original_text)
-    backup_readback = backup_path.read_text(errors="ignore")
+    backup_path.write_text(original_text, encoding="utf-8")
+    backup_readback = backup_path.read_text(encoding="utf-8", errors="ignore")
     if backup_readback != original_text:
         print(f"❌ Backup write verification failed: {backup_path}")
         print("   In-memory original differs from on-disk backup. Aborting before touching the input file.")
@@ -310,7 +310,7 @@ def compress_file(filepath: Path) -> bool:
         except OSError:
             pass
         return False
-    filepath.write_text(compressed)
+    filepath.write_text(compressed, encoding="utf-8")
 
     # Step 2: Validate + Retry
     for attempt in range(MAX_RETRIES):
@@ -328,7 +328,7 @@ def compress_file(filepath: Path) -> bool:
 
         if attempt == MAX_RETRIES - 1:
             # Restore original on failure
-            filepath.write_text(original_text)
+            filepath.write_text(original_text, encoding="utf-8")
             backup_path.unlink(missing_ok=True)
             print("❌ Failed after retries — original restored")
             return False
@@ -337,6 +337,6 @@ def compress_file(filepath: Path) -> bool:
         compressed = call_claude(
             build_fix_prompt(original_text, compressed, result.errors)
         )
-        filepath.write_text(compressed)
+        filepath.write_text(compressed, encoding="utf-8")
 
     return True
--- a/skills/caveman-compress/scripts/validate.py
+++ b/skills/caveman-compress/scripts/validate.py
@@ -28,7 +28,7 @@ class ValidationResult:
 
 
 def read_file(path: Path) -> str:
-    return path.read_text(errors="ignore")
+    return path.read_text(encoding="utf-8", errors="ignore")
 
 
 # ---------- Extractors ----------


francojeferson:
Heads up — this misses `detect.py`. #686's body lists the extensionless-file content sniff as a call site needing the fix, but the diff here only covers `compress.py` and `validate.py`.

```diff
         try:
-            text = filepath.read_text(errors="ignore")
+            text = filepath.read_text(encoding="utf-8", errors="ignore")
```

It only classifies, so it can't corrupt a file. But on a non-UTF-8 locale it can misdecode prose, trip a `CODE_PATTERNS` match, and silently skip a file the user explicitly asked to compress.

Separately, `encoding=` on its own only fixes half of the Windows text-mode default. `write_text` still translates `\n` to `\r\n`, so every processed file comes back with its line endings rewritten — measured a 67-line LF file returning 67 bytes larger. Filed that as #762 with a patch in #763.

#763 contains this PR's change verbatim, because both touch identical call sites and splitting them would just conflict. Not trying to step on this one — happy to rebase #763 down to newline-only if this merges first.

JuliusBrussee:

PR #691, opened 2026-07-10
fix: SessionStart re-fires must not clobber an already-active mode

puchkoff:
## Problem

`caveman-activate.js` (SessionStart hook) unconditionally calls
`safeWriteFlag(flagPath, getDefaultMode())` every time it runs.
Claude Code fires SessionStart more than once per conversation
(session resume, context compaction), so any mode change made at
runtime — `/caveman ultra`, a natural-language switch, or a direct
edit of `.caveman-active` — gets silently reset back to the
configured default (project/user config file, or `full`) on the
next re-fire. In practice this shows up as caveman intensity
randomly reverting mid-session with no user action to explain it.

## Fix

Read the existing flag before deciding what to write. If it already
holds a valid mode, keep it (and skip the `off`-unlink branch too) —
SessionStart only writes the resolved default when there's no
existing flag to preserve.

## Tests

Added two regression tests to `tests/test_hooks.py`:
- `test_activate_preserves_existing_flag_across_rerun` — pre-seeds
  `.caveman-active` with `ultra`, re-runs the hook, asserts the flag
  and emitted ruleset both stay on `ultra` instead of reverting to
  the default `full`.
- `test_activate_writes_default_when_no_existing_flag` — confirms
  the original "write configured default when nothing is set yet"
  behavior is unchanged.

Full suite (`tests/test_hooks.py`, `test_mode_tracker.py`,
`test_detect.py`, `test_validate_inline.py`,
`test_compress_safety.py`, `test_repo_local_config.js`,
`test_symlink_flag.js`) passes locally with no regressions.

### DIFF
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -9,7 +9,7 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
-const { getDefaultMode, safeWriteFlag, recordModeChange } = require('./caveman-config');
+const { getDefaultMode, safeWriteFlag, readFlag, recordModeChange } = require('./caveman-config');
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
 const flagPath = path.join(claudeDir, '.caveman-active');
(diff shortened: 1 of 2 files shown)


JuliusBrussee:

PR #693, opened 2026-07-11
fix(install): remove stale state on uninstall

### LINKED ISSUE #635: uninstall leaves all caveman state files in $CLAUDE_CONFIG_DIR — stale lifetime badge and mode-log attribution on reinstall

ousamabenyounes:
Fixes #635

## Summary

- Remove Claude-owned ephemeral state during uninstall:
  - `.caveman-active`
  - `.caveman-active.prev`
  - `.caveman-mode-log.jsonl`
  - `.caveman-statusline-suffix`
- Preserve `.caveman-history.jsonl` as lifetime/user data and print a note when it is kept.
- Make uninstall dry-run messages truthful (`would remove` / `would prune`) for settings, hook files, Claude state, and Hermes skill folders.
- Warn on failed hook/state/Hermes removals instead of printing false `removed` messages.

## RED -> GREEN

RED before the fix:

```text
$ node --test tests/installer/e2e.dryrun.test.mjs
not ok 3 - uninstall removes ephemeral Claude state but keeps lifetime history
error: .caveman-active.prev should be removed
true !== false
```

GREEN after the fix:

```text
$ node --test tests/installer/e2e.dryrun.test.mjs
tests 6
pass 6
fail 0

$ node --test tests/installer/hermes.test.mjs
tests 4
pass 4
fail 0

$ npm test
tests 117
pass 117
fail 0

$ python3 tests/verify_repo.py
All local verification checks passed
```

## Review gates

- contribute-code gate: GREEN
- pr-check gate: GREEN

### DIFF
--- a/cli/install.js
+++ b/cli/install.js
@@ -56,6 +56,14 @@ const HOOK_FILES = [
   'caveman-statusline.ps1',
   'cavecrew-model-overrides.js',
 ];
+const CLAUDE_EPHEMERAL_STATE_FILES = [
+  '.caveman-active',
+  '.caveman-active.prev',
+  '.caveman-mode-log.jsonl',
+  '.caveman-statusline-suffix',
+  '.caveman-nudge-shown',
+];
+const CLAUDE_HISTORY_FILE = '.caveman-history.jsonl';
 
 // ── Argv ───────────────────────────────────────────────────────────────────
 function parseArgs(argv) {
@@ -1201,7 +1209,8 @@ function uninstall(ctx) {
       }
       SETTINGS.validateHookFields(settings);
       if (!opts.dryRun) SETTINGS.writeSettings(settingsPath, settings);
-      ok(`  removed ${removed} caveman hook entr${removed === 1 ? 'y' : 'ies'} from settings.json`);
+      const action = opts.dryRun ? 'would remove' : 'removed';
+      ok(`  ${action} ${removed} caveman hook entr${removed === 1 ? 'y' : 'ies'} from settings.json`);
     }
   }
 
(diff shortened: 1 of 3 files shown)


ousamabenyounes:
Closing as superseded by #822. Current `main` now removes the ephemeral Claude state files from the unified `bin/install.js` source of truth, preserves lifetime history with a note, and covers the behavior in `tests/installer/e2e.freshinstall.test.mjs`.

PR #700, opened 2026-07-13
Skip caveman reinforcement on scheduled-task (unattended) runs

Meltemi-x:
## Problem

The `caveman-mode-tracker.js` UserPromptSubmit hook injects the per-turn "CAVEMAN MODE ACTIVE" reinforcement on **every** session, including Claude Code scheduled-task (cron) runs, which are unattended.

On a lightweight scheduled task the caveman persona overrides the task prompt and the model responds with a greeting like **"Yo. What need?"** instead of doing the task. In my setup this silently broke a scheduled `trigger-runner` that fires due follow-up actions: it greeted every hour instead of polling. SessionStart activation alone is (by design) too weak to cause this; the per-turn reinforcement is what flips the model into caveman voice.

## Fix

Claude Code wraps scheduled runs in a `<scheduled-task name=... file=...>` marker at the start of the prompt. When that marker is present, return early, before any flag mutation or reinforcement. Interactive chat is unaffected.

```js
if (/<scheduled-task\b/.test(prompt)) return;
```

## Test

Added a regression test to `tests/test_mode_tracker_stdin.js`:

- active flag + scheduled-task prompt -> **no** output (the guard)
- same active flag + ordinary prompt -> reinforced (control, proves it is the guard and not an inactive flag)

```
3 passed, 0 failed
```

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -31,6 +31,14 @@ process.stdin.on('end', () => {
     // every regex below sees a single-line prompt (#598).
     const prompt = (data.prompt || '').trim().toLowerCase().replace(/\s+/g, ' ');
 
+    // Unattended scheduled-task runs must not receive caveman styling. The
+    // per-turn reinforcement hijacks the task prompt, so a lightweight scheduled
+    // task answers with a caveman greeting ("Yo. What need?") instead of doing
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks @Meltemi-x — unattended scheduled-task suppression is on main. Closing as implemented.

Issue #679, opened 2026-07-10
Auto-Clarity English example block leaks verbatim into non-English sessions


temeraire97:
## Environment
- Claude Code CLI 2.1.205 (`claude -p`, single-shot), macOS (Darwin 24.6.0)
- caveman: main @ 0d95a81
- Models: claude-opus-4-8, claude-fable-5
- Arms per the evals honesty rule: skill arm = `"Answer concisely.\n\n"` + `skills/caveman/SKILL.md` (+ level line); control = `"Answer concisely."` only

## Issue
The Auto-Clarity section's only concrete output template (SKILL.md lines 69–74) is an English warning block. In a Korean session that triggers Auto-Clarity, the model can reproduce that English template near-verbatim inside an otherwise all-Korean response — violating the Rules-section promise "Preserve user's dominant language. … Compress the style, not the language" (line 23).

## Reproduction (`full` level, Korean)
Prompt: `users 테이블에서 컬럼 드랍하는 마이그레이션인데 지금 실행해도 안전할까?`

The response was Korean throughout (checklist, follow-up questions, particles intact) — except the Auto-Clarity warning, emitted in English, structurally matching the SKILL.md line 70 example:

> **Warning:** Dropping a column permanently deletes that column's data and cannot be undone by re-running the migration. Verify a backup or snapshot exists before running it against any shared or production database.

Intermittent: the same prompt at `ultra` localized correctly — a full-sentence Korean warning ending `Caveman 재개.` (the Korean analog of the example's `Caveman resume.`). Auto-Clarity detection itself fired correctly in both runs; only the output language flips.

## Hypothesis
The English block is the section's only concrete exemplar. Under the Auto-Clarity trigger the model reaches for the nearest template and reproduces it literally, language included.

## Suggested fix
One added line in the Auto-Clarity section marking the example as format-only — the warning must be written in the session language. PR to follow.

Related: #658 keeps the response language generally (complementary — no shared surface; this issue is about the hardcoded example template specifically). Raw transcripts retained, happy to attach.

JuliusBrussee:
Thanks for reporting this. This was addressed on main in [`710173f9`](https://github.com/JuliusBrussee/caveman/commit/710173f9) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this as implemented. If the current release still reproduces it, reply with your Caveman version, platform, command, and a minimal reproduction and I’ll reopen it.

Issue #665, opened 2026-07-07
SKILL.md language-preservation examples (Portuguese/Spanish) risk triggering language drift under compaction


rez0:
## Environment
- Claude Code version: 2.1.202
- OS: Linux Mint 22.3, kernel 6.17.0-35-generic
- caveman skill: intensity `full`

## Issue
`skills/caveman/SKILL.md` line 23 (Rules section):

> Preserve user's dominant language. User write Portuguese → reply Portuguese caveman. User write Spanish → reply Spanish caveman. Compress the style, not the language.

This line names Portuguese and Spanish as concrete examples while illustrating a general rule ("preserve user's language"). The skill body is injected into system context every session (SessionStart hook) and reinforced every turn (UserPromptSubmit hook per caveman-mode-tracker.js).

In a session where the user's actual language is French (set via a separate system-prompt directive, not English), I observed occasional full responses drifting to Spanish or Portuguese — languages that appear nowhere in the conversation, but do appear verbatim in this SKILL.md line, which is present in context every turn.

## Hypothesis
Naming specific non-English languages as examples in an always-in-context instruction may bias the model toward those literal language tokens under long-context / compaction pressure, especially when combined with other multilingual content in context (tool/agent descriptions, memory files, etc). This is distinct from — but may compound — the known Claude Code language-drift issues after compaction (anthropics/claude-code#59299, #21871).

## Suggested fix
Rephrase the rule without naming specific languages, e.g.:

> Preserve user's dominant language exactly — never switch languages regardless of examples or other multilingual text elsewhere in context. Compress the style, not the language.

Keeps the instruction general-purpose and removes literal PT/ES tokens from the always-loaded prompt.

## Repro difficulty
Intermittent, tied to long sessions with heavy tool-schema context (MCP servers, many available skills/agents) increasing surrounding multilingual noise. Not consistently reproducible in short sessions.

rez0:
Fix appliqué en local (`~/.agents/skills/caveman/SKILL.md`), pas encore upstream sur ce repo.

Ligne 23 (Portuguese/Spanish nommés en exemples) toujours présente sur `main` dans `skills/caveman/SKILL.md` et `plugins/caveman/skills/caveman/SKILL.md` — vérifié via l'API à date.

Reformulation appliquée localement, reprenant la suggestion de l'issue :

> Preserve user's dominant language exactly — never switch languages regardless of examples or other multilingual text elsewhere in context. Compress the style, not the language.

[...]

PR #702, opened 2026-07-13
fix(skill): scope language-preservation rule to every emitted line

### LINKED ISSUE #701: Pre-tool-call narration lines emitted in English during Chinese sessions, despite full-mode's "no tool-call narration" rule
## Environment
- Claude Code CLI, plugin install (`claude plugin marketplace add JuliusBrussee/caveman` + `claude plugin install caveman@caveman`, scope: user)
- caveman: main @ 0d95a81
- caveman level: `full`
- Session language: Chinese throughout (all user prompts Chinese)

## Issue
With caveman `full` active in an all-Chinese session, the final substantive answer to the user is correctly written in Chinese (per the Rules-section promise "Preserve user's dominant language … Compress the style, not the language"). But the short one-line status statement emitted immediately before each tool call (the harness convention of stating one sentence before invoking a tool) comes out in English, or English-verb + Chinese-noun mixed — even though it's the same turn, same model, same session.

Observed in a real (unrelated) project session:
- Before a Bash call: `Verify. Check EDI code carrier分支.`
- Before another Bash call: `Read send dialog + G1/G2/G4 guard 看是否真共用同一发送机制。`
- The turn's actual answer bullet, fully Chinese: `不是"完全没变化只是PDF不一样"。发送机制共用一套，但有实质分支。`

Note: caveman `full` mode's own Intensity table explicitly says **"No tool-call narration"** — so arguably these lines shouldn't be emitted at all, in any language. Their presence suggests the narration-suppression rule isn't reliably honored, and when narration does slip through, it isn't covered by the language-preservation rule either (that rule's wording targets "reply", and this status line may not register as "the reply" to the model).

## Hypothesis
The pre-tool-call narration is a harness-level convention independent of SKILL.md, so SKILL.md's language-preservation rule doesn't obviously apply to it. When caveman `full` compresses it into terse fragments, the compression pattern likely borrows sentence structure from SKILL.md's own English-only examples (e.g. "Bug in auth middleware. Fix:") — same general mechanism as #679 and #665 (English-only in-context example text biasing output under compression pressure), but a different specific surface (tool narration, not the reply body or the Auto-Clarity block).

## Suggested fix
Either:
1. Extend the language-preservation rule's scope to explicitly cover any auxiliary/status text the model emits, not just the final reply, or
2. Reinforce "No tool-call narration" so this text is suppressed entirely at `full`/`ultra`, sidestepping the language question.

Related: #679 (Auto-Clarity English block leak), #665 (named-language examples risk drift under compaction) — same class of bug, different surface.

serenNan:
Fixes #701.

## Problem

Pre-tool-call status statements (harness-required, not caveman-specific — the "state one sentence before tool use" convention) weren't covered by the Rules-section language-preservation promise, since that rule's wording targeted "reply" specifically. In a fully Chinese session with `full` active, the final answer stayed correctly Chinese, but the one-line status statement immediately before each tool call came out in English or English-verb + Chinese-noun mixed:

- `Verify. Check EDI code carrier分支.`
- `Read send dialog + G1/G2/G4 guard 看是否真共用同一发送机制。`

while the turn's actual answer bullet was fully Chinese. Repro detail and screenshots in #701.

## Fix — 2 added lines in `skills/caveman/SKILL.md`, pure addition, no line rewritten

One sentence appended after the existing language-preservation paragraph (Rules section), making explicit that the rule covers every line the model emits this turn, not just the final reply, and that a harness-required pre-tool-call status line must not default to English.

I considered instead just reinforcing the existing "no tool-call narration" instruction so the line wouldn't be emitted at all — but that narration is a harness-level convention outside caveman's control (the instruction to state one sentence before tool use lives in the harness system prompt, not in SKILL.md), so suppressing it isn't reliably actionable from the skill side. Scoping the language rule to cover it instead is.

Related: #679 (Auto-Clarity English block leak), #665 (named-language examples risk drift under compaction) — same class of bug (English-only in-context text biasing output under compression pressure), different surface. `plugins/caveman/skills/` mirror left for the CI sync workflow per CLAUDE.md.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -22,6 +22,8 @@ Drop: articles (a/an/the), filler (just/really/basically/actually/simply), pleas
 
 Preserve user's dominant language. User write Portuguese → reply Portuguese caveman. User write Spanish → reply Spanish caveman. Compress the style, not the language. No forced English openings or status phrases. ALWAYS keep technical terms, code, API names, CLI commands, commit-type keywords (feat/fix/...), and exact error strings verbatim — unless user explicitly ask for translation.
 
+Rule covers every line the model outputs this turn, not just the final reply — including any harness-required pre-tool-call status line. Never let that line default to English when session language differs.
+
 No self-reference. Never name or announce the style. No "caveman mode on", "me caveman think", no third-person caveman tags. Output caveman-only — never normal answer plus "Caveman:" recap. Exception: user explicitly ask what the mode is.
 
 Pattern: `[thing] [action] [reason]. [next step].`


JuliusBrussee:

PR #708, opened 2026-07-16
fix: make caveman hooks robust to non-standard or absent interpreters

### LINKED ISSUE #65: Use `#!/usr/bin/env sh` in scripts
**What happen**

```
SessionStart:startup says: Plugin hook error: /bin/sh: line 1: node: command not found
```

**Expected**

Script should use `#!/usr/bin/env sh` after shebang in case the interpreter is located at another path.

**Platform**
- [X] Claude Code

**Version / install method**
Add marketplace/plugin as readme suggests.

mvanhorn:
## Summary
Two coordinated changes that both serve issue #65's theme. (1) Portable shebangs: change `src/hooks/caveman-statusline.sh`, `src/hooks/install.sh`, and `src/hooks/uninstall.sh` from hardcoded `#!/bin/bash` to `#!/usr/bin/env bash` (keep `bash` where the scripts use bash-isms; downgrade `caveman-statusline.sh` to `#!/usr/bin/env sh` only if it is verified POSIX-clean).

## Why this matters
Installing the caveman plugin emits `Plugin hook error: /bin/sh: line 1: node: command not found` (SessionStart) and `UserPromptSubmit hook error` for users whose environment does not expose `node` on the hook runner's PATH, or who have no Node at all. Three distinct reporters confirmed it: a NixOS user (`sh` not at the usual path), a Bun-only user (no `node`), and a Debian user who intentionally runs without Node. The literal ask in the title is to use portable `#!/usr/bin/env` shebangs; the thread's most detailed comment (cgty, 2026-07-13) clarifies the real failure: `.claude-plugin/plugin.json` launches the hooks as bare `node "${CLAUDE_PLUGIN_ROOT}/src/hooks/*.js"`, so a shebang change alone cannot help when Node is absent. The single concern is: caveman's hooks must not hard-fail the Claude Code session when the interpreter is at a non-standard path or missing.

## Testing
- Happy path: with `node` on PATH, `run-hook.sh <script.js>` execs Node and the hook runs exactly as before (existing `tests/test_hooks.py` behavior preserved).
- Node absent: with a PATH that has no `node`, `run-hook.sh` exits 0 and produces no error, so the session start / prompt submit degrades gracefully.
- Structural: `.claude-plugin/plugin.json` SessionStart and UserPromptSubmit commands route through `run-hook.sh` (no bare `node` invocation remains).
- Shebang assertions: `src/hooks/caveman-statusline.sh`, `install.sh`, `uninstall.sh` start with a `#!/usr/bin/env` shebang, not a hardcoded `#!/bin/bash`.
- Regression: `bash src/hooks/install.sh` still installs the statusline hook and writes settings.json correctly (existing `HookScriptTests` continue to pass).

Fixes #65

AI was used for assistance.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -11,7 +11,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\"",
+            "command": "sh \"${CLAUDE_PLUGIN_ROOT}/src/hooks/run-hook.sh\" \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\"",
             "timeout": 5,
             "statusMessage": "Loading caveman mode..."
           }
@@ -23,7 +23,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-mode-tracker.js\"",
+            "command": "sh \"${CLAUDE_PLUGIN_ROOT}/src/hooks/run-hook.sh\" \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-mode-tracker.js\"",
             "timeout": 5,
             "statusMessage": "Tracking caveman mode..."
           }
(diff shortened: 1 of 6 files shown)


JuliusBrussee:
Thanks @mvanhorn — current hooks fail open across missing siblings, stream errors, and slow Windows startup. Closing this superseded patch.

Issue #145, opened 2026-04-14
Caveman burns more tokens than saves?


t-lock:
❯ are you using the caveman plugin

● Yes. Full mode.

❯ is it making you faster and burning less tokens?

● Output tokens: yes, cut maybe half. Short replies = fewer tokens out.

  Input tokens: no, worse. Plugin instructions load every turn. Adds overhead.

  Speed: output shorter = done sooner. Per-token speed same.

  Net win only if replies were verbose to start. For terse tech Q&A, plugin overhead might eat the savings.

❯ how much plugin overhead is there on every prompt?

● Rough guess: ~800-1200 tokens per turn for caveman rules block. Plus ~300 for caveman skill entries in skills list. Call it ~1-1.5k
  tokens input overhead per turn.

  Can't measure exact without token counter. To verify: check ~/.claude/plugins/cache/caveman/ files or run them through tokenizer.

  Break-even: if normal reply would be ~1.5k output tokens, caveman saves nothing net. Short Q&A = loss. Long explanation tasks = win.

❯ so for programming like what we have been doing in this chat, what is the average token response size before i turned on caveman?

● Eyeballing my pre-caveman replies this chat: ~100-200 output tokens each. Maybe 150 avg.

  Caveman cuts text to ~30-80 tokens. Save ~70-100 per reply.

  Lose ~1000+ on input overhead per turn.

  Net for this workload: caveman costs more tokens, not saves.

  Win case: long explain/doc tasks where replies run 1k+ tokens. Not terse code Q&A.

t-lock:
Sorry for paste just Claude chat issue, Bad form. Sorry. If want delete, delete.

Benchmark example proof better. If burn more tokens input overhead, core thesis maybe invalid.

Thank

t-lock:
This inspired the following prompt in CLAUDE.md:

```md
## Reply Style

Keep prose replies maximally brief without losing technical substance. No pleasantries, preamble, or trailing summaries. Use line breaks and short paragraphs so replies are easily scannable. Applies to prose only — write code, comments, and commit messages normally.
```

70 token overhead, compared to 1000+ with caveman, with similar results to caveman lite. Works for me!

PR #718, opened 2026-07-18
stats: surface the net-negative regime (#145), not just gross savings

vuminhquang:
## The pain

`docs/HONEST-NUMBERS.md` is blunt about when caveman loses:

> the skill costs ~1–1.5k input tokens every turn. If it saves less output than that, you are paying to use it.

#145 measured exactly that. But the tool the docs point users to — `/caveman-stats` — can't show it. `deriveSavings` only computes gross output savings, so `Est. tokens saved` is always positive and the net-negative case is invisible in the one instrument meant to reveal it. The page's own creed — *"If caveman lose for your workload, this page tell you to turn it off"* — isn't wired into the page.

## The change

Adds the missing half:

- **`Est. rule overhead`** — the per-turn input cost of the injected rules (default `1250` tokens ≈ the ~5 KB `SKILL.md` at the ~4-chars/token approximation already used here; override with `CAVEMAN_RULE_OVERHEAD_TOKENS` if you've measured your own setup).
- **`Est. net`** = output saved − rule input. Negative → it says so and tells you to turn caveman off for that workload.

Shown in the single-mode session view and the lifetime `--all` view.

Before / after on a short reply (120 output tokens, 1 turn, `full`):

```
Est. tokens saved:     223 (~65% of output)          # previously the whole story
Est. rule overhead:    1,250 (input, ~1,250/turn over 1 turn)
Est. net:              -1,027 (caveman cost more than it saved — turn it off for this workload)
```

## Honesty guardrails

- Net is shown only where savings can actually be estimated (full-mode spans). Unattributed spans get **no** net line — no guessing, consistent with the #601 attribution logic.
- Overhead is an estimate, labeled as such, cited to `HONEST-NUMBERS.md`; it's a named constant, env-overridable.
- No new output-token claims; the reduction line still says output-only. Lifetime net only counts sessions logged since this change (older rows carry no turn count, so they're left out rather than guessed).

## Tests

`node tests/test_caveman_stats.js` — **44 pass** (37 existing + 7 new: positive net, the negative-net #145 case, the env override, the no-guessing unattributed path, lifetime aggregation, and the legacy-row omission).

### DIFF
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -18,6 +18,18 @@ const { readFlag, appendFlag, readHistory, safeWriteFlag, VALID_MODES, MODE_LOG_
 // run is committed.
 const COMPRESSION = { 'full': 0.65 };
 
+// Per-turn INPUT cost the skill adds: the SKILL.md rules (~5 KB ≈ ~1,250 tokens
+// at the ~4-chars/token approximation used elsewhere here) are injected into
+// context every turn, plus the per-turn reinforcement the mode tracker emits.
+// A real "net" must subtract this from the output savings — caveman is
+// net-negative when it saves less output than the rules cost (docs/HONEST-NUMBERS.md).
+// Override with CAVEMAN_RULE_OVERHEAD_TOKENS if you've measured your own setup.
+const DEFAULT_RULE_OVERHEAD_TOKENS_PER_TURN = 1250;
+function ruleOverheadPerTurn() {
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks for doing this work. Equivalent behavior is now on main through [`d833f4ad`](https://github.com/JuliusBrussee/caveman/commit/d833f4ad) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected. If your branch contains behavior still missing from current release, point me to exact case and I’ll reopen it.

Issue #596, opened 2026-07-02
openclaw: stray begin marker in SOUL.md → duplicate bootstrap on install, user content deleted on uninstall


AmirF194:
**What happen**

`bin/lib/openclaw.js` append/strip of the SOUL.md bootstrap block can destroy user content when the file contains a stray or truncated begin marker:

1. `appendBootstrapToSoul` (`openclaw.js:119-133`) treats the block as present only if **both** markers exist. With `<!-- caveman-begin -->` present but no end marker (interrupted write, partial user edit), it appends a **second** full block.
2. `stripBootstrapFromSoul` (`openclaw.js:135-155`) then cuts from the **first** `caveman-begin` to the **first** `caveman-end` — which now spans everything between the stray marker and the appended block's end marker.

**Repro (confirmed)**

SOUL.md: truncated begin marker, then `USER IMPORTANT CONTENT`. Install → second block appended. Uninstall → entire file deleted including the user content (`{"changed":true,"removed":true}`).

Idempotency violation → data loss chain. The same single-span strip pattern is duplicated for the opencode AGENTS.md fence at `bin/install.js:1148-1162`.

**Fix**

Treat a lone begin marker as "present but damaged" (repair in place rather than append), and strip marker pairs iteratively/pairwise rather than first-begin→first-end.

main @ 25d22f8.

PR #720, opened 2026-07-18
fix(openclaw): stop stamping stale version, add --no-always opt-out

pierrz:
## Problem

Found while integrating caveman into an OpenClaw deployment and reading `bin/lib/openclaw.js`:

1. **Stale version stamp.** `SKILL_VERSION = '1.0.0'` is hardcoded and gets written into every installed skill's frontmatter regardless of the actual release — currently v1.9.1. `openclaw skills list` reports a version that's never been accurate since the constant was introduced.
2. **No opt-out for always-on.** `installOpenclaw()` unconditionally stamps `always: true` and writes the SOUL.md bootstrap block. There's no flag to install caveman as a discoverable, load-on-demand skill only (which the module's own top-of-file comment describes as a supported mode — "Skills there... are loaded on-demand by the model" — but nothing in the installer actually lets you land in that mode).

## Fix

Both scoped narrowly, no default-behavior change:

- **Version**: `mergeOpenclawFrontmatter()` / `installOpenclaw()` accept an explicit `version` override. `bin/install.js`'s `--only openclaw` path now passes `PINNED_REF` (already bumped on every release for hook-integrity pinning — this reuses that existing, maintained value instead of introducing a second version source). `SKILL_VERSION` stays as the fallback for callers that don't pass one explicitly.
- **`--no-always`**: new flag. Skips the `always: true` frontmatter stamp and the SOUL.md write; only the skill file is installed. Re-running `--no-always` over an already always-on install does **not** retroactively strip the existing SOUL.md block — it only changes what a *new* install does (covered by a test).

## Before / after

```diff
- version: 1.0.0
+ version: 1.9.1
```

```bash
# new: install caveman for OpenClaw without forcing always-on
npx -y github:JuliusBrussee/caveman -- --only openclaw --no-always
```

## Tests

4 new cases in `tests/installer/e2e.freshinstall.test.mjs`:
- version threading (`CAVEMAN_REF` override → frontmatter reflects it, not the fallback)
- `--no-always`: skill written, versioned, no `always:` key, SOUL.md not created
- `--no-always` idempotency: doesn't strip a SOUL.md block from a prior always-on install

Existing openclaw suite (always-on default, idempotent re-run, uninstall, marker damage-tolerance from #596) is unmodified and still green — confirms the default install path is unchanged.

`npm test`: 113/115 pass locally, including all 4 new tests. The 2 failures are **pre-existing** — they reproduce on an unmodified checkout of `main` in this sandbox, caused by a real `claude` CLI + `~/.openclaw` present in the environment (tests documented as "skipped without `claude` CLI" detect it and take a different path). Unrelated to this change; not touched here.

## Scope note

`src/tools/caveman-init.js` (the per-repo rule-file installer) also calls `helper.installOpenclaw()` for its `openclaw` target, and is left unchanged — it's a differently-scoped tool (drops rule files for many editors into a target repo, explicitly documented as installing an "always-on caveman rule") and extending it is a separate concern per the one-PR-one-concern guidance in CONTRIBUTING.md. Happy to follow up there if useful.

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -59,7 +59,7 @@ function parseArgs(argv) {
     withHooks: 'auto', withInit: false, withMcpShrink: false,
     all: false, minimal: false, listOnly: false, noColor: false,
     only: [], uninstall: false, nonInteractive: false,
-    configDir: null, help: false,
+    configDir: null, help: false, openclawAlways: true,
   };
   for (let i = 0; i < argv.length; i++) {
     const a = argv[i];
@@ -102,6 +102,7 @@ function parseArgs(argv) {
         break;
       }
       case '--no-mcp-shrink': opts.withMcpShrink = false; break;
+      case '--no-always': opts.openclawAlways = false; break;
       case '--all': opts.all = true; break;
       case '--minimal': opts.minimal = true; break;
       case '--list': opts.listOnly = true; break;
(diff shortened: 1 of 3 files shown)


JuliusBrussee:

PR #730, opened 2026-07-22
fix(hooks): act on a complete payload instead of waiting for stdin EOF

### LINKED ISSUE #729: UserPromptSubmit hook waits for stdin EOF, so Claude Code kills it at the 5s timeout — 2.3% of runs inject nothing
## Summary

`src/hooks/caveman-mode-tracker.js` does all of its work inside `process.stdin.on('end', …)`. The hook therefore does nothing until the caller closes stdin — and when that EOF is late, Claude Code kills the process at its 5000 ms timeout. The flag file is never written and the per-turn reinforcement is never injected.

The failure is silent. Nothing is printed, the prompt goes through normally, and the only trace is a `hook_cancelled` attachment in the session transcript.

Not a duplicate of #538 (missing `stdin.on('error')`, already fixed on `main`) or #397 (broken-pipe socket write). Those are about stdin closing *abnormally*; this is about stdin not closing *promptly*.

## Evidence

30 days of Claude Code transcripts on one machine, counting `attachment` rows by hook:

| hook | delivered | cancelled at timeout | rate |
|---|---:|---:|---:|
| `UserPromptSubmit` — *"Tracking caveman mode…"* | 2028 | **48** | **2.3%** |
| `SessionStart:startup` — *"Loading caveman mode…"* | 223 | 1 | 0.4% |

The SessionStart hook is the control: `caveman-activate.js` never reads stdin and is essentially never cancelled. The tracker is, and the recorded `durationMs` on the cancelled runs goes up to **25618 ms** against a `timeoutMs` of 5000 — the process is sitting in the event loop waiting for an EOF, not doing work.

Reproduction against `main` (0d95a81) — write one complete payload, leave stdin open, which is exactly what a caller that has not yet closed the pipe does:

```python
p = subprocess.Popen(['node', 'src/hooks/caveman-mode-tracker.js'], stdin=PIPE, stdout=PIPE)
p.stdin.write(b'{"prompt":"hello"}'); p.stdin.flush()   # complete payload, stdin stays open
p.wait(timeout=6)
```

```
UPSTREAM main: STILL RUNNING at 6.01s -> would be killed at the 5s hook timeout
```

With `~/.claude/.caveman-active` set to `full`, the injection the hook exists to produce is simply absent:

```
BEFORE: STILL RUNNING at 6.01s -> killed at the 5s hook timeout
   injected: b''
```

## Impact

Roughly 1 prompt in 43 loses its per-turn reinforcement — the mechanism added precisely because models drift back to prose when other plugins inject competing style instructions (#233, #303). It also means a `/caveman <level>` or `/caveman off` typed on an unlucky turn is dropped: the flag file is never written, so the mode silently does not change.

## Suggested fix

Act on a **complete payload** rather than on EOF: parse on each `data` chunk, run the body as soon as the accumulated input is valid JSON, then destroy stdin. A partial payload is not an error — it just keeps reading — and the `end` listener stays as the fallback for input that never parses.

Destroying stdin rather than calling `process.exit()` matters: stdin is the only handle holding the event loop open, so releasing it lets Node drain and flush stdout on its own, whereas exiting outright can truncate the injection.

PR follows with three regression tests, each verified red against the current hook.

kniknak:
Fixes #729

`caveman-mode-tracker.js` does all of its work inside `process.stdin.on('end', …)`, so the hook sits idle until the caller closes stdin. When that EOF is late, Claude Code kills the process at its 5000 ms timeout — before the flag file is written and before the per-turn reinforcement is injected. Silently: nothing is printed, and the only trace is a `hook_cancelled` attachment in the transcript.

Measured over 30 days of transcripts on one machine: **48 of 2076** `UserPromptSubmit` runs cancelled at the timeout (**2.3%**), some with a recorded `durationMs` of 25618 against `timeoutMs: 5000`. The sibling `SessionStart` hook — which never reads stdin — was cancelled once in 224 runs. Full numbers in the issue.

## The change

Parse on each `data` chunk and run the body as soon as the accumulated input is valid JSON, then destroy stdin. A partial payload is not an error, it just keeps reading; the `end` listener stays as the fallback for input that never parses, and a `handled` flag keeps the body from running twice when both fire.

**`process.stdin.destroy()` rather than `process.exit()`** — stdin is the only handle holding the event loop open, so releasing it lets Node drain and flush stdout on its own. Exiting outright races the write and can truncate the injection.

The diff is confined to the stdin wiring; the body is unchanged apart from being moved into a named function (the early `return` in the `/caveman-stats` branch behaves identically inside it).

## Tests

Three cases added to `tests/test_mode_tracker_stdin.js`, each verified to fail without the fix:

```
$ node tests/test_mode_tracker_stdin.js          # with the fix
  ✓ stdin "error" event does not crash the hook (exit 0)
  ✓ normal stdin (valid JSON + clean EOF) still exits 0
  ✓ a complete payload is acted on before the 5s hook timeout, without EOF
  ✓ the per-turn reinforcement is still emitted when stdin never closes
  ✓ a payload split across chunks is handled once, and only once
  5 passed, 0 failed
```

Reverting only the hook (keeping the tests):

```
  ✗ a complete payload is acted on before the 5s hook timeout, without EOF
    hook did not exit on its own with stdin held open
    (code=null signal=SIGKILL after 10006ms) — Claude Code kills it at 5000ms
  ✗ the per-turn reinforcement is still emitted when stdin never closes
    expected the additionalContext injection on stdout, got ""
  3 passed, 2 failed
```

The chunked case guards the new code path, so it is proven against a mutation of the fix instead — dropping the `handled` guard:

```
  ✗ a payload split across chunks is handled once, and only once
    expected exactly one injection for one payload, got 2
```

The driver spawns the hook through a child process so the assertions stay synchronous, matching the existing file's `spawnSync` style.

Also green, unchanged: `npm test` (112/112), `tests/test_mode_tracker.py` (23), `tests/test_hooks.py`, `tests/test_symlink_flag.js` (12), `tests/test_caveman_init.js` (8), `tests/test_repo_local_config.js` (12).

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -19,12 +19,22 @@ const flagPath = path.join(claudeDir, '.caveman-active');
 const prevPath = path.join(claudeDir, '.caveman-active.prev');
 
 let input = '';
-process.stdin.on('data', chunk => { input += chunk; });
+let handled = false;
 // Abnormal stdin close (broken pipe, parent crash) emits 'error'; without a
 // listener Node throws it as an uncaught exception and the hook exits
 // non-zero — a spurious hook failure (#538). Hooks must always exit 0.
 process.stdin.on('error', () => process.exit(0));
-process.stdin.on('end', () => {
+
+// Act on a COMPLETE payload rather than waiting for stdin to close. The whole
+// body used to live in the 'end' listener alone, so the hook did nothing until
+// the caller sent EOF — and Claude Code killed it at the 5s timeout whenever
+// that EOF was late. The sibling SessionStart hook never reads stdin and never
+// times out; this one did, on ~2% of runs. Destroying stdin (rather than
+// exiting) releases the only thing holding the event loop open, so Node drains
+// and flushes stdout on its own.
+function handlePayload() {
+  if (handled) return;
+  handled = true;
   try {
     const data = JSON.parse(input);
     // Collapse whitespace so phrase triggers still match multiline prompts —
@@ -203,4 +213,13 @@ process.stdin.on('end', () => {
   } catch (e) {
     // Silent fail
   }
+}
+
+process.stdin.on('data', chunk => {
+  input += chunk;
+  // A partial payload is not an error — keep reading until it parses.
+  try { JSON.parse(input); } catch (e) { return; }
+  handlePayload();
+  process.stdin.destroy();
 });
+process.stdin.on('end', handlePayload);
--- a/tests/test_mode_tracker_stdin.js
+++ b/tests/test_mode_tracker_stdin.js
@@ -81,5 +81,102 @@ test('normal stdin (valid JSON + clean EOF) still exits 0', () => {
   }
 });
 
+// --- Claude Code's 5s hook timeout -----------------------------------------
+// A complete payload can arrive well before the caller closes stdin. Doing the
+// work only in the 'end' listener meant the hook sat idle until EOF and was
+// killed at the timeout — silently, injecting nothing.
+//
+// Driven through a child so the assertions here stay synchronous: the driver
+// writes one complete payload, leaves stdin OPEN, and reports how long the hook
+// took to exit on its own plus whatever it managed to write.
+const HOOK_TIMEOUT_MS = 5000;
+
+function runWithStdinHeldOpen(configDir, prompt) {
+  const driver = `
+    const { spawn } = require('child_process');
+    const started = Date.now();
+    const p = spawn(process.execPath, [${JSON.stringify(HOOK_PATH)}], {
+      stdio: ['pipe', 'pipe', 'ignore'],
+      env: { ...process.env, CLAUDE_CONFIG_DIR: ${JSON.stringify(configDir)} },
+    });
+    let out = '';
+    p.stdout.on('data', c => { out += c; });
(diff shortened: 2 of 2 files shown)


kniknak:
Superseded — closing.

`783dbea` landed the same fix on `main`: the tracker now parses per `data` chunk and `unref()`s stdin instead of waiting for EOF, which is strictly better than the `destroy()` this branch used. Rebasing onto current `main` drops this commit to a zero-line diff, and `5429b5d` added `tests/test_hook_stdin_lifecycle.js`, which already covers the cases added here.

Verified against `main` (17f9f2e) with the payload written and stdin deliberately held open — the #729 condition:

```
PASS hold  "hello there"     ms=  30 code=0 injected=true  flag=full
PASS hold  "/caveman ultra"  ms=  28 code=0 injected=true  flag=ultra
```

Same probe against the pre-fix EOF-only wiring, as a control:

```
FAIL hold  "hello there"     ms=10005 SIGKILL injected=false
FAIL hold  "/caveman ultra"  ms=10006 SIGKILL injected=false flag unchanged
```

Suites green on `main`: `test_hook_stdin_lifecycle.js` 4/4, `test_mode_tracker_stdin.js` 17/17, `test_mode_tracker.py` 24/24.

#729 is still open, but is fixed on `main` by `783dbea` — worth closing too.

PR #732, opened 2026-07-22
Integrate caveman as a native plugin within Codex

jamesqo:
I noticed that with Claude the caveman speak seemed a lot more ingrained, but with Codex it would eventually stop and go back to normal language over time. Codex recently caught up to Claude Code and added native support for plugins; this PR adds it as a native plugin to Codex so it doesn't "drift" back to the normal speak over time -- I've tested it on my local machine (M4 MacBook Air), and it seems to be working well.

One thing to note is that if you run Codex from the command line it's gonna be slightly annoying because it'll remind you with every prompt that it's running the caveman hook (unlike Claude which lets you silence that). But if you're running it from the desktop app, it's not an issue.

I asked my agent (5.6 Sol) to look over this PR a few times to add tests / validate correctness and make sure it follows the repo's coding conventions -- also did a manual review once it was finished. Let me know if there's anything else you think I can add!


JuliusBrussee:
Thanks @jamesqo — native Codex integration now ships through the current profile/runtime. Closing this old implementation.

Issue #711, opened 2026-07-16
statusline: empty .caveman-statusline-suffix file makes script exit 1 — Claude Code hides the status bar


redagavin:
## Summary

If `$CLAUDE_CONFIG_DIR/.caveman-statusline-suffix` exists but is **empty** (0 bytes), `src/hooks/caveman-statusline.sh` prints the badge correctly but exits with status **1**. Claude Code hides the status bar entirely when the statusline command exits non-zero, so the caveman badge silently disappears even though the script's stdout is fine.

## Root cause

The savings-suffix block is the last thing in the script:

```bash
if [ "${CAVEMAN_STATUSLINE_SAVINGS:-1}" != "0" ]; then
  SAVINGS_FILE="${CLAUDE_CONFIG_DIR:-$HOME/.claude}/.caveman-statusline-suffix"
  if [ -f "$SAVINGS_FILE" ] && [ ! -L "$SAVINGS_FILE" ]; then
    SAVINGS=$(head -c 64 "$SAVINGS_FILE" 2>/dev/null | tr -d '\000-\037')
    [ -n "$SAVINGS" ] && printf ' \033[38;5;172m%s\033[0m' "$SAVINGS"
  fi
fi
```

When the suffix file exists but is empty (or contains only control bytes), `[ -n "$SAVINGS" ]` is false, so the `A && B` compound returns 1. Since it's the last command executed, the whole script exits 1.

The same failure mode applies to a suffix file that ends up empty for any reason — e.g. a `caveman-stats.js` run interrupted mid-write, or a session with no recorded savings.

## Repro

```bash
touch "${CLAUDE_CONFIG_DIR:-$HOME/.claude}/.caveman-statusline-suffix"
echo '{}' | bash src/hooks/caveman-statusline.sh; echo "exit=$?"
# prints the badge, then: exit=1
# → Claude Code stops rendering the status bar
```

## Fix

Terminate the script with an explicit success, e.g. append at the end:

```bash
exit 0
```

(or restructure so `[ -n "$SAVINGS" ] && printf ...` isn't the final command, e.g. `if [ -n "$SAVINGS" ]; then printf ...; fi`). Verified locally that with `exit 0` appended, all three states — no suffix file, empty suffix file, populated suffix file — exit 0 and render correctly.

## Environment

- caveman plugin via Claude Code plugin marketplace (cache rev `0d95a81d35a9`); bug also present in current `main`
- Claude Code on Linux (RHEL 9), bash 5.x

[...]

PR #735, opened 2026-07-23
fix(hooks): make caveman-statusline.sh always exit 0

AmirF194:
Refs #711.

## Update: superseded by 7693a04, narrowed to regression tests

This PR originally added the same fix independently. Rebasing onto current
`main` showed that `7693a04` already shipped the identical fix upstream
(an explicit `exit 0` after the savings-suffix block in
`caveman-statusline.sh`, for the same reason described below), so that part
of this PR is now redundant and has been dropped in the rebase. The
`caveman-statusline.sh` diff here is empty; it now matches `main` exactly.

What remains is `StatuslineExitCodeTests` in `tests/test_hooks.py`, which
`main` does not have. It pins the behavior in place: empty suffix file,
missing suffix file, and populated suffix file must all exit 0.

Entirely your call whether these tests are worth merging as-is, folding
into another PR, or just closing this one.

## Original root cause (for context, already fixed on main by 7693a04)

`caveman-statusline.sh`'s last statement is:

```sh
[ -n "$SAVINGS" ] && printf ' \033[38;5;172m%s\033[0m' "$SAVINGS"
```

When `.caveman-statusline-suffix` exists but is empty (interrupted write, or
a fresh install before `caveman-stats` has produced its first value),
`$SAVINGS` is empty, the test fails, and since this is the last command in
the script its exit status (1) becomes the script's exit status, even
though the badge itself already printed correctly. Claude Code hides the
whole statusline on a non-zero exit, so the badge disappears entirely
instead of just dropping the optional suffix.

## Verification

Rebased onto current `main` (`fcf7663`) and ran the full `tests/test_hooks.py`
and `npm test` suites in a `node:22-slim` + `python3` container: 10/10 and
125/125 (121 pass, 4 pre-existing skips) green, no regressions. The three
new `StatuslineExitCodeTests` pass against `main`'s already-fixed script.

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -215,5 +215,50 @@ class HookScriptTests(unittest.TestCase):
             self.assertIn("PLUGIN ROOT MARKER RULESET", result.stdout)
 
 
+class StatuslineExitCodeTests(unittest.TestCase):
+    def run_statusline(self, config_dir, extra_env=None):
+        env = os.environ.copy()
+        env["CLAUDE_CONFIG_DIR"] = str(config_dir)
+        if extra_env:
+            env.update(extra_env)
+        return subprocess.run(
+            ["bash", str(REPO_ROOT / "src" / "hooks" / "caveman-statusline.sh")],
+            env=env,
+            text=True,
+            capture_output=True,
+        )
+
+    def test_exits_zero_with_empty_savings_suffix_file(self):
+        with tempfile.TemporaryDirectory(prefix="caveman-statusline-empty-suffix-") as tmp:
+            config_dir = Path(tmp)
+            (config_dir / ".caveman-active").write_text("full")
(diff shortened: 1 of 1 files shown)


AmirF194:
No rush, just flagging this is still open: mergeable and no CI failures on this repo, so it should apply cleanly whenever you get a chance to look.

AmirF194:
Thanks for landing the fix in 7693a04, that already covers the underlying bug in #711. I rebased this branch onto current main and narrowed it down: the exit-0 change is dropped since it's redundant now, and the only thing left is the three regression tests in `StatuslineExitCodeTests` (empty suffix file, missing suffix file, populated suffix file), which main doesn't have yet. All pass against main's already-fixed script. Happy for you to merge these as-is, fold them into another PR, or just close this one, whatever works best for you.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #737, opened 2026-07-23
test(hooks): pin curl-pipe REPO_URL to src/hooks against regression

### LINKED ISSUE #736: src/hooks/install.sh and install.ps1 curl-pipe fallback 404s: REPO_URL still points at pre-refactor hooks/ path
## Summary

The curl-pipe fallback in `src/hooks/install.sh` and `src/hooks/install.ps1` downloads every hook file from a hardcoded URL that still points at the pre-refactor top-level `hooks/` directory. Every file in `HOOK_FILES`/`$HookFiles` 404s, so the documented one-liner install fails.

## Repro

```bash
curl -o /dev/null -w '%{http_code}\n' https://raw.githubusercontent.com/JuliusBrussee/caveman/main/hooks/caveman-config.js
# 404
curl -o /dev/null -w '%{http_code}\n' https://raw.githubusercontent.com/JuliusBrussee/caveman/main/src/hooks/caveman-config.js
# 200
```

Running the script's own documented curl-pipe form:

```bash
bash <(cat src/hooks/install.sh)
```

prints `Installing caveman hooks...`, then aborts on the first `curl -fsSL "$REPO_URL/$hook" -o ...` with a 404 under `set -e`, so zero hooks get installed.

## Root cause

`REPO_URL` in `src/hooks/install.sh:38` and `$RepoUrl` in `src/hooks/install.ps1:25` are both hardcoded to `.../main/hooks`. Commit `1ce38a2` ("refactor: consolidate hooks/ rules/ tools/ mcp-servers/ under src/") moved the physical files to `src/hooks/` but never updated these two download URLs, which only run on the curl-pipe path (no local clone). The local-clone path (`cp` from `$SCRIPT_DIR`) is unaffected, which is why the existing tests, all of which run against a local clone, never caught it.

## Fix

Update both URLs to `.../main/src/hooks`.

## Environment

Reproduced against current `main` (`0d95a81`).

AmirF194:
## Root cause

The curl-pipe fallback in `src/hooks/install.sh` (`REPO_URL`, line 38) and `src/hooks/install.ps1` (`$RepoUrl`, line 25) was hardcoded to `.../main/hooks` after commit `1ce38a2` moved the physical hook files to `src/hooks/`. Since fixed independently on `main` by `82864c8` (the 2.0.0 release), as a side effect of the src/ restructure, this PR no longer changes either URL.

## What is left

A static assertion in `tests/verify_repo.py` pinning both URLs to `main/src/hooks`, so a future path move cannot silently reintroduce this class of bug (the repo restructures paths often, most recently in 82864c8 itself).

## How I verified

- Confirmed both URLs already match on current `main` (`82864c8`).
- Ran the repo own `python3 tests/verify_repo.py` on the rebased branch: all checks pass, including the new assertion.

Refs #736 (the underlying URL bug there is already fixed on main too; leaving it to a maintainer to close since it was never a PR of ours to close).

### DIFF
--- a/src/hooks/install.ps1
+++ b/src/hooks/install.ps1
@@ -22,7 +22,7 @@ if (-not (Get-Command node -ErrorAction SilentlyContinue)) {
 $ClaudeDir = if ($env:CLAUDE_CONFIG_DIR) { $env:CLAUDE_CONFIG_DIR } else { Join-Path $env:USERPROFILE ".claude" }
 $HooksDir = Join-Path $ClaudeDir "hooks"
 $Settings = Join-Path $ClaudeDir "settings.json"
-$RepoUrl = "https://raw.githubusercontent.com/JuliusBrussee/caveman/main/hooks"
+$RepoUrl = "https://raw.githubusercontent.com/JuliusBrussee/caveman/main/src/hooks"
 
 $HookFiles = @("package.json", "caveman-config.js", "caveman-activate.js", "caveman-mode-tracker.js", "caveman-stats.js", "caveman-statusline.sh", "caveman-statusline.ps1", "cavecrew-model-overrides.js")
 
--- a/src/hooks/install.sh
+++ b/src/hooks/install.sh
@@ -35,7 +35,7 @@ fi
 CLAUDE_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"
 HOOKS_DIR="$CLAUDE_DIR/hooks"
 SETTINGS="$CLAUDE_DIR/settings.json"
-REPO_URL="https://raw.githubusercontent.com/JuliusBrussee/caveman/main/hooks"
+REPO_URL="https://raw.githubusercontent.com/JuliusBrussee/caveman/main/src/hooks"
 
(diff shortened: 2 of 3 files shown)


AmirF194:
No rush, just flagging this is still open: mergeable and no CI failures on this repo, so it should apply cleanly whenever you get a chance to look.

AmirF194:
CI failed on typescript and windows: `pnpm-lock.yaml` never picked up `@caveman-ai/cli@^1.1.0` that's already in `package.json`, so `pnpm install --frozen-lockfile` fails. That's on main too, not something this branch introduced. Regenerated the lockfile at 3f0207e to unblock CI here.

AmirF194:
Update: main's 2.0.0 release (82864c8) fixed REPO_URL/$RepoUrl independently, as a side effect of the src/ restructure, so this PR's URL-string change is now a no-op against main. Rebased onto current main and dropped it, along with the lockfile-regen commit from an earlier pass (main's own e3ccb25 already fixed that separately).

What's left and still real: the static assertion in tests/verify_repo.py pinning both URLs to main/src/hooks, so a future path move can't silently reintroduce this. Ran tests/verify_repo.py on the rebased branch, all checks pass. Force-pushed the same branch/PR.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #738, opened 2026-07-23
fix: remove command files that duplicate same-named skills

rootelement:
## Problem

Claude Code registers both `commands/*.md` and `skills/*/SKILL.md`. Since the plugin ships both for `caveman`, `caveman-commit`, `caveman-review`, and `caveman-stats`, each appears **twice** in every session's available-skills listing (once with the short command description, once with the long skill description) — roughly 500 wasted context tokens per session for every plugin user, which runs against the plugin's whole purpose.

## Fix

Delete the four shadowed command `.md` files. The skills are already user-invocable as `/caveman`, `/caveman-commit`, `/caveman-review`, `/caveman-stats`, so no functionality is lost.

- `commands/caveman-init.md` kept — it has no skill counterpart.
- `.toml` command variants (other runtimes) untouched.

https://claude.ai/code/session_01BSexBNDnKmWe64ARv5r1wv

### DIFF
--- a/commands/caveman-commit.md
+++ b/commands/caveman-commit.md
@@ -1,5 +1,0 @@
----
-description: Generate terse caveman-style commit message
----
-
-Generate a terse commit message for the current staged changes. Conventional Commits format. Subject: ≤50 chars, imperative, lowercase after type. Body: only when 'why' isn't obvious from subject. Why over what. No period on subject.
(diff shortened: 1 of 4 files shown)


JuliusBrussee:
Thanks @rootelement — duplicate command shadows are removed and guarded. Closing as implemented.

PR #739, opened 2026-07-23
Clarify rule: shave tokens not characters, keep word forms intact

motlin:


JuliusBrussee:
Thanks @motlin — the canonical skill now states the token-not-character and intact-word rule directly. Closing as implemented.

PR #740, opened 2026-07-23
fix(opencode): record mode changes

### LINKED ISSUE #734: OpenCode plugin never calls recordModeChange - /caveman-stats has no data
**Bug: OpenCode plugin.js never writes history, so /caveman-stats is always empty**

Environment
caveman installed via npx -y github:JuliusBrussee/caveman -- --only opencode
Files installed: ~/.config/opencode/plugins/caveman/plugin.js, caveman-config.cjs, package.json
OS: Windows 11, PowerShell 5.1
What's happening
plugins/caveman/plugin.js imports only a subset of exports from caveman-config.cjs:

const { getDefaultMode, safeWriteFlag, readFlag, VALID_MODES } = config;
But caveman-config.cjs also exports appendFlag, readHistory, recordModeChange, MODE_LOG_BASENAME — none of which are imported or called anywhere in plugin.js.

Meanwhile, commands/caveman-stats.md (installed into ~/.config/opencode/commands/caveman-stats.md) instructs the model to:

Read the lifetime history log at ~/.config/caveman/.caveman-history.jsonl (or wherever the caveman-stats script writes it). Output: total saved, sessions counted, avg ratio.

Since nothing in the OpenCode plugin ever calls recordModeChange/appendFlag, this history file is never created or updated. Running /caveman-stats in OpenCode therefore always reads an empty/nonexistent log — it can never report real numbers, unlike the Claude Code integration where hooks/caveman-mode-tracker.js presumably wires this up.

Expected
plugin.js's chat.message handler (or the session-created handler) should call recordModeChange/appendFlag whenever the caveman mode is toggled, mirroring whatever caveman-mode-tracker.js does for Claude Code, so /caveman-stats has real data to read in OpenCode too.

Suggested fix location
plugin.js around applyModeChange() (~line 163) and handleSessionCreated() (~line 175) — both currently only call safeWriteFlag/unlinkSync, never recordModeChange/appendFlag.

ousamabenyounes:
Fixes #734

## Summary

The opencode plugin imported `getDefaultMode`, `safeWriteFlag` and `readFlag` from
`caveman-config`, but never `recordModeChange` — so every caveman mode transition in
opencode wrote the flag file and nothing else. `.caveman-mode-log.jsonl` was never
created, and `caveman-stats` had no transitions to attribute messages to.

This wires `recordModeChange` into all four transition sites the plugin owns:

- `applyModeChange` — slash/natural-language activation (`set`) and deactivation (`clear`)
- `handleSessionCreated` — session start, both the resolved-mode path and the
  `off` path

`opencodeConfigDir()` is hoisted to a module-level `opencodeDir` so the log lands
beside the flag it describes, instead of recomputing the directory per call.

Ordering matters and is deliberate: `recordModeChange` reads the *current* flag to
fill the row's `prev` field, so it runs before `safeWriteFlag`/`removeFlag`, never
after. Writes go through the existing symlink-safe `appendFlag` helper and inherit
its silent-fail contract, so a bad log path can never break session start.

## Test verification (RED → GREEN)

RED — unmodified upstream `main` (`a42ef76`) with **only** the test changes applied,
production `plugin.js` untouched:

```text
$ git status --porcelain
 M tests/installer/opencode.test.mjs

$ node --test --test-force-exit tests/installer/opencode.test.mjs
not ok 11 - opencode plugin handles /caveman ultra, stop caveman, and session init via real hooks
  error: 'mode log missing after slash activation'
  code: 'ERR_ASSERTION'
  expected: true
  actual: false
# tests 12
# pass 11
# fail 1
```

GREEN — same tests with the fix:

```text
$ node --test --test-force-exit tests/installer/opencode.test.mjs
# tests 12
# pass 12
# fail 0
```

Every `recordModeChange` call site the fix adds is asserted, not merely executed:
activation asserts the row's `mode`, deactivation asserts `mode: null`, and both
`handleSessionCreated` branches assert `{mode, prev}` so the session-start
transition is pinned in both directions.

## Scope note

`src/plugins/opencode/commands/caveman-stats.md` still points the model at
`~/.config/caveman/.caveman-history.jsonl`. That is the lifetime-savings file, a
different artifact from the mode log, and nothing on the opencode side writes it
today — it was already stale before this change. Deciding what opencode's
`/caveman-stats` should read is a separate call, so it is left out of a PR titled
"record mode changes" rather than folded in silently.

### DIFF
--- a/src/plugins/opencode/plugin.js
+++ b/src/plugins/opencode/plugin.js
@@ -74,7 +74,7 @@ function loadConfig() {
 }
 const config = loadConfig();
 
-const { getDefaultMode, safeWriteFlag, readFlag } = config;
+const { getDefaultMode, safeWriteFlag, readFlag, recordModeChange } = config;
 
 // Load the shared mode-change parser (#602) the same way loadConfig() loads
 // caveman-config.js — see the doc comment above loadConfig() for why this
@@ -103,7 +103,8 @@ function opencodeConfigDir() {
   return path.join(os.homedir(), '.config', 'opencode');
 }
 
-const flagPath = path.join(opencodeConfigDir(), '.caveman-active');
+const opencodeDir = opencodeConfigDir();
+const flagPath = path.join(opencodeDir, '.caveman-active');
 
 function removeFlag() {
   try {
@@ -170,10 +171,12 @@ function reinforcementLine(mode) {
 function applyModeChange(change) {
   if (!change) return;
   if (change.action === 'clear') {
+    recordModeChange(opencodeDir, null);
     removeFlag();
     return;
   }
   if (change.action === 'set' && change.mode) {
+    recordModeChange(opencodeDir, change.mode);
     safeWriteFlag(flagPath, change.mode);
   }
 }
@@ -184,9 +187,11 @@ function applyModeChange(change) {
 function handleSessionCreated() {
   const mode = getDefaultMode();
   if (mode === 'off') {
+    recordModeChange(opencodeDir, null);
     removeFlag();
     return;
   }
+  recordModeChange(opencodeDir, mode);
   safeWriteFlag(flagPath, mode);
 }
 
--- a/tests/installer/opencode.test.mjs
+++ b/tests/installer/opencode.test.mjs
@@ -19,6 +19,7 @@ const REPO_ROOT = path.resolve(HERE, '..', '..');
 const INSTALLER = path.join(REPO_ROOT, 'bin', 'install.js');
 const requireCjs = createRequire(import.meta.url);
 const SETTINGS = requireCjs(path.join(REPO_ROOT, 'bin', 'lib', 'settings.js'));
+const MODE_LOG_BASENAME = '.caveman-mode-log.jsonl';
 
 const IS_WIN = process.platform === 'win32';
 
@@ -357,6 +358,7 @@ test('opencode plugin handles /caveman ultra, stop caveman, and session init via
 
     const pluginPath = path.join(xdg, 'opencode', 'plugins', 'caveman', 'plugin.js');
     const flagPath = path.join(xdg, 'opencode', '.caveman-active');
+    const modeLogPath = path.join(xdg, 'opencode', MODE_LOG_BASENAME);
 
     // Set XDG_CONFIG_HOME so the plugin's flagPath resolves to our temp dir,
     // and pin the default mode so session-init is deterministic regardless of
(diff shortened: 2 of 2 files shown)


ousamabenyounes:
Rebased onto `main` (`a42ef76`) and pushed `95ce814`.

**What changed since the last push**

- Rebased off the stale base. That alone turned `typescript` green — the old base
  still carried the pre-`f20acbb` `pnpm-lock.yaml`, so the job was dying in
  `pnpm install --frozen-lockfile` before running a test. No code change was
  needed for it.
- Added assertions so every `recordModeChange` call site the fix introduces is
  asserted rather than merely executed. `handleSessionCreated` previously ran
  under test without anything checking that it logged its transition; both of its
  branches now assert `{mode, prev}`.

**Validation**

RED on unmodified `main` with only the test changes applied (production
`plugin.js` untouched) → `not ok 11 … error: 'mode log missing after slash
activation'`, 11/12 passing. GREEN with the fix → 12/12. Full excerpts are in the
PR description.

Full local replay of both CI workflows on `origin/main` and on this head, same
host: iso-baseline, with `npm test` 171/171 and `root-surfaces` green on both.

Remote CI on `95ce814`: 11/12 green. The one red job, `windows`, is the standing
failure already on `main` — `not ok 157 - pruneOrphanedManagedHooks keeps a
foreign-platform absolute path it cannot judge`, `# fail 1` — identical on
`main`'s own run and untouched by this diff.

**Two things worth your call**

1. `src/plugins/opencode/commands/caveman-stats.md` still points the model at
   `~/.config/caveman/.caveman-history.jsonl`. That is the lifetime-savings file,
   not the mode log, and nothing on the opencode side writes it — stale before
   this change and still stale after. Deciding what opencode's `/caveman-stats`
   should actually read felt like a separate call, so I left it out of a PR
   titled "record mode changes" rather than fold it in quietly. Happy to take it
   here if you'd prefer one PR.
2. #734 was closed by its reporter on 2026-08-18, with no commit or PR attached.
   The defect it describes is still present on `main` — `plugin.js` never calls
   `recordModeChange` — so I've kept the PR open, but flagging it in case you
   consider the issue settled.

AmirF194 (commented):
I looked at the failing windows job here (run 32422471669, job 96597231497) since mergeStateStatus reads UNSTABLE.

The only failing test is `pruneOrphanedManagedHooks keeps a foreign-platform absolute path it cannot judge` (tests/installer/unit.settings.test.mjs:292), which has nothing to do with plugin.js. That guard was POSIX-only by construction until 1b7ff8f (2026-08-25, five days after this branch's last commit) made the platform flavour explicit instead of host-derived. Rebasing onto main should turn this green without touching anything in this PR.

```
not ok 157 - pruneOrphanedManagedHooks keeps a foreign-platform absolute path it cannot judge
  error: must not prune a hook whose path belongs to another platform / 1 !== 0
```

JuliusBrussee:
Adopted into #1015, cherry-picked with authorship preserved. Correct and minimal, and `recordModeChange` was the right helper — it diffs before appending and silent-fails internally, so it can't spam or throw from inside.

One thing I hardened: the new export was destructured alongside the three long-standing ones. `loadConfig()` reads whatever `caveman-config.cjs` sits in the *installed* plugin dir, which can predate `plugin.js` (#848), and `handleSessionCreated()` runs at factory time outside any `try` — so on a drifted install that call throws during plugin construction and caveman doesn't load on opencode at all. Now resolved against a no-op stub, with a test mirroring the stale-config case already in that suite.

Scope note, not a criticism of the patch: this closes the write half only. `commands/caveman-stats.md` reads `.caveman-history.jsonl` (the lifetime savings log, written by `caveman-stats.js`, which doesn't run on opencode), while `recordModeChange` writes `.caveman-mode-log.jsonl`. So `/caveman-stats` on opencode still reads a file nothing writes — flagged for the maintainer as separate work.

---

JuliusBrussee:

PR #741, opened 2026-07-23
fix(config): read POSIX fallback on Windows

### LINKED ISSUE #723: Windows: ~/.config/caveman/config.json silently ignored — real path is %APPDATA%\caveman\config.json
## Problem

On native Windows (`process.platform === 'win32'`), `getConfigDir()` in
`caveman-config.js` resolves the user config path to
`%APPDATA%\caveman\config.json`. It only falls back to
`~/.config/caveman/config.json` if `XDG_CONFIG_HOME` is set — which it
isn't, by default, on a stock Windows install.

The doc comment at the top of `caveman-config.js` lists the resolution
order as:
  1. CAVEMAN_DEFAULT_MODE env var
  2. repo-local .caveman/config.json / .caveman.json
  3. ~/.config/caveman/config.json (macOS / Linux fallback) — OR
     %APPDATA%\caveman\config.json (Windows fallback)
  4. 'full'

This is accurate, but easy to misread on a skim — nothing else in the
plugin (README, /caveman-help card) foregrounds the Windows-specific
path. A Windows user (or an AI assistant editing config on their
behalf, using generic Unix-style docs/muscle memory) naturally writes
`defaultMode` to `~/.config/caveman/config.json` — which resolves via
Git Bash / any POSIX-ish shell to a real, writable path
(`C:\Users\<user>\.config\caveman\config.json`) — and sees **zero
effect**, silently. No warning, no error. The file just isn't in the
lookup chain.

## Impact

Wasted a full debugging session (confirmed independently twice, same
day, in two parallel Claude Code sessions) before finding the actual
active path. Symptom looked exactly like "caveman ignores 'stop
caveman' / ignores defaultMode" — actually just two config files, one
live, one silently inert.

## Suggested fixes (any one would help)

1. On win32, also check `~/.config/caveman/config.json` as a secondary
   fallback before the hardcoded `'full'` — cheap, matches what users
   intuitively write.
2. Or: if no config is found at the platform-correct path but one
   *is* found at the other platform's path, emit a warning (stderr,
   or a SessionStart notice) — "found a config at X, but the active
   path on this OS is Y".
3. Or at minimum, make the Windows-specific path more prominent in
   README / caveman-help — right now it's one clause in a longer
   bullet.

Related: #375 (same category — Windows config-path assumption
silently no-ops, different plugin surface). #447's proposed `manual`
defaultMode would have surfaced this faster (off-by-default rather
than fighting the `full` fallback).

## Environment
- OS: Windows 11
- Shell: PowerShell primary, Git Bash available
- `XDG_CONFIG_HOME`: unset

ousamabenyounes:
Fixes #723

## Summary

- Add a Windows-only fallback from `%APPDATA%\caveman\config.json` to `~/.config/caveman/config.json` when `XDG_CONFIG_HOME` is unset.

- Cover the silent fallback case with a simulated win32 config-resolution regression test.

## Test verification (RED / GREEN)
RED before production fix:
```text
node tests/test_repo_local_config.js
'full' !== 'lite'

```
GREEN after fix:
```text
node tests/test_repo_local_config.js
13 passed, 0 failed

node tests/test_symlink_flag.js
12 passed, 0 failed

npm test
# tests 112
# pass 112
# fail 0
```

### DIFF
--- a/src/hooks/caveman-config.js
+++ b/src/hooks/caveman-config.js
@@ -24,12 +24,14 @@ const VALID_MODES = [
   'wenyan-lite', 'wenyan', 'wenyan-full', 'wenyan-ultra',
   'commit', 'review', 'compress'
 ];
+const CONFIG_FILE_NAME = 'config.json';
+const WINDOWS_PLATFORM = 'win32';
 
 function getConfigDir() {
   if (process.env.XDG_CONFIG_HOME) {
     return path.join(process.env.XDG_CONFIG_HOME, 'caveman');
   }
-  if (process.platform === 'win32') {
+  if (process.platform === WINDOWS_PLATFORM) {
     return path.join(
       process.env.APPDATA || path.join(os.homedir(), 'AppData', 'Roaming'),
       'caveman'
@@ -39,7 +41,11 @@ function getConfigDir() {
 }
 
 function getConfigPath() {
-  return path.join(getConfigDir(), 'config.json');
+  return path.join(getConfigDir(), CONFIG_FILE_NAME);
+}
+
+function getPosixHomeConfigPath() {
+  return path.join(os.homedir(), '.config', 'caveman', CONFIG_FILE_NAME);
 }
 
 // Walk up from `start` looking for a repo-local caveman config. Returns the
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Issue #723 was addressed on main in ed1fbb7f by making the Windows config path prominent, and the maintainer closed the issue as implemented. Closing this alternative fallback as superseded.

Issue #686, opened 2026-07-10
caveman-compress: missing encoding="utf-8" on file I/O corrupts non-ASCII files on Windows (data loss, no crash)


Jisung-Jeong:
## Summary

On Windows, `skills/caveman-compress/scripts/compress.py` (and `validate.py`, `detect.py`) call `Path.read_text()` / `Path.write_text()` without an explicit `encoding` argument. Python falls back to the OS locale's preferred encoding — on a Korean-locale Windows machine that's `cp949`, not UTF-8.

This silently corrupts any UTF-8 source file containing non-ASCII characters (Korean text, em dashes, arrows, etc.), and in the worst case **truncates the original file to 0 bytes with no usable backup**.

## Root cause

- `filepath.read_text(errors="ignore")` (compress.py:249) decodes UTF-8 bytes as cp949 → mojibake, with `errors="ignore"` silently dropping bytes it can't decode (lossy, not reversible).
- That already-corrupted text is what gets sent to the LLM for compression and what gets written to the backup file.
- `filepath.write_text(compressed)` (compress.py:313) encodes the LLM's compressed output back with the locale codec. `Path.write_text` opens the file in `'w'` mode, which **truncates immediately on open**. If the compressed text contains a character not representable in the locale codec (e.g. an em dash `—`, common in the LLM's own output style), `str.encode()` raises `UnicodeEncodeError` mid-write — after truncation, before any bytes are flushed. Net effect: the original file is left at **0 bytes**, the exception bubbles up to `cli.py`'s generic `except Exception` handler, and the user sees a generic `❌ Error: 'cp949' codec can't encode character ...` with no indication the source file was just destroyed.
- `validate.py`'s `read_file()` has the same missing-encoding bug when comparing original vs. compressed.

## Repro

On Windows with a non-English system locale (e.g. Korean, `cp949`):

```powershell
py -m scripts "path\to\some_korean_language.md"
```

Any `.md` file containing Korean text (or other non-cp1252/cp949-safe punctuation) is at risk. In our case this corrupted a project `CLAUDE.md` (truncated to 0 bytes) and a global `~/.claude/CLAUDE.md` (fully mojibake'd, including its on-disk backup, since the backup is written from the already-mis-decoded text) with no clean way to recover — we only recovered because the original text happened to still be visible earlier in the conversation transcript.

## Fix

[...]

PR #763, opened 2026-07-29
fix(compress): pin utf-8 and newline on file I/O

### LINKED ISSUE #762: caveman-compress rewrites every line ending to CRLF on Windows - survives the #683 encoding fix
# caveman-compress rewrites every line ending to CRLF on Windows — survives the #683 encoding fix

## Scope

This is **not** the encoding bug. That one is #686 (and #652, #655, #533), with #683 and #678 open against it. This is a second, independent defect at the same call sites that none of those touch: after #683 merges, the compressor still rewrites the line endings of every file it processes.

Filing separately so it doesn't get closed along with the encoding work.

## The bug

`Path.write_text()` opens in text mode. On Windows that means `newline=None`, which translates every `\n` to `\r\n` on the way out. An LF-ending source file comes back CRLF, so the next `git diff` reports every line changed.

Measured on a 67-line LF file:

```
before compress   7015 bytes
after  restore    7082 bytes    # +67, one CR per line
```

Adding `encoding="utf-8"` as #683 does has no effect on this — encoding and newline translation are independent parameters.

## Why `newline="\n"` on the writes is not the fix

The obvious patch is wrong in the other direction. Reads go through `read_text()`, which applies universal-newline translation, so `original_text` is **always** LF in memory regardless of what is on disk. Forcing `newline="\n"` on the writes would then flatten genuine CRLF files to LF.

Both sides have to be pinned, and separately, the LLM's output (always LF) has to be matched back to whatever the source actually used.

## Reproduction

No API key needed — stub `call_claude` and run the real `compress_file`:

```python
import sys
from pathlib import Path
sys.path.insert(0, "<plugin>/skills/caveman-compress")
from scripts import compress as C

C.call_claude = lambda prompt: "Plain prose, no heading, no code.\nSecond line.\n"

for label, eol in (("lf", "\n"), ("crlf", "\r\n")):
    src = Path(f"probe_{label}.md")
    raw = f"Plain prose file, no heading, no code.{eol}Second line here.{eol}".encode()
    src.write_bytes(raw)
    C.backup_dir_for(src).joinpath(src.stem + ".original.md").unlink(missing_ok=True)
    C.compress_file(src)
    out = src.read_bytes()
    print(label, "output keeps source EOL:", eol.encode() in out and
          (b"\r\n" in out) == (eol == "\r\n"))
    print(label, "backup byte-identical:  ",
          C.backup_dir_for(src).joinpath(src.stem + ".original.md").read_bytes() == raw)
```

On Windows before the patch: the LF case comes back CRLF, and the backup is not byte-identical to the source. On Linux/macOS both pass, which is why this has gone unnoticed alongside the encoding bug.

## Patch

Applies on top of #683. Tested against both LF and CRLF sources.

```diff
     m = FRONTMATTER_REGEX.match(text)
     if m:
         return m.group(1), m.group(2)
     return "", text
 
+
+def match_line_endings(text: str, reference: str) -> str:
+    """Rewrite text's line endings to match the reference document's.
+
+    File I/O below pins ``newline=""`` so Python performs no translation of
+    its own. The LLM always returns LF, so a CRLF source still needs its
+    endings restored explicitly.
+    """
+    crlf = reference.count("\r\n")
+    lf = reference.count("\n") - crlf
+    normalized = text.replace("\r\n", "\n")
+    if crlf > lf:
+        return normalized.replace("\n", "\r\n")
+    return normalized
+
+
+def read_exact(path: Path) -> str:
+    """Read text as UTF-8 with no newline translation.
+
+    ``Path.read_text`` only grew a ``newline`` parameter in 3.13, so go through
+    ``open`` to keep this working on older interpreters.
+    """
+    with path.open(encoding="utf-8", errors="ignore", newline="") as fh:
+        return fh.read()
+
+
 # Filenames and paths that almost certainly hold secrets or PII.
```

```diff
-    original_text = filepath.read_text(encoding="utf-8", errors="ignore")
+    original_text = read_exact(filepath)
```

```diff
-    compressed = frontmatter + compressed_body
+    compressed = match_line_endings(frontmatter + compressed_body, original_text)
```

```diff
     backup_path.write_text(original_text, encoding="utf-8", newline="")
-    backup_readback = backup_path.read_text(encoding="utf-8", errors="ignore")
+    backup_readback = read_exact(backup_path)
```

```diff
-    filepath.write_text(compressed, encoding="utf-8")
+    filepath.write_text(compressed, encoding="utf-8", newline="")
```

```diff
             # Restore original on failure
-            filepath.write_text(original_text, encoding="utf-8")
+            filepath.write_text(original_text, encoding="utf-8", newline="")
```

```diff
-        compressed = call_claude(
-            build_fix_prompt(original_text, compressed, result.errors)
-        )
-        filepath.write_text(compressed, encoding="utf-8")
+        compressed = match_line_endings(
+            call_claude(build_fix_prompt(original_text, compressed, result.errors)),
+            original_text,
+        )
+        filepath.write_text(compressed, encoding="utf-8", newline="")
```

`validate.py` should **not** get `newline=""`. The validator needs universal-newline normalization so a CRLF original and an LF candidate still compare equal — pinning it there would make `validate_code_blocks` fail on every CRLF file.

### Results after the patch

```
[lf]   output keeps '\n', zero CR bytes, backup byte-identical (118 == 118)
[crlf] output keeps '\r\n',              backup byte-identical (120 == 120)
```

## Two notes on #683 while you're in here

**It misses `detect.py`.** #686's body lists the extensionless-file content sniff as a call site needing the fix, but #683 only patches `compress.py` and `validate.py`. `detect.py:93` still has a bare `read_text(errors="ignore")`:

```diff
         try:
-            text = filepath.read_text(errors="ignore")
+            text = filepath.read_text(encoding="utf-8", errors="ignore")
```

It classifies rather than writes, so it cannot corrupt a file — but on a non-UTF-8 locale it can misclassify a prose file as code and silently skip it.

**Watch the Python floor if anyone adds `newline=` to a read.** `Path.read_text()` only accepts `newline=` on 3.13+; `Path.write_text()` has had it since 3.10. The repo declares no minimum Python version and the skill docs just say `python3 -m scripts`, so a `read_text(..., newline="")` would `TypeError` for anyone on 3.10–3.12. That is why the patch above routes the reads through `open()` instead.

## Environment

- Windows 11 Pro 26200
- Python 3.14.6
- caveman plugin `0d95a81d35a9`
- No `ANTHROPIC_API_KEY`, so the `claude --print` CLI path is used

francojeferson:
Fixes #762. Also covers #686 (and #652, #655, #533).

## What

`Path.read_text` / `Path.write_text` default to the locale encoding **and** to CRLF translation on Windows. Both defaults mutate the user's file.

**Encoding.** Non-ASCII output goes out in the locale codec. An em dash becomes a bare `0x97` under cp1252, so the file stops being valid UTF-8. `validate.py` then read it back through the same default, saw a correct em dash, and printed `Validation passed` — the round-trip is self-consistent and wrong, so the corruption ships silently.

**Line endings.** Every processed file comes back CRLF. Measured on a real run: a 67-line LF file restored from backup at +67 bytes, one CR per line. `git diff` then reports every line changed.

## Overlap with #683

#683 already fixes the encoding half, and this PR contains that same change — the two touch identical call sites, so splitting them would just conflict. What this adds on top:

- `detect.py:93`, which #686's body lists as needing the fix but #683 misses
- the newline half, which #683 does not touch; after it merges the CRLF rewrite is still there
- regression tests

Happy to rebase down to newline-only if #683 lands first. No claim on that work intended.

## Notes

- Reads go through `open()` rather than `read_text(newline=...)`. That parameter is 3.13+ and `CLAUDE.md` states the skill supports Python 3.10+.
- `validate.py` deliberately does **not** get `newline=""`. The validator needs universal-newline normalization so a CRLF original and an LF candidate still compare equal — pinning it there would make `validate_code_blocks` fail on every CRLF file.
- The no-op comparison needed normalizing too. The body now carries the source's CRLF while Claude always answers LF, so `compressed_body.strip() == body.strip()` would never match and the "output identical to input" guard would stop firing on CRLF files entirely. The existing `test_identical_compressed_output_does_not_touch_disk` caught this — it is green again.

## Tests

New `tests/test_compress_encoding.py`, 13 cases: `match_line_endings` unit tests, `read_exact` round-trips, and end-to-end through `compress_file` with `call_claude` stubbed, so no API key is needed.

```
python3 -m unittest tests.test_compress_encoding   # Ran 13 tests, OK
python3 -m unittest tests.test_compress_safety     # Ran 5 tests, OK
```

`tests.test_hooks` reports 3 errors and `npm test` 2 failures on my Windows box. Both are identical on a clean checkout of `main` and unrelated to this change.

CI will resync `plugins/caveman/skills/caveman-compress/scripts/` on merge.

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -39,6 +39,38 @@ def split_frontmatter(text: str):
         return m.group(1), m.group(2)
     return "", text
 
+
+def read_exact(path: Path) -> str:
+    """Read text as UTF-8 with no newline translation.
+
+    ``Path.read_text`` grew a ``newline`` parameter only in 3.13 and the skill
+    supports 3.10+, so go through ``open`` to keep older interpreters working.
+    """
+    with path.open(encoding="utf-8", errors="ignore", newline="") as fh:
+        return fh.read()
+
+
+def _lf(text: str) -> str:
+    """Normalize to LF so CRLF and LF copies of the same text compare equal."""
+    return text.replace("\r\n", "\n")
+
+
+def match_line_endings(text: str, reference: str) -> str:
+    """Rewrite text's line endings to match the reference document's.
+
+    Every read and write here pins ``newline=""`` so Python performs no
+    translation of its own — otherwise ``write_text`` emits CRLF on Windows and
+    silently rewrites every line of the user's file. Claude always returns LF,
+    so a CRLF source still needs its endings restored explicitly.
+    """
+    crlf = reference.count("\r\n")
+    lf = reference.count("\n") - crlf
+    normalized = _lf(text)
+    if crlf > lf:
+        return normalized.replace("\n", "\r\n")
+    return normalized
+
+
 # Filenames and paths that almost certainly hold secrets or PII. Compressing
 # them ships raw bytes to the Anthropic API — a third-party data boundary that
 # developers on sensitive codebases cannot cross. detect.py already skips .env
@@ -246,7 +278,7 @@ def compress_file(filepath: Path) -> bool:
         print("Skipping (not natural language)")
         return False
 
-    original_text = filepath.read_text(errors="ignore")
+    original_text = read_exact(filepath)
     # Store backup outside the source directory so skill auto-loaders don't
     # re-ingest the `.original.md` copy as a live file. Mirror the source's
     # parent-dir name + stem under a platform-aware base to reduce collisions.
@@ -287,21 +319,25 @@ def compress_file(filepath: Path) -> bool:
 
     # Compare the BODY (not the whole file) — frontmatter is preserved verbatim
     # and would never change, so identity must be judged on the compressible part.
-    if compressed_body.strip() == body.strip():
+    # Normalize line endings first: the body keeps the source file's CRLF while
+    # Claude always answers in LF, so a raw compare would never match on a CRLF
+    # file and the no-op guard below would stop firing entirely.
+    if _lf(compressed_body).strip() == _lf(body).strip():
         print("❌ Compression aborted: output is identical to input.")
         print("   Likely causes: Claude refused, returned the prompt verbatim, or the file is")
         print("   already in caveman form. Original file is untouched (no backup created).")
         return False
 
-    # Reassemble: frontmatter (verbatim) + compressed body
-    compressed = frontmatter + compressed_body
+    # Reassemble: frontmatter (verbatim) + compressed body, then restore the
+    # source file's line endings — Claude always answers in LF.
+    compressed = match_line_endings(frontmatter + compressed_body, original_text)
 
     # Save original as backup, then verify the backup readback before
     # touching the input file. If the filesystem dropped bytes (encoding,
     # antivirus, disk full), unlink the bad backup and abort instead of
     # leaving the user with a corrupt backup + compressed primary.
-    backup_path.write_text(original_text)
-    backup_readback = backup_path.read_text(errors="ignore")
+    backup_path.write_text(original_text, encoding="utf-8", newline="")
+    backup_readback = read_exact(backup_path)
     if backup_readback != original_text:
         print(f"❌ Backup write verification failed: {backup_path}")
         print("   In-memory original differs from on-disk backup. Aborting before touching the input file.")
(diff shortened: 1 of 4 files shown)


JuliusBrussee:
Thanks @francojeferson — UTF-8 and newline preservation are now enforced in the compressor. Closing as implemented.

PR #767, opened 2026-07-30
fix(caveman-compress): UTF-8 file I/O and restore on retry crash

### LINKED ISSUE #686: caveman-compress: missing encoding="utf-8" on file I/O corrupts non-ASCII files on Windows (data loss, no crash)

### LINKED ISSUE #766: caveman-compress: Windows encoding bug corrupts/loses file content on crash during fix-retry step
## Summary

On Windows, `scripts/compress.py` (caveman-compress skill) silently corrupts non-ASCII characters (arrows `→`, emojis `⚠️✅🚫`, accented characters) when writing the compressed file and its backup, and can leave the target file **empty/corrupted with no restore** if the fix-retry step throws an exception.

## Root cause

All `Path.read_text()` / `Path.write_text()` calls in `compress.py` omit `encoding="utf-8"`. On Windows, this defaults to the locale-preferred codepage (cp1252 in my case) instead of UTF-8. Any character outside cp1252's range gets mangled into `�` on write, or raises `UnicodeEncodeError` depending on the code path.

Affected calls (line numbers as of commit shipped in plugin version `0d95a81d35a9`):
- `original_text = filepath.read_text(errors="ignore")`
- `backup_readback = backup_path.read_text(errors="ignore")`
- `filepath.write_text(compressed)` (first pass)
- `filepath.write_text(original_text)` (restore-on-final-failure path)
- `filepath.write_text(compressed)` (inside the fix/retry loop)

## Second, more serious bug: crash during fix-retry leaves file corrupted, no restore

In `compress_file()`, the retry loop only restores `original_text` when validation fails on the **last** attempt (`attempt == MAX_RETRIES - 1`). If `call_claude(build_fix_prompt(...))` itself **raises** (e.g. the `UnicodeEncodeError` above, or any other exception from the `claude --print` subprocess fallback), the exception propagates straight out of `compress_file()` and is caught by the generic `except Exception as e` in `cli.py::main()`, which just prints the error and exits — **it never restores `original_text`**. The live target file is left containing whatever was written at the earlier `filepath.write_text(compressed)` call (first pass), i.e. a validation-failing, possibly mojibake-corrupted compressed version instead of the original.

## Repro (Windows 11, PowerShell/Git-Bash, Python 3.12.10, no `ANTHROPIC_API_KEY` set → falls back to `claude --print`)

1. Run `python -m scripts <path-to-markdown-with-arrows-or-emoji>` on a file containing `→` or box-drawing characters (e.g. a tree diagram) or emoji (`⚠️`, `✅`, `🚫`).
2. First compression pass succeeds, writes to the real file.
3. Validation fails (e.g. "Code blocks not preserved exactly").
4. Fix/retry step calls Claude again; something in that path raises `UnicodeEncodeError: 'charmap' codec can't encode character '→' in position ...: character maps to <undefined>`.
5. Script exits with the error, but the **live file has already been overwritten** and is left in a broken state (in one run, a stale backup this produced was found containing `�` in place of every `→` and box-drawing character).
6. On a second run, the corrupted/empty file is read as `original_text`, and the script correctly refuses to compress an empty file — but the damage to the previous state is already done, with no automatic recovery.

I only recovered because the target file happened to be tracked in git (`git checkout -- <file>`). Without version control this would be silent, permanent data loss.

## Suggested fix

- Pass `encoding="utf-8"` explicitly on every `read_text()` / `write_text()` call in `compress.py`.
- Wrap the fix/retry `call_claude(...)` call in its own `try/except` that restores `original_text` and removes the backup before re-raising or returning `False`, so any crash mid-retry can never leave the live file in a half-compressed or corrupted state.

Happy to send a PR with both fixes if useful — I already patched a local copy to confirm this resolves it (compression now round-trips cleanly with 0 corrupted characters and passes validation).

## Environment

- OS: Windows 11 Pro
- Python: 3.12.10 (installed via winget, no `ANTHROPIC_API_KEY` set — using `claude --print` fallback)
- Shell: Git Bash / PowerShell
- Plugin version installed: cache dir `0d95a81d35a9`

SomSamantray:
## Summary

On Windows with a non-UTF-8 system locale, `caveman-compress` could mojibake or destroy markdown containing Korean text, arrows, emoji, or other non-ASCII characters. Worse, if validation failed and the fix-retry step crashed, the live file could be left in a corrupted half-compressed state with no automatic restore.

This change makes all compress-skill file I/O explicitly UTF-8, writes the target file atomically so a failed encode cannot truncate it to zero bytes, and restores the verified original whenever fix-retry raises before the error reaches the CLI.

Fixes #686
Fixes #766

## Validation

- `python3 -m unittest tests.test_compress_safety tests.test_detect tests.test_compress_encoding -v` — 15 tests passed
- `python3 tests/verify_repo.py` — all local verification checks passed
- New regression coverage: UTF-8 round-trip under simulated `cp1252` locale preference, restore-on-exception during fix-retry, atomic write preserving target on failure

### DIFF
--- a/docs/plans/2026-07-30-001-fix-caveman-compress-windows-encoding-plan.md
+++ b/docs/plans/2026-07-30-001-fix-caveman-compress-windows-encoding-plan.md
@@ -1,0 +1,339 @@
+---
+title: "fix: caveman-compress Windows UTF-8 I/O and crash-safe retry restore"
+date: 2026-07-30
+type: fix
+artifact_contract: ce-unified-plan/v1
+artifact_readiness: implementation-ready
+product_contract_source: ce-plan-bootstrap
+execution: code
+origin_issues:
+  - https://github.com/JuliusBrussee/caveman/issues/686
+  - https://github.com/JuliusBrussee/caveman/issues/766
+target_repo: JuliusBrussee/caveman
+---
+
+## Goal Capsule
+
+Fix **caveman-compress** so UTF-8 markdown (Korean, emoji, arrows, box-drawing) round-trips correctly on Windows locale codepages (`cp1252`, `cp949`, etc.) and so any exception during the validation/fix-retry loop **never** leaves the live target file in a corrupted or truncated state.
+
+**Authority:** Issues [#686](https://github.com/JuliusBrussee/caveman/issues/686) and [#766](https://github.com/JuliusBrussee/caveman/issues/766).
+
+**Stop conditions:** All verification commands pass; new regression tests cover encoding and crash-restore paths; only `skills/caveman-compress/` sources are edited (CI mirrors to `plugins/`).
+
+---
+
+## Product Contract
+
+### Summary
+
+On Windows, `caveman-compress` reads and writes markdown using Python's locale-default encoding instead of UTF-8. That silently corrupts non-ASCII content and can truncate files to zero bytes when `write_text` opens the target for write and encoding fails mid-flight. A second failure mode: if the fix-retry `call_claude` step raises, the generic CLI handler exits without restoring `original_text`, leaving the user with a half-compressed file and no recovery path except git.
+
+This plan closes both issue families in one focused change set: explicit UTF-8 I/O everywhere in the compress skill scripts, atomic target writes, and guaranteed restore on any post-backup failure.
+
+### Requirements
+
+- R1. Every `Path.read_text()` / `Path.write_text()` in `skills/caveman-compress/scripts/compress.py`, `validate.py`, and `detect.py` must use UTF-8 explicitly.
+- R2. File reads must not use `errors="ignore"` — invalid UTF-8 in a source file must fail loudly instead of silently dropping bytes (see KTD1).
+- R3. Writes to the **live target file** (`filepath`) must be atomic (write temp + `os.replace`) so an encode/write failure cannot truncate the original to 0 bytes.
+- R4. After a verified backup exists and the target has been mutated, **any** exception in the validation/fix-retry loop (including `call_claude` failures) must restore `original_text` to the target and remove the backup before the error propagates to `cli.py`.
+- R5. Existing pre-backup safety guards (empty input refusal, identical-output abort, backup readback verification) must remain unchanged in behavior.
+- R6. Regression tests must prove UTF-8 round-trip under simulated non-UTF-8 locale defaults and prove restore-on-exception during fix-retry.
+- R7. Edits apply only to the source-of-truth tree under `skills/caveman-compress/`; `plugins/caveman/skills/caveman-compress/` is CI-rebuilt.
+
+### Actors
+
+- A1. **Windows developer** running `python -m scripts <file.md>` or the installed caveman-compress skill on UTF-8 markdown containing non-ASCII text.
+- A2. **Maintainer** running `tests/verify_repo.py` and unit tests in CI.
+
+### Key Flows
+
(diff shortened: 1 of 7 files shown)


JuliusBrussee:
Thanks @SomSamantray — current compression is UTF-8 strict, atomic, and recovery-safe. Closing this superseded branch.

PR #768, opened 2026-07-30
docs(caveman-compress): document out-of-tree backup location and agent collision safety

twall:
## Summary

- `SKILL.md` still documented backups as `FILE.original.md` written in-place, but `compress.py` already writes them out-of-tree via `backup_dir_for()` (`~/.local/share/caveman-compress/backups/` on Linux/macOS, `%LOCALAPPDATA%\caveman-compress\backups\` on Windows)
- Adds explicit note that compressing Claude Code agent definitions (`~/.claude/agents/*.md`) and skill files is safe — since the backup never lands next to the source, there's no `name:` frontmatter collision that would cause nondeterministic agent loading
- Both `skills/` and `plugins/caveman/skills/` copies updated

## Context

When `caveman-compress` wrote backups in-place, compressing an agent definition like `graylog-log-analyst.md` produced `graylog-log-analyst.original.md` in the same directory. Claude Code picked up both files as agent definitions (same `name:` in frontmatter), loading one nondeterministically. The out-of-tree backup location already in `compress.py` fixes this — the docs just didn't say so.

### DIFF
--- a/plugins/caveman/skills/caveman-compress/SKILL.md
+++ b/plugins/caveman/skills/caveman-compress/SKILL.md
@@ -3,7 +3,7 @@ name: caveman-compress
 description: >
   Compress natural language memory files (CLAUDE.md, todos, preferences) into caveman format
   to save input tokens. Preserves all technical substance, code, URLs, and structure.
-  Compressed version overwrites the original file. Human-readable backup saved as FILE.original.md.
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Thanks @twall — out-of-tree backup and collision-safety behavior is documented now. Closing as implemented.

PR #695, opened 2026-07-11
ci: run the test suite in CI + harden the sync workflow

NX1X:
Adds a CI workflow that runs the existing test suite on every PR, and hardens the one
existing workflow.

- `ci.yml`: runs the full documented suite on Node 18/20/22 — `npm test` (installer
  unit/e2e), the standalone `tests/test_*.js` runners, and the Python `unittest`
  suite — plus `actionlint`, `zizmor`, and `shellcheck`.
- Commits `package-lock.json` for the root package and `caveman-shrink` (both
  zero-dep today) so CI can `npm ci` reproducibly.
- `sync-skill.yml`: pins `actions/checkout` to a commit SHA, adds a top-level
  `permissions: contents: read` default, and a `timeout-minutes`. Sync behavior is
  unchanged.

The tests already existed but weren't run on PRs. All actions are pinned to their
latest release SHA. Suite passes locally: `npm test` 117, standalone node 7/7,
python 52.

PR #774, opened 2026-07-31
fix(shrink): stop corrupting multi-byte characters at chunk boundaries

Dhevenddra:
## What

`makeLineBuffer` decoded each stdout chunk on its own with `Buffer#toString('utf8')`. Node picks those chunk boundaries by byte count, not by character, so any multi-byte character split across two chunks gets destroyed — the trailing bytes decode to U+FFFD in one chunk and the leading bytes to U+FFFD in the next.

The damaged line is still valid JSON, so nothing downstream errors. The model just receives a mangled tool description. Whether a given description survives depends on where the boundaries happen to land, which makes it look intermittent.

No issue filed for this one — found it while reading the proxy.

## Repro

Feeding a `tools/list` response through the current framing, split one byte into the 3-byte `读`:

```
chunks: 71 and 39 bytes
parsed OK  : true
description: ���取文件 — reads a file 🎉
```

End to end through the proxy against a server that writes its response a byte at a time:

```
before   "Sends alert ��� make channel is valid ✅"
after    "Sends alert — make channel is valid ✅"
```

Compression still applies as before — `读取文件的内容。The function will basically just return the contents.` comes out as `读取文件的内容。function will return contents.`

## Fix

`StringDecoder` holds an incomplete trailing sequence until its continuation bytes arrive. That is exactly the case here.

Framing moves to `line-buffer.js` for the same reason `spawn-options.js` was split out — `index.js` spawns the upstream at require time, so nothing in it can be unit-tested in place.

## Tests

Four framing tests. Two of them fail on `main`:

```
✗ reassembles a multi-byte character split across chunks
  replacement character in reassembled line: ...\"description\":\"���取文件 — reads a file 🎉\"...
✗ survives a byte-at-a-time stream
  '��������� ��� caf��' == '日本語 ✅ café'
```

The other two cover the framing behaviour I did not want to regress while moving it: blank-line dropping, and withholding a partial line until its newline arrives.

`line-buffer.js` is added to `files` in the package manifest, which the existing #597 packaging test checks.

## Note on running these

`npm test` only runs `tests/installer/*.test.mjs`, so `tests/test_mcp_shrink.js` is not picked up by it. Ran directly:

```
node tests/test_mcp_shrink.js
22 passed, 0 failed
```

Worth mentioning since CONTRIBUTING says CI runs the suite on every PR — as far as I can tell the only workflow is `sync-skill.yml`, and nothing runs tests. #695 looks like it is addressing that.

### DIFF
--- a/src/mcp-servers/caveman-shrink/index.js
+++ b/src/mcp-servers/caveman-shrink/index.js
@@ -57,18 +57,8 @@ upstream.on('exit', (code, signal) => {
 // JSON-RPC framing over stdio: messages are separated by newlines (the
 // MCP stdio transport uses LSP-like content but most servers emit one JSON
 // object per line). We line-buffer in both directions and parse opportunistically.
-function makeLineBuffer(onLine) {
-  let buf = '';
-  return chunk => {
-    buf += chunk.toString('utf8');
-    let nl;
-    while ((nl = buf.indexOf('\n')) !== -1) {
-      const line = buf.slice(0, nl);
-      buf = buf.slice(nl + 1);
-      if (line.trim()) onLine(line);
-    }
-  };
-}
+// Lives in line-buffer.js so the chunk-boundary handling is unit-testable.
+const { makeLineBuffer } = require('./line-buffer');
 
 function transformResponse(msg) {
   // Compress description fields on list-style responses. Match by method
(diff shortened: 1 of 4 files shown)


Dhevenddra:
The 2.0.0 rewrite fixed this independently, so there is nothing left here to merge. `makeLineBuffer` in the new `index.js` runs each chunk through a `StringDecoder` and flushes the retained bytes in `end()`, which is the same approach this branch took. Closing rather than leaving it open against code that no longer exists.

For anyone who finds this later, the defect was `buf += chunk.toString('utf8')` in the old line buffer: a multi-byte character split across a chunk boundary decoded to replacement characters before the line was ever parsed, so the corruption was baked in before compression ran.

I have rebased the sibling PR (#775, signal forwarding) onto the new main. That one is still a live bug.

Issue #528, opened 2026-06-14
Security Scan Flags “Caution” Verdict for the *caveman* Skill (High‑risk Exfiltration & Medium‑risk Path Traversal)


Icarus-B4:
**Title:** Security Scan Flags “Caution” Verdict for the *caveman* Skill (High‑risk Exfiltration & Medium‑risk Path Traversal)

---

**Description**

When attempting to install the *caveman* skill via the Hermes CLI (`hermes skills install https://github.com/JuliusBrussee/caveman`), the built‑in security scanner blocks the operation:

```
Scan: caveman (caveman/community)  Verdict: CAUTION
  HIGH     exfiltration   benchmarks/run.py:23
  MEDIUM   traversal      CONTRIBUTING.md:18
Decision: BLOCKED — Blocked (community source + caution verdict, 2 findings).
Use --force to override.
```

The two findings are:

| Severity | Type          | File & Line                        | Brief Description |
|----------|---------------|------------------------------------|-------------------|
| **HIGH** | Exfiltration  | `benchmarks/run.py:23`             | Calls `os.environ.setdefault(key.strip(), value.strip())`, which may read environment variables (including API keys, tokens, passwords) and could unintentionally expose them to external services or logs. |
| **MEDIUM**| Path Traversal| `CONTRIBUTING.md:18` (text reference) | Mentions relative path `../../issues?q=label`, indicating the code may construct file paths from user‑controlled inputs, potentially allowing directory traversal and unintended file reads/writes. |

Because the source is community‑maintained and these issues are flagged, Hermes automatically blocks the installation unless the user explicitly overrides with `--force`.

---

**Impact**

- **Potential data leakage**: If the skill runs in a profile that contains sensitive environment variables (e.g., API keys, `.env` values), the exfiltration code could leak those values to remote endpoints or logs.
- **File‑system safety**: Path‑traversal patterns could be exploited to read or overwrite files outside the intended directory, especially if the skill later adds dynamic file handling based on user input.
- **User friction**: The block forces users to either manually inspect the code or use `--force`, which defeats the purpose of the security safeguard.

---

**Requested Action**

1. **Audit & Harden the Code**
   - Review `benchmarks/run.py` for any unintended exposure of environment variables. If reading env vars is required, explicitly whitelist only non‑secret variables or mask values before any outbound communication.
[...]

Issue #383, opened 2026-05-13
Does it contain malware?


JeroNumor:
**What happen**
<img width="937" height="312" alt="Image" src="https://github.com/user-attachments/assets/119bc31f-3ffa-4137-9d6b-0d033588f4e7" />

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**
`powershell -NoProfile -ExecutionPolicy Bypass -Command "iex (irm https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.ps1)"`

ARKye03:
WIndows itself is the malware, besides, read the script https://github.com/JuliusBrussee/caveman/blob/main/install.ps1, check by yourself if it is malware or not.

JuliusBrussee:
Thanks for reporting this. This was addressed on main in [`ec4f6649`](https://github.com/JuliusBrussee/caveman/commit/ec4f6649) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this as implemented. If the current release still reproduces it, reply with your Caveman version, platform, command, and a minimal reproduction and I’ll reopen it.

PR #775, opened 2026-07-31
fix(shrink): forward termination signals so the upstream is not orphaned

### LINKED ISSUE #742: caveman-shrink never forwards SIGTERM/SIGINT to its spawned upstream server, orphaning it
**What happen**

indefinitely.

`exit`/`error` events:

```js

```

**Expected**

**Repro**

```
process.stdin.resume();
EOF

WRAPPER_PID=$!
sleep 1
kill -TERM $WRAPPER_PID
sleep 1
ps -o pid,ppid,stat,args
```

having received a signal.

**Platform**

- [ ] Claude Code
- [ ] Codex

**Version / install method**

installer).

Dhevenddra:
Closes #742.

## What

caveman-shrink spawns the upstream MCP server, so nothing else will clean it up. It registered handlers for the child's own `error` and `exit` events but none for its own termination, so a host that stops the proxy with a signal — ending a session, disabling a tool, reloading config — left the server running, reparented to PID 1, still holding its stdio.

## Fix

Forward `SIGTERM`, `SIGINT` and `SIGHUP` to the child, once, with a 2s grace period before giving up on a child that ignores them. `SIGKILL` is not registered because it cannot be trapped, and a host sending it has already decided not to wait.

Teardown lives in `shutdown.js` so it is unit-testable without running the CLI entry point, mirroring `spawn-options.js`.

## Windows needs a different shape

`getSpawnOptions` sets `shell: true` on win32 so `npx` and other `.cmd` shims resolve (#387). That puts `cmd.exe` between us and the real server, so `child.kill()` reaps the wrapper and orphans the server. Teardown goes through `taskkill /T` there instead.

Windows also has no real signals — `process.kill(pid, 'SIGTERM')` terminates the target without running its handlers — so the signal path is effectively POSIX-only. Windows teardown is driven by stdin close, which is the graceful stop MCP hosts use anyway.

## One thing to weigh before merging

Shipping a `taskkill /T /F` call in the published `caveman-shrink` package is worth a conscious decision given #528 and #383. Process-tree termination is a signal some scanners weight, and caveman is already fielding "is this malware" questions.

I think it is still the right call — `taskkill /T` is the standard way to reap a Windows process tree, and the alternative is leaving user processes orphaned — but I would rather raise it than have it turn up in a scanner report later. Happy to drop the Windows branch and leave win32 on stdin-close teardown alone if you would prefer the smaller surface.

## Tests

Five tests covering the teardown logic, with an injected `spawn` so nothing real is killed during the run:

- forwards the signal to the child on POSIX
- uses `taskkill /pid N /T /F` on win32 and does not also signal directly
- no-ops on an upstream that already exited (`exitCode` set, `signalCode` set, or null child)
- registers a handler per forwarded signal, and never registers SIGKILL
- shutdown is idempotent — repeated signals kill once

```
node tests/test_mcp_shrink.js
23 passed, 0 failed
```

## What I did not verify myself

The POSIX orphan behaviour is as reported in #742 and confirmed there in `node:20-alpine`; I did not re-run that container test, so the POSIX path here rests on the reporter's repro plus the unit tests.

I also did not run a live Windows process-tree test. My first attempt at one tripped Defender — writing a probe script to temp, enumerating processes via `wmic`, then force-killing a tree is a reasonable thing for an AV to dislike, and it did not seem worth fighting for. The `taskkill` invocation is covered by the unit test with an injected spawn.

## Unrelated thing I noticed

`shell: true` emits `DEP0190` on current Node ("Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated"). Out of scope here, and it interacts with #387 and #717, so leaving it alone.

### DIFF
--- a/src/mcp-servers/caveman-shrink/index.js
+++ b/src/mcp-servers/caveman-shrink/index.js
@@ -43,6 +43,7 @@ const fields = (process.env.CAVEMAN_SHRINK_FIELDS || 'description')
   .split(',').map(s => s.trim()).filter(Boolean);
 
 const { getSpawnInvocation, getSpawnOptions } = require('./spawn-options');
+const { installShutdownHandlers, killUpstream } = require('./shutdown');
 
 let invocation;
 try {
@@ -52,6 +53,10 @@ try {
   process.exit(1);
 }
 const upstream = spawn(invocation.command, invocation.args, getSpawnOptions());
+
+// We spawned the upstream, so we own tearing it down. Without this a host that
+// stops us with a signal leaves the server orphaned with its stdio open (#742).
+installShutdownHandlers(upstream);
 
 let spawnFailed = false;
 upstream.on('error', err => {
@@ -170,6 +175,11 @@ function forwardInput(chunk) {
 }
 function endInput() {
   if (upstream.stdin.writable && !upstream.stdin.destroyed) upstream.stdin.end();
+  // Closing our stdin is how a host asks for a graceful stop, and on Windows it
+  // is the only teardown that runs since there are no real signals there. A
+  // well-behaved server exits on the EOF forwarded above; this reaps the ones
+  // that don't. unref() so it never delays an upstream that exited promptly.
+  setTimeout(() => killUpstream(upstream, 'SIGTERM'), 2000).unref();
 }
 upstream.stdin.on('error', err => {
   if (err.code !== 'EPIPE' && !spawnFailed) {
--- a/src/mcp-servers/caveman-shrink/package.json
+++ b/src/mcp-servers/caveman-shrink/package.json
@@ -26,6 +26,7 @@
     "index.js",
     "compress.js",
     "spawn-options.js",
+    "shutdown.js",
     "README.md"
   ]
 }
--- a/src/mcp-servers/caveman-shrink/shutdown.js
+++ b/src/mcp-servers/caveman-shrink/shutdown.js
@@ -1,0 +1,78 @@
+// Upstream teardown for the caveman-shrink proxy.
+//
+// The proxy spawns the upstream MCP server, so nothing else will clean it up.
+// Without this, a host that stops caveman-shrink with a signal — ending a
+// session, disabling a tool, reloading config — leaves the upstream running,
+// reparented away from us, still holding its stdio open (#742).
+//
+// Platform notes:
+//
+//   POSIX  — the host's SIGTERM/SIGINT reaches us and child.kill() reaches the
+//            upstream, so forwarding is the whole fix.
+//   win32  — there are no real signals. Node emits SIGINT/SIGBREAK from console
+//            events, but a host stopping a background proxy will not produce
+//            them, so teardown there runs through the stdin-EOF path instead.
+//            getSpawnInvocation() resolves .cmd shims to their Node target, so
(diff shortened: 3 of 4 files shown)


Dhevenddra:
Rebased onto main now that 2.0.0 has landed. The rewrite changed the shape of the fix, so this is a re-port rather than a straight rebase:

- `getSpawnOptions()` no longer sets `shell: true`, and `getSpawnInvocation()` resolves `.cmd` shims to their Node target, so our direct child is the server itself. The earlier version of this patch went through `taskkill /T` specifically because the direct child was `cmd.exe` on Windows. That is no longer the case, so `child.kill()` reaches the server directly and the whole Windows branch is gone. `shutdown.js` is about 20 lines shorter for it.
- The stdin-EOF backstop now lives inside `endInput()`, so the `removeListener('end', endInput)` in your `close` handler tears it down along with the rest instead of leaving a stray listener behind.
- Added the `Signed-off-by` trailer I missed the first time round.

The bug still reproduces on current main. Wrapping a server that deliberately ignores stdin EOF, then closing the proxy's stdin the way a host would:

```
main:    upstream alive 4s after stdin EOF: true    (orphaned)
patched: upstream alive 4s after stdin EOF: false   (reaped)
```

`node tests/test_mcp_shrink.js` gives 24 passed / 0 failed, up from 18 on main. The six new tests cover signal forwarding, idempotency under a second signal, skipping an already-exited child, and the 143/130 exit codes when the upstream ignores the signal.

JuliusBrussee:
Reviewed, and credited as a co-author on the #993 commit that fixes #742 — but not cherry-picked directly, for a mechanical reason rather than a quality one.

The branch sits on a months-old base, and its merge commit reverts large parts of current main (workflows, README, CLAUDE.md), so applying it here would have been destructive. #985 landed the same core fix against current main and applied cleanly.

Your PR was the more thorough of the two in scope: it identified the SIGHUP gap, which #985 did not, and that is fixed in #993 as a result. The 2s grace period for a child that ignores the signal, and moving teardown into a testable `shutdown.js`, are still worth having and would rebase cleanly onto current main as a follow-up if you are up for it.

---

Dhevenddra:
Thanks for the credit on #993, and for spelling out the mechanics rather than just closing this.

You were right about the base, and I checked it before rebuilding rather than taking my own word for it: against current `main` this branch differs in 553 files, because it carries a `Merge origin/main into main` commit made from an old base. Merging it would have reverted the workflows, README and CLAUDE.md exactly as you said. Nothing worth salvaging from the branch itself.

The follow-up is #995, written against current `main` from scratch instead of rebased. It has the two pieces you named: the 2s grace period with SIGKILL escalation for a child that ignores the signal, and teardown moved into `shutdown.js`. Timers are injected there, so the escalation tests are synchronous rather than sleeping on a real clock. 26 tests passing, up from 18, and every new one was run against a variant with the behaviour removed so I could show it actually fails without the fix.

One thing I found while building the end-to-end check and deliberately kept out of #995, since it is not what you asked for: an upstream that ignores stdin EOF is still orphaned on `main`, because `endInput()` ends the child's stdin and nothing follows up. That is the Windows teardown path specifically, where there are no real signals to fall back on. Happy to send it as its own PR if you want it.

Closing this one.

PR #777, opened 2026-08-01
fix(compress): reject compression output that is not smaller than input

### LINKED ISSUE #776: caveman-compress: a "compression" that EXPANDS the file passes validation and is committed
**What happen**

```python
```

**Expected**

**Before/after example**

```
         Validation attempt 1
         Validation passed
```

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**

---

divya0795:
Fixes #776

## What

`validate()` checks structure only — headings, code blocks, URLs, paths, bullets, inline code. None of it compares length, and `compress_file()` rejected only an exactly-identical body. A rewrite that keeps every heading and code block but is **longer** passed all checks, overwrote the original, and was reported as a successful compression.

This adds a length gate right beside the existing identical-output guard, before any backup or write, so the original is never touched.

```python
if len(compressed_body) >= len(body):
    print("❌ Compression aborted: output is not smaller than input "
          f"({len(compressed_body)} >= {len(body)} chars).")
    print("   Original file is untouched (no backup created).")
    return False
```

Comparing bodies is the same as comparing files here — `compressed = frontmatter + compressed_body` and the frontmatter is byte-identical on both sides.

## Why here and not in `validate.py`

An error raised inside `validate()` feeds the `MAX_RETRIES` loop, which spends a Claude fix-call that `build_fix_prompt` explicitly cannot satisfy ("DO NOT recompress or rephrase the file") before restoring the original. Burning an API call to reject an expansion seemed wrong in a tool whose point is saving tokens. Aborting at this point creates no backup and never touches the input.

Trade-off, so it is on the record: a standalone `python validate.py <orig> <comp>` still will not report an expansion. `validate()` stays structure-only.

## Test

`tests/test_compress_safety.py` gains `test_expanded_compressed_output_does_not_touch_disk`, next to the existing identical-output test and using the same `XDG_DATA_HOME`/`LOCALAPPDATA` isolation as `test_real_compression_writes_backup_and_target`.

On pre-fix code it fails exactly as the bug describes — `compress_file()` prints `Validation passed` and returns `True` for an output 2.6x larger than the input:

```
AssertionError: True is not false
...
Validation attempt 1
Validation passed
```

```
$ python -m unittest tests.test_compress_safety
Ran 6 tests in 0.072s
OK
```

Note for anyone reproducing on Windows: this suite has 4 pre-existing errors from the `❌` status `print()`s hitting a cp1252 console (the same ones #619 mentions). `PYTHONIOENCODING=utf-8` clears them; I left that alone since several open PRs are already fixing the encoding layer.

## Scope

One concern only. No mirror files touched — `skills/caveman-compress/scripts/compress.py` is the source of truth and CI resyncs `plugins/caveman/skills/caveman-compress/` on merge.

<sub>Found while auditing the repo for functional bugs; diagnosis and patch drafted with Claude Code, then verified by hand against a checkout — the failing/passing output above is from real runs.</sub>

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -293,6 +293,18 @@ def compress_file(filepath: Path) -> bool:
         print("   already in caveman form. Original file is untouched (no backup created).")
         return False
 
+    # The whole point of the skill is token reduction, but validate() only
+    # checks structural invariants (headings/code blocks/URLs/paths/bullets/
+    # inline code) — it never compares length. A rewrite that is structurally
+    # faithful but LONGER therefore passes every check and gets written and
+    # reported as a successful "compression". Gate on length here, beside the
+    # identical-output guard, so we abort before any backup or write.
+    if len(compressed_body) >= len(body):
+        print("❌ Compression aborted: output is not smaller than input "
+              f"({len(compressed_body)} >= {len(body)} chars).")
+        print("   Original file is untouched (no backup created).")
+        return False
+
     # Reassemble: frontmatter (verbatim) + compressed body
     compressed = frontmatter + compressed_body
 
--- a/tests/test_compress_safety.py
+++ b/tests/test_compress_safety.py
@@ -67,6 +67,25 @@ class CompressSafetyTests(unittest.TestCase):
             self.assertEqual(path.read_text(), original)
             self.assertFalse((Path(tmp) / "task.original.md").exists())
 
+    def test_expanded_compressed_output_does_not_touch_disk(self):
+        # A structurally faithful but LONGER rewrite passes every validate()
+        # check (same heading, no URLs/code/inline code lost), so without a
+        # length gate it was written and reported as a successful compression.
+        # Isolate the backup data dir so no backup lands in the real home dir.
+        with tempfile.TemporaryDirectory() as tmp, \
+             tempfile.TemporaryDirectory() as data_home, \
+             mock.patch.dict(os.environ, {"XDG_DATA_HOME": data_home, "LOCALAPPDATA": data_home}):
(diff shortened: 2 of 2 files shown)


AmirF194 (commented):
Ran this against `561c1ee` in a clean container.

The length gate only runs once, before the retry loop. `compress_file()` still writes an over-length result at `skills/caveman-compress/scripts/compress.py:352`, inside the fix-attempt loop, with no length check on that path.

I mocked `call_claude` to return a short-but-structurally-invalid body first (drops the heading, fails `validate_headings`), then a fix-attempt body that restores the heading but is almost double the original length:

```
$ python3 -m unittest repro_retry_bypass -v
Validation attempt 1
Heading count mismatch: 1 vs 0
Fixing with Claude...
Validation attempt 2
Validation passed
compress_file() returned: True
final file length: 211 original length: 110
final == fix_pass (the longer output)? True
```

`compress_file()` returns `True` and the file on disk ends up longer than it started, the exact bug this PR sets out to close, just reached through `build_fix_prompt` instead of the first pass.

Moving the same `len(...) >= len(...)` check to run on `compressed` before line 352 too (comparing against `original_text`, since `compressed` here already includes frontmatter) would close both paths.

JuliusBrussee:
Reviewed. Your diagnosis of #776 was correct and filed first — thank you for reporting the issue and following it with a fix.

Not cherry-picked, for a mechanical reason: the branch is based on an old main, so its diff against current main deletes workflow files and reverts a large amount of unrelated content. @kavish-19's #986 carries the same fix against current main and applied cleanly, so that is what #993 adopted.

If you would like the credit on the commit rather than only in the PR body, rebasing this onto current main would make it straightforward to take yours instead — the fix itself is not in dispute.

---

PR #779, opened 2026-08-01
fix(opencode): do not toggle caveman on "normal mode" mentions or questions

### LINKED ISSUE #778: opencode plugin: any prompt mentioning "normal mode" deactivates caveman; questions about caveman activate it
**What happen**

`parseModeChange()` in `src/plugins/opencode/plugin.js` deactivates on a bare `/\bnormal mode\b/i` test, so *any* prompt containing the phrase turns caveman off — vim's normal mode being the obvious collision. Natural-language activation in the same function has no `isQuestion` guard, so asking *about* caveman switches it on.

```js
if (/\b(stop|disable|deactivate|turn off)\b.*\bcaveman\b/i.test(prompt) ||
    /\bcaveman\b.*\b(stop|disable|deactivate|turn off)\b/i.test(prompt) ||
    /\bnormal mode\b/i.test(prompt)) {          // <- no anchor, no caveman context
  return 'off';
}
```

This looks like a port gap rather than a fresh bug. `src/hooks/caveman-mode-tracker.js` guards both cases already — #598 anchored "normal mode" to a command position or caveman context (its comment calls out `"how do I exit vim normal mode"` by name) and added the `isQuestion` check. The opencode port never picked either up, so the two implementations disagree.

**Expected**

The opencode plugin should behave like the canonical tracker: "normal mode" counts only in command position or alongside "caveman", and questions are not commands.

**Before/after example**

```
Input: how do I exit vim normal mode?
Got:   parseModeChange -> 'off'   (caveman silently deactivated; user never mentioned caveman)
Want:  parseModeChange -> null

Input: what is caveman mode?
Got:   parseModeChange -> 'full'  (caveman silently activated by a question)
Want:  parseModeChange -> null
```

Verified side by side against the canonical hook on the same prompts — `caveman-mode-tracker.js` leaves the flag alone for both, the opencode port does not.

**Platform**
- [ ] Claude Code
- [ ] Codex
- [x] Other: opencode

**Version / install method**

`main` @ `0d95a81`, repo checkout.

---

Happy to send a PR porting the two guards from `caveman-mode-tracker.js` verbatim, with tests.

<sub>Found while auditing the repo for functional bugs; diagnosis and patch drafted with Claude Code, then verified by hand against a checkout (the outputs above are real runs of both implementations, not predictions).</sub>

divya0795:
Fixes #778

## What

`parseModeChange()` in the opencode plugin deactivated on a bare `/\bnormal mode\b/i` test, so any prompt containing the phrase turned caveman off — `"how do I exit vim normal mode?"` being the obvious collision. Natural-language activation in the same function had no `isQuestion` guard, so `"what is caveman mode?"` silently switched caveman on.

`src/hooks/caveman-mode-tracker.js` already guards both cases from #598 — it anchors "normal mode" to a command position or caveman context, and skips activation on questions. This port never picked those up. The change ports both guards verbatim so the two implementations agree.

The `/caveman` slash branch and the `"activate caveman mode:"` template branch stay ungated — neither can match the question regex, and gating them would break real commands.

## Verification

Ran the same prompt set through both implementations. The port now matches the canonical tracker on every case, including the `"go back to normal mode"` quirk (the alternation is single-shot, so it matches `"back to normal mode"` but not `"go back to normal mode"`) — inherited from the canonical regex, not introduced here.

| prompt | canonical tracker | opencode (before) | opencode (after) |
|---|---|---|---|
| `how do I exit vim normal mode?` | unchanged | **off** | unchanged |
| `in vim, normal mode is the default` | unchanged | **off** | unchanged |
| `what is caveman mode?` | unchanged | **on** | unchanged |
| `does caveman mode help?` | unchanged | **on** | unchanged |
| `normal mode` | off | off | off |
| `back to normal mode` / `switch to normal mode` / `return to normal mode` | off | off | off |
| `put caveman in normal mode` | off | off | off |
| `stop caveman` / `turn off caveman` | off | off | off |

## Test

New `tests/test_opencode_plugin.mjs` (21 cases, `node --test`). It imports `parseModeChange` — which needed a named export alongside the existing `export default CavemanPlugin`; opencode still loads the default export. Importing from the source tree works because `loadConfig()` already has the dev fallback to `src/hooks/caveman-config.js`.

On the pre-fix parser the 6 regression cases fail and the 15 activation/deactivation cases already pass, so behaviour that worked is provably unchanged:

```
$ node --test tests/test_opencode_plugin.mjs      # pre-fix
ℹ tests 21
ℹ pass 15
ℹ fail 6

$ node --test tests/test_opencode_plugin.mjs      # post-fix
ℹ tests 21
ℹ pass 21
ℹ fail 0
```

Nothing else regressed — `python -m unittest tests.test_mode_tracker` 23/23 OK, `node tests/test_mode_tracker_stdin.js` OK, `npm test` 102/112 with the same 10 pre-existing Windows path-separator failures as on `main`.

`npm test` globs `tests/installer/*.test.mjs`, so the new file is run manually, same as `test_mode_tracker_stdin.js` and the Python suites. Broadening that glob felt like a second concern — happy to add it here if you would rather.

## Out of scope

One difference from the canonical tracker remains: `"caveman off"` deactivates there (via `/\bcaveman(\s+mode)?\s+(off|stop|disabled?)\b/`) but is still a no-op in this port. Pre-existing and unrelated to the two guards above — flagging it rather than widening this PR.

<sub>Found while auditing the repo for functional bugs; diagnosis and patch drafted with Claude Code, then verified by hand against a checkout — the table and test output above are from real runs.</sub>

### DIFF
--- a/src/plugins/opencode/plugin.js
+++ b/src/plugins/opencode/plugin.js
@@ -106,11 +106,21 @@ function parseModeChange(promptRaw) {
   prompt = prompt.toLowerCase();
   if (!prompt) return null;
 
+  // Questions about caveman are not commands ("what is caveman mode?",
+  // "does caveman mode help?"). Mirrors caveman-mode-tracker.js (#598).
+  const isQuestion =
+    /^(what|whats|what's|how|why|when|where|who|does|do|did|is|are|can|could|would|should|tell me|explain)\b/.test(prompt);
+
   // Natural-language deactivation — checked before activation so "stop talking
   // like caveman" doesn't trip the activation regex.
   if (/\b(stop|disable|deactivate|turn off)\b.*\bcaveman\b/i.test(prompt) ||
       /\bcaveman\b.*\b(stop|disable|deactivate|turn off)\b/i.test(prompt) ||
-      /\bnormal mode\b/i.test(prompt)) {
+      // "normal mode" only as a command (prompt-initial, optionally led by a
+      // switch-back verb) or with caveman context — never mid-sentence for
+      // e.g. vim's normal mode ("how do I exit vim normal mode"). This port
+      // kept the bare match after caveman-mode-tracker.js was fixed in #598.
+      /^(please\s+)?(go\s+|back\s+to\s+|switch\s+(back\s+)?to\s+|return\s+to\s+)?normal\s+mode\b/.test(prompt) ||
+      (/\bnormal\s+mode\b/.test(prompt) && /\bcaveman\b/.test(prompt))) {
     return 'off';
   }
 
@@ -129,9 +139,13 @@ function parseModeChange(promptRaw) {
     return getDefaultMode();
   }
 
-  // Natural-language activation
-  if (/\b(activate|enable|turn on|start|talk like)\b.*\bcaveman\b/i.test(prompt) ||
-      /\bcaveman\b.*\b(mode|activate|enable|turn on|start)\b/i.test(prompt)) {
+  // Natural-language activation. Gated on !isQuestion so "what is caveman
+  // mode?" asks about caveman instead of silently switching it on. The
+  // template branch above and the slash branch below stay ungated — neither
+  // can match the question regex, and gating them would break real commands.
+  if (!isQuestion &&
+      (/\b(activate|enable|turn on|start|talk like)\b.*\bcaveman\b/i.test(prompt) ||
+       /\bcaveman\b.*\b(mode|activate|enable|turn on|start)\b/i.test(prompt))) {
     const mode = getDefaultMode();
     return mode === 'off' ? null : mode;
   }
@@ -227,3 +241,6 @@ export const CavemanPlugin = async (_ctx) => {
 };
 
 export default CavemanPlugin;
+
+// Named export for unit tests only — opencode loads the default export.
+export { parseModeChange };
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
Nice fix, the before/after table and the pre/post-fix test run make this easy to verify.

One gap in the deactivation side. The new co-occurrence branch at src/plugins/opencode/plugin.js:123, `(/\bnormal\s+mode\b/.test(prompt) && /\bcaveman\b/.test(prompt))`, has no `isQuestion` guard, unlike the activation branch below it. So a genuine question that mentions both words still flips caveman off:

```
"why does caveman mode not work in normal mode?" -> "off"
"does caveman mode work in normal mode" -> "off"
"is caveman broken in normal mode?" -> "off"
```

Reproduced against this PR's head (7d9f40c) with `parseModeChange` imported directly, same harness as the new test file.

Not a regression, main has the same result today since the old bare `/\bnormal mode\b/i` test doesn't look at context either. But it's the same failure mode the rest of this PR fixes (a question silently changing state), just left open on this one branch. Might be worth gating line 123 with `!isQuestion` too, same as the activation check at line 146.

JuliusBrussee:
Thanks for fixing the false mode toggles. Shared parsing and question/“normal mode” guards landed on main through [7693a046](https://github.com/JuliusBrussee/caveman/commit/7693a046) and [da92907f](https://github.com/JuliusBrussee/caveman/commit/da92907f), shipped in [v2.1.0](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this PR as superseded, not rejected.

PR #783, opened 2026-08-01
feat(compress): add provider model selection

### LINKED ISSUE #759: opencode - support other models support other models
**What you want**
Right now all the skills/commands relies only on Claude models, will be great to support other models (like opencode free ones) for operations like `caveman-compress` where there's not the need of a lot of "brain juice"

**Before/after example**
```
Before: compress.py script relies on `call_claude` function
After:  all the skills read the provider/model to use from ENV (or similar)
```

**Why good**
Not all operations (like compressing) need high end models to be efficient/accurate, we can save $ using cheaper/free/opensource models, especially in third parties Harnesses (opencode/py/hermes...)

ousamabenyounes:
Fixes #759

## Summary

- Add env-driven provider routing for `caveman-compress`: default Claude path, explicit `anthropic`, and explicit `opencode`.
- Add compress-specific model env support through `CAVEMAN_COMPRESS_MODEL`, with `CAVEMAN_MODEL` as fallback.
- Send opencode prompts through a temporary `--file` attachment instead of argv, then delete the temp file.
- Document provider/model env behavior and security implications.

CI will resync `plugins/caveman/skills/caveman-compress/...` on merge.

## Test verification (RED -> GREEN)

RED, upstream behavior ignored opencode provider and still called Claude CLI:

```text
FAIL: test_opencode_provider_uses_configured_model
Expected: run(['/usr/local/bin/opencode', 'run', '--model', 'github-copilot/gpt-4.1', 'Compress this memory.'], ...)
  Actual: run(['/usr/local/bin/opencode', '--print'], input='Compress this memory.', ...)
FAILED (failures=1, errors=1)
```

RED, Claude CLI ignored configured model:

```text
FAIL: test_claude_cli_uses_configured_model
Expected: run(['/usr/local/bin/claude', '--model', 'claude-haiku-4-5', '--print'], ...)
  Actual: run(['/usr/local/bin/claude', '--print'], ...)
FAILED (failures=1)
```

RED, opencode path needed non-argv prompt attachment:

```text
ERROR: test_opencode_provider_uses_configured_model
ValueError: '--file' is not in list
FAILED (errors=1)
```

GREEN, provider/model tests:

```text
Ran 3 tests in 0.002s

OK
```

Full local CI baseline on upstream main:

```text
# tests 112
# pass 112
python compress safety: Ran 5 tests ... OK
```

Full local CI final branch:

```text
# tests 112
# pass 112
python compress safety: Ran 8 tests ... OK
```

Review gates:

```text
contribute-code:     PASS
local-ci-docker:     PASS
pr-check:            PASS
ocr-codex:           PASS
main-codex-review:   PASS
```

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -554,8 +554,8 @@ def call_opencode_cli(prompt: str) -> str:
             suffix=OPENCODE_PROMPT_SUFFIX,
             delete=False,
         ) as prompt_file:
-            prompt_file.write(prompt)
             prompt_path = Path(prompt_file.name)
+            prompt_file.write(prompt)
         args.extend(
             [OPENCODE_FILE_ARG, str(prompt_path), OPENCODE_PROMPT_MESSAGE]
         )
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -18,7 +18,7 @@ import sys
 import tempfile
 import time
 from pathlib import Path
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 _IS_WINDOWS = os.name == "nt" or sys.platform == "win32"
 _O_NOFOLLOW = getattr(os, "O_NOFOLLOW", 0)  # unix-only; refuses to open through a pre-placed symlink at the lock path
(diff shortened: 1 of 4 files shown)


skills/caveman-compress/scripts/compress.py
```diff
@@ -394,14 +425,158 @@ def _is_smaller_than_body(candidate_body: str, body: str) -> bool:
...
+        args.extend(
+            [OPENCODE_FILE_ARG, str(prompt_path), OPENCODE_PROMPT_MESSAGE]
+        )
+        return run_cli(OPENCODE_CLI, args)
+    finally:
+        if prompt_path is not None:
+            try:
+                prompt_path.unlink(missing_ok=True)
+            except OSError as error:
+                raise RuntimeError(
```
AmirF194, line 568: I ran this against ab8af98f in a clean container, simulating an opencode call failure together with a failed temp-file unlink.

`call_opencode_cli`'s `finally` re-raises on `unlink` failure (compress.py:567-570). If `run_cli` already raised (opencode auth error, bad prompt, whatever), the delete failure replaces it, and `cli.py`'s top-level handler only prints `str(e)`, so the user never sees why opencode actually failed.

```
$ python3 repro.py
Error: Failed to delete temporary opencode prompt: /tmp/.caveman-compress-prompt-XXXX.md
Real underlying cause, only in e.__context__: OSError('Permission denied (simulated AV lock)')
```

`_unlock` a few functions up swallows this exact class of error on purpose ("swallows errors since callers use this in a finally block", compress.py:139), and the other unlink cleanups in this file do the same (293, 841). Matching that here, instead of raising, would stop a cleanup failure from hiding the call failure that actually mattered.

PR #794, opened 2026-08-04
fix(stats): count each API response once, not once per content-block line

### LINKED ISSUE #793: caveman-stats counts multi-block API responses once per JSONL line — output tokens and turns inflated ~2x on tool-heavy sessions
**Component:** `src/hooks/caveman-stats.js` (`parseSession`)
**Platform:** macOS (darwin 25.5.0), Claude Code

## Summary

`parseSession()` sums `usage.output_tokens` (and `cache_read_input_tokens`) for every `type: "assistant"` line in the session JSONL, and increments `turns` per line.

But Claude Code writes **one JSONL line per content block** of the same API response: a response containing a text block plus two `tool_use` blocks lands as three assistant lines — same `message.id`, same `requestId`, and the **same `usage` object repeated verbatim** on each line. `parseSession` counts that response's tokens three times.

Everything derived from these numbers inflates together:

- `Output tokens` — roughly doubled on tool-heavy sessions
- `Est. tokens saved` and `Est. saved (USD)` — computed from the inflated output figure
- `Turns` — so `Est. rule overhead` (~1,250/turn) is also overstated, in the other direction
- The lifetime history (`.caveman-history.jsonl`) — every snapshot logs the inflated totals
- The statusline `⛏` savings badge — rendered from the inflated lifetime aggregate
- Per-mode attribution (#601) — each duplicate line lands in `messages[]`, so the split inherits the same inflation

The project's own `docs/HONEST-NUMBERS.md` promise is the reason I'm flagging this as a bug and not a nitpick: `/caveman-stats` currently overstates savings ~2x on exactly the kind of agentic session caveman targets.

## Measured on real sessions

Comparing `parseSession`'s method (sum every assistant line) against deduplication by `(requestId, message.id)` across 8 real Claude Code session transcripts:

```
session A   summed=119,513  deduped=57,132  inflation=2.09x
session B   summed= 46,823  deduped=22,377  inflation=2.09x
session C   summed=  3,402  deduped= 1,886  inflation=1.80x
session D   summed=    656  deduped=   435  inflation=1.51x
(4 trivial sessions with no tool calls: 1.00x — single-block responses only)
```

In session B, 35 of 44 API responses had duplicate lines (up to 3 lines per response, `usage` identical on each). `turns` reported 82 where 44 responses actually happened.

Inflation is 1.00x only on sessions with no tool use — one content block per response, one line each. Any real coding session multiplies.

## Reproduction

```python
import json, sys
from collections import defaultdict

f = sys.argv[1]  # any Claude Code session .jsonl with tool calls
total = 0; by_id = defaultdict(list)
for line in open(f):
    line = line.strip()
    if not line: continue
    try: e = json.loads(line)
    except: continue
    if e.get('type') != 'assistant' or not e.get('message'): continue
    u = e['message'].get('usage')
    if not u: continue
    total += u.get('output_tokens', 0)                       # parseSession's method
    by_id[(e.get('requestId'), e['message'].get('id'))].append(u.get('output_tokens', 0))

dedup = sum(v[-1] for v in by_id.values())
print(f"summed={total}  deduped={dedup}  inflation={total/dedup:.2f}x")
```

Run it against any session where the agent used tools; the two numbers diverge immediately.

## Suggested fix

Dedupe by `(requestId, message.id)` inside `parseSession`'s loop — count each API response once:

```js
const seenResponses = new Set();
// inside the loop, after the `usage` guard:
if (entry.message.id) {
  const key = (entry.requestId || '') + ':' + entry.message.id;
  if (seenResponses.has(key)) continue;
  seenResponses.add(key);
}
```

Notes on the key choice:

- `message.id` alone is not enough: a retried request re-sends the same `message.id` under a new `requestId`, and those are distinct billed responses — both must count. (This is the same dedupe key the ccusage tool settled on.)
- Entries without a `message.id` (synthetic logs, older formats) keep per-line counting — there is nothing to dedupe on, and the existing test fixtures take this shape.

Already-written `.caveman-history.jsonl` rows keep their inflated totals — they're snapshots and `aggregateHistory` keeps the latest row per session, so one `/caveman-stats` run per session after the fix self-corrects the lifetime figure for sessions still on disk.

I have this patched locally with three regression tests (multi-block dedupe, retry-path double-count guard, no-id fallback). The dedupe test fails against unpatched `parseSession` and passes with the fix; the full `tests/test_caveman_stats.js` suite passes with no new failures. Happy to open a PR.

pedrohlmelo:
Fixes #793

`parseSession()` sums `usage.output_tokens` for every `type: "assistant"` line in the session JSONL. But Claude Code writes one JSONL line per content block of the same API response — a response with a text block plus two tool_use blocks lands as three assistant lines, same `message.id` + `requestId`, same `usage` object repeated on each. Every multi-block response is counted once per block.

Measured on 8 real session transcripts: 1.5-2.1x inflation on any session with tool calls (1.00x only on trivial no-tool sessions). One example: summed=119,513 vs deduped=57,132 output tokens; `turns` reported 82 where 44 responses happened.

Everything downstream inflates with it: Est. tokens saved, Est. saved (USD), the lifetime history snapshots, the statusline ⛏ badge, and the per-mode attribution from #601 — against the spirit of docs/HONEST-NUMBERS.md.

## Approach

Dedupe by `(requestId, message.id)` inside the parse loop — count each API response once:

- `message.id` alone is not enough: a retried request re-sends the same `message.id` under a new `requestId`, and those are distinct billed responses (same key ccusage uses).
- Entries without `message.id` (synthetic/legacy logs, existing test fixtures) keep per-line counting — nothing to dedupe on, no behaviour change for them.

Existing `.caveman-history.jsonl` rows keep their inflated snapshots, but `aggregateHistory` keeps only the latest row per session — one /caveman-stats run after the fix self-corrects each session still on disk.

## Testing

Three regression tests added to tests/test_caveman_stats.js:

- multi-block response counted once (fails against unpatched `parseSession`)
- same `message.id` under different `requestId`s still counts per response
- entries without `message.id` keep per-line counting

Full tests/test_caveman_stats.js suite passes with no new failures. Neighbouring suites unaffected: test_caveman_parse.js 33/33, test_mode_tracker_stdin.js 11/11, test_symlink_flag.js 15/15, test_repo_local_config.js 14/14.

### DIFF
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -100,6 +100,14 @@ function parseSession(filePath) {
   let turns = 0;
   let model = null;
   const messages = []; // per-message {ts, outputTokens} for mode attribution (#601)
+  // Claude Code writes one JSONL line PER CONTENT BLOCK of an API response
+  // (text block, then each tool_use block), all sharing the same message.id +
+  // requestId and repeating the same usage object. Summing every line counts
+  // the same response's tokens once per block — 1.5-2.1x inflation measured
+  // on real tool-heavy sessions. Count each (requestId, message.id) once.
+  // Entries without a message.id (synthetic/legacy logs) keep per-line
+  // counting — there is no key to dedupe on.
+  const seenResponses = new Set();
   for (const line of raw.split('\n')) {
     if (!line.trim()) continue;
     let entry;
@@ -107,6 +115,11 @@ function parseSession(filePath) {
     if (entry.type !== 'assistant' || !entry.message) continue;
     const usage = entry.message.usage;
     if (!usage) continue;
+    if (entry.message.id) {
+      const key = (entry.requestId || '') + ':' + entry.message.id;
+      if (seenResponses.has(key)) continue;
+      seenResponses.add(key);
+    }
     outputTokens    += usage.output_tokens           || 0;
     cacheReadTokens += usage.cache_read_input_tokens || 0;
     turns++;
--- a/tests/test_caveman_stats.js
+++ b/tests/test_caveman_stats.js
@@ -54,6 +54,55 @@ test('reads --session-file directly and sums output tokens', (tmp) => {
   assert.match(out, /Cache-read tokens:\s+250/);
 });
 
+test('counts a multi-block API response once, not once per JSONL line', (tmp) => {
+  // Claude Code writes one assistant line per content block (text + each
+  // tool_use) of the same API response — same message.id + requestId, same
+  // usage repeated. Only one line per response may count.
+  const usage = { output_tokens: 487, cache_read_input_tokens: 1000 };
+  const sess = makeSession(tmp, [
+    { type: 'assistant', requestId: 'req_1', message: { id: 'msg_a', usage } },
+    { type: 'assistant', requestId: 'req_1', message: { id: 'msg_a', usage } },
+    { type: 'assistant', requestId: 'req_1', message: { id: 'msg_a', usage } },
+    { type: 'user', message: { content: 'tool result' } },
+    { type: 'assistant', requestId: 'req_2', message: { id: 'msg_b', usage: { output_tokens: 13, cache_read_input_tokens: 500 } } },
+  ]);
+  const out = execFileSync(process.execPath, [STATS, '--session-file', sess], {
+    encoding: 'utf8',
+    env: { ...process.env, CLAUDE_CONFIG_DIR: path.join(tmp, '.claude') },
+  });
+  assert.match(out, /Turns:\s+2/);
+  assert.match(out, /Output tokens:\s+500\b/);
+  assert.match(out, /Cache-read tokens:\s+1,?\.?500\b/);
+});
+
+test('same message.id under different requestIds counts per response (retry path)', (tmp) => {
+  // A retried request re-sends the same message.id under a new requestId —
+  // those are distinct billed responses and must both count.
+  const sess = makeSession(tmp, [
+    { type: 'assistant', requestId: 'req_1', message: { id: 'msg_a', usage: { output_tokens: 100 } } },
+    { type: 'assistant', requestId: 'req_2', message: { id: 'msg_a', usage: { output_tokens: 100 } } },
+  ]);
+  const out = execFileSync(process.execPath, [STATS, '--session-file', sess], {
+    encoding: 'utf8',
+    env: { ...process.env, CLAUDE_CONFIG_DIR: path.join(tmp, '.claude') },
+  });
+  assert.match(out, /Turns:\s+2/);
+  assert.match(out, /Output tokens:\s+200\b/);
+});
+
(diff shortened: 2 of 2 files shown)


src/hooks/caveman-stats.js
```diff
@@ -100,13 +100,26 @@ function parseSession(filePath) {
...
+  const seenResponses = new Set();
   for (const line of raw.split('\n')) {
     if (!line.trim()) continue;
     let entry;
     try { entry = JSON.parse(line); } catch { continue; }
     if (entry.type !== 'assistant' || !entry.message) continue;
     const usage = entry.message.usage;
     if (!usage) continue;
+    if (entry.message.id) {
+      const key = (entry.requestId || '') + ':' + entry.message.id;
```
AmirF194, line 119: I cloned this PR's head (0780bca9c) into a clean `node:20-alpine` container and ran the stats test suite there.

The dedup key at line 119 (`(entry.requestId || '') + ':' + entry.message.id`) matches how the JSONL actually repeats itself: every duplicate line for one API response carries the same `message.id`, so the key still holds on entries with no `requestId`. The same `continue` also dedupes `turns` and the per-message array that feeds the #601 mode-attribution tests, and those tests still pass with the new dedup in place, so this doesn't only fix the top-line token counts.

```
$ docker run --rm -v $(pwd):/repo:ro -w /repo node:20-alpine sh -c \
    "apk add --no-cache bash >/dev/null 2>&1 && node tests/test_caveman_stats.js"
...
49 passed, 0 failed
```

I did not test this against a live Claude Code session, only the repo's own fixtures.

Alot1z:
Cross-linking per competing-PR scan: PR #931 (`fix(stats): correct multi-block accounting...`) independently implements the same `(requestId, message.id)` dedup in `src/hooks/caveman-stats.js` that this PR does. Overlap is on the same hunk (the dedup added at #931 line 192). Recommend agreeing merge order so whichever lands first makes the other rebase cleanly / be closed as redundant. See also #931 review comment discussion_r3889378348.

Alot1z:
Heads-up, since your `(requestId, message.id)` dedupe overlaps with work in #931 (stats-accounting).

#931 lands the same key at `src/hooks/caveman-stats.js` — `(requestId, message.id)` with the same ID-normalization guard and the same per-line fallback for ID-less legacy entries. I compared the two line-by-line against #931's head: your multi-block-collapse and distinct-requestId scenarios are both covered there (one is literally a "#794 parity" test). Same counting, no behavioral fork between the two.

Since #931's stats rewrite is a superset (it adds lifetime `--all`/`--since`/`--reset`, a bounded recent-window read, and a test-count drift CI gate), the cleanest outcome is probably to close this one as superseded once #931 lands — no rebase needed on your side, since the content is already inside it.

If you'd rather land #794 first instead, I'm happy to isolate the dedupe into its own first commit in #931 so a rebase is trivially those lines. This is informational only — closing is your call.

Issue #538, opened 2026-06-18
Missing stdin error handler in caveman-mode-tracker.js causes zombie process on broken pipe


HetCreep:
## Version

v1.9.0 - latest release (verified on tag v1.9.0, 2026-06-18)

## Problem

`src/hooks/caveman-mode-tracker.js` lines 19-20 register `stdin.on('data')` and `stdin.on('end')` but have no corresponding `stdin.on('error')` handler.

If stdin closes abnormally (broken pipe, parent process crash, network interruption), Node emits an unhandled `error` event on the stream, which - without a listener - becomes an **uncaught exception** and crashes the process with a non-zero exit code. Under Claude Code's hook system this can surface as a spurious hook failure or leave the subprocess running.

## Code (lines 17-21)

`javascript
let input = '';
process.stdin.on('data', chunk => { input += chunk; });
process.stdin.on('end', () => {
  // ... main body
});
// No stdin.on('error') handler
`

## Fix

Add an error handler before the data/end listeners:

`javascript
process.stdin.on('error', () => {
  // stdin closed abnormally - exit silently, never block the session
  process.exit(0);
});
let input = '';
process.stdin.on('data', chunk => { input += chunk; });
process.stdin.on('end', () => { ... });
`

Using `process.exit(0)` stays consistent with the hook contract: hooks must always exit 0 and never block session start.

## Severity

**HIGH** - Unhandled stream error causes uncaught exception; hook may crash non-zero under broken-pipe conditions

JuliusBrussee:
Thanks for reporting this. This was addressed on main in [`e7ee55f`](https://github.com/JuliusBrussee/caveman/commit/e7ee55f) and shipped in tag [`v2.1.0`](https://github.com/JuliusBrussee/caveman/tree/v2.1.0). I’m closing this as implemented. If the current release still reproduces it, reply with your Caveman version, platform, command, and a minimal reproduction and I’ll reopen it.

HetCreep:
Verified against v2.1.0 (main @ f8b6bfc, tag v2.1.0 == bin-v1.1.0 same commit), Windows 11 + Node v24.19.0.

Not reproducible on the current release -- the fix at e7ee55f is doing what it's meant to. Confirmed by running rather than reading:
- Stripped the `stdin.on('error')` handler back out (matching the pre-fix code) and emitted `error` on `process.stdin` directly -- uncaught exception, exit 1, matching the originally reported crash.
[...]

PR #795, opened 2026-08-05
fix(hooks): handle stdout/stderr 'error' in UserPromptSubmit hook

### LINKED ISSUE #397: UserPromptSubmit hook surfaces "Failed to write to socket (Broken pipe, errno 32)" in Claude Code
I'm a Claude Code session. My user asked me to debug an issue he had locally, and once we identified it was in `caveman` (not on his end), instead of having me apply a patch he said it's better if I just open an issue and propose a prompt to fix it — so maintainers can feed the issue to their own agent to validate, reproduce if they can, and decide on a fix.

## What he saw

In Claude Code, immediately under a normal user prompt, a red status line:

```
UserPromptSubmit hook error
└ Failed with non-blocking status code: Error: Failed to write to socket (Broken pipe, errno 32)
```

The hook is `hooks/caveman-mode-tracker.js`, registered in `.claude-plugin/plugin.json` under `UserPromptSubmit`. The error is marked **non-blocking**, and we confirmed the hook's *side effect* (writing `~/.claude/.caveman-active`) does succeed — the flag file's mtime updates on each prompt. So this is cosmetic, not a functional break, but it's noisy and shows up under most prompts.

## Pre-flight: not a duplicate

Searched existing issues + PRs before filing:

- #392 (open) — UserPromptSubmit hook fires *twice* because installer wires both plugin + standalone hooks. Different root cause (duplicate registration, not stdio).
- #313 (closed) — different non-blocking error (`node: command not found` on GitBash Windows).
- #364 (closed) — `curl | bash` install pipe failure, unrelated to runtime hook stdio.
- PR #377 (open) — also modifies the UserPromptSubmit hook, but for short-circuiting on non-interactive `sdk-cli` probes. The EPIPE here surfaces under normal interactive Claude Code sessions and is orthogonal to that PR; a fix can merge before or after #377.

No existing thread matches "broken pipe", "EPIPE", or "errno 32" against the mode tracker.

## Root cause (our read)

`caveman-mode-tracker.js` does the following:

1. Registers `process.stdin.on('data', …)` to accumulate the prompt JSON.
2. On `'end'`, parses, optionally writes the flag file.
3. Never writes anything to stdout, never explicitly exits — relies on Node's event loop draining naturally.

Two things combine to surface EPIPE on the harness side:

- The hook never emits a stdout payload, so Node exits the moment the `'end'` callback returns. If the harness still has any pending write to the child's stdio (status pipe, or a late stdin flush on some platforms), it gets `EPIPE` when the kernel reaps the child.
- There is no `'error'` handler on `process.stdout` / `process.stderr`. If Node *does* attempt a write to a closed pipe (e.g. via `console.*` or any diagnostic), the EPIPE bubbles as an uncaught exception, which Claude Code's hook executor turns into the visible "Failed with non-blocking status code" line.

We don't have a deterministic repro — it's a race — but on the affected machine it fires under most prompts in a fresh session. `caveman-activate.js` (SessionStart) already writes a payload and then exits, which is likely why it doesn't show the same symptom.

## Reproduction

Environment we saw it in:
- macOS, arm64
- Latest Claude Code CLI at time of report
- Node v20.x system install
- `caveman` plugin installed via the marketplace (`extraKnownMarketplaces.caveman.source.repo = JuliusBrussee/caveman`), enabled in `~/.claude/settings.json`
- Default mode `full`

Steps:
1. Enable the `caveman` plugin in `~/.claude/settings.json`:
   ```json
   "enabledPlugins": { "caveman@caveman": true }
   ```
2. Start a Claude Code session in any directory.
3. Send any short prompt (e.g. `test`).
4. Observe the red "UserPromptSubmit hook error … Broken pipe, errno 32" line under your prompt.

Verifying the hook itself works in isolation (it does):
```bash
echo '{"prompt":"test"}' | node \
  ~/.claude/plugins/cache/caveman/caveman/<hash>/hooks/caveman-mode-tracker.js
# exit 0, no error, ~/.claude/.caveman-active gets written
```

So the bug only manifests under Claude Code's hook executor's stdio plumbing, not under a plain shell pipe.

## Suggested fix (what worked locally)

Two small changes to `hooks/caveman-mode-tracker.js`:

1. Attach an EPIPE-tolerant `error` handler on stdout/stderr.
2. Write a tiny ack (`"OK"`) and only `process.exit(0)` after the write has flushed.

Patch:

```js
// near the top of caveman-mode-tracker.js, after the requires:
for (const stream of [process.stdout, process.stderr]) {
  stream.on('error', err => { if (err && err.code === 'EPIPE') return; throw err; });
}

// at the very end of the 'end' handler, replacing the implicit fall-through:
try {
  process.stdout.write('OK', () => process.exit(0));
} catch (e) {
  process.exit(0);
}
```

After applying both, the red status disappears on every prompt tried, and the flag-file behavior is unchanged. `caveman-activate.js` already follows this pattern (writes a payload, then exits), which is why the SessionStart hook doesn't have the same symptom.

## Prompt a maintainer can feed to their own agent

> You are working on the `JuliusBrussee/caveman` repo. There is a bug report that the `UserPromptSubmit` hook (`hooks/caveman-mode-tracker.js`) intermittently produces a "Failed to write to socket (Broken pipe, errno 32)" non-blocking error in Claude Code, even though the hook's side effect (writing `~/.claude/.caveman-active`) succeeds.
>
> 1. Reproduce: install the plugin from this repo into a clean Claude Code config, send a few short prompts, and watch the status line. The error should appear under the prompt. If it does not reproduce on your machine, document that and stop — do not "fix" speculatively.
> 2. If you reproduce, confirm the hypothesis: the hook never writes to stdout and never installs an `error` handler on stdout/stderr, so any late write attempt or pipe close from the harness side surfaces EPIPE.
> 3. Apply the minimal patch: (a) attach `'error'` handlers on `process.stdout` and `process.stderr` that swallow EPIPE and re-throw other errors, (b) at the end of the `'end'` callback, write a small ack and only `process.exit(0)` after the flush callback fires. Mirror the pattern already used by `caveman-activate.js`.
> 4. Add a test under `tests/` that pipes a short JSON payload into the hook and asserts exit 0 + no thrown errors, plus a smoke check that the flag file is written for a `/caveman ultra` prompt.
> 5. Do not change the hook's public contract (the flag file path, the parsed commands, the deactivation regex). Keep the diff minimal.
> 6. If your investigation reveals a deeper root cause (e.g. a Claude Code side issue), report it instead of patching defensively — the EPIPE swallow is fine if it's the right defensive move, but document why.

## Notes

- My user reviewed this issue text before posting; nothing sensitive from his local config has been included.
- Happy to test a fix on the affected machine if useful — just ping back here.

AmirF194:
`caveman-mode-tracker.js` listens for `'error'` on `process.stdin` (added for #538) but never on `process.stdout`/`process.stderr`. When the harness closes its end of the pipe after the hook has already started writing a payload, the write throws `EPIPE` as an uncaught exception and the hook exits non-zero, surfacing as the reported "Failed to write to socket (Broken pipe, errno 32)" status line under an otherwise normal prompt. There are two write sites this can hit: the per-turn reinforcement (`additionalContext` for an active mode) and the `/caveman-stats` block.

Fix mirrors #538's own pattern for stdin: attach an `'error'` listener to `stdout`/`stderr` that exits cleanly instead of letting Node throw. No change to the flag file, the parsed commands, or the deactivation regex.

## Verification

- New test (`tests/test_mode_tracker_stdout.js`), modeled on the existing `test_mode_tracker_stdin.js` stdin-error test: injects an `'error'` event on `process.stdout`/`process.stderr` in a child process running the real hook. Fails on unpatched `main` (exit 1, "Unhandled 'error' event"), passes on this branch (exit 0).
- Reproduced the original report end-to-end in Docker: spawned the hook as a child with an active mode, destroyed the parent's read end of the child's stdout before the reinforcement write, forcing a real `EPIPE`. Child exits 1 with the traceback on `main`, exit 0 on this branch. Same repro against the `/caveman-stats` write site.
- Full suite green on this branch, matching the repo's `ci.yml` job set exactly: `npm test` (installer) plus every `tests/test_*.js` on node 18/20/22, and `python -m unittest discover -s tests` (`test_caveman_stats.js`'s 5 failures are pre-existing on `main`, verified, unrelated to this change).
- Not verified: production frequency of the race (the repro forces the pipe closed deterministically; how often the harness actually does this on a live machine is unmeasured).

Fixes #397


AmirF194:
Checking in a week after opening this one, no rush. CI is green (4/4) and the branch is clean against main, so it's ready whenever you get to it.

AmirF194:
Flagging for whoever reviews this: #849 also touches caveman-mode-tracker.js and mentions this issue in passing, worth knowing about before merging either. #849's fix is for a different root cause though (a missing sibling module, #848/#801), and drains stdin before exiting as a side benefit for the module-load-failure path specifically; it does not add an error listener on stdout/stderr, so the broken-pipe crash this PR fixes (an unhandled 'error' event when the harness closes its end of our output) is still live in `#849`'s branch. The two edits land in different regions of the file (near the top-level requires vs the existing error handlers a few lines down), so I don't expect a real conflict, just worth a heads-up.

JuliusBrussee:
Merged through #929 after adversarial review. The original stream-error handling remains, with added lifecycle and Windows timing hardening. Final cross-platform CI passed.

AmirF194:
Thanks for merging, and for folding in the extra lifecycle and Windows timing hardening through #929. Glad the stream-error handling held up through that.

PR #798, opened 2026-08-05
fix(compress): serialize concurrent compress runs with a cross-session lock

hunleyd:
## Summary
- `compress_file` had no coordination between two processes targeting the same file at once — two Claude Code sessions (or any two invocations) running `caveman-compress` on the same file concurrently could interleave reads/writes, producing corrupted output (observed: a stray unrelated bullet from one session's edit surviving into the other's compression output, plus a literal timestamp string getting written into file content).
- Adds `file_lock()`, a cross-platform exclusive lock backed by the OS's own file lock (`fcntl.flock` on POSIX, `msvcrt.locking` on Windows). `compress_file()` now resolves the path, acquires the lock, then runs the existing body (moved unchanged into `_compress_file_locked`).
- Because the lock is the OS's own advisory lock rather than a hand-rolled marker file, a crashed or killed holder releases it automatically the instant its file descriptor closes — there's no age-based staleness bookkeeping to get wrong, and no window for two processes to race over whether a lock looks abandoned. A live lock still held after `LOCK_WAIT_SECONDS` (900s — see below) raises `LockTimeoutError` with a clear message instead of corrupting the file or hanging indefinitely. The marker file itself is intentionally never deleted (only unlocked) — unlinking it while a lock is live would let a third process race a fresh file into the same path.
- The lock is keyed on the same `(parent-dir-name, stem)` domain `backup_dir_for` already uses for its own collision guard, hashed rather than embedded as plaintext in the filename so a lock for a file later refused as sensitive (e.g. `id_rsa`) never leaks its name into the state directory.
- `backup_dir_for`'s platform-detection logic (Windows `%LOCALAPPDATA%` vs XDG) is factored into a shared `_state_base_dir(kind)` helper, reused for the new lock directory — no behavior change to `backup_dir_for` itself.
- Hardening to match this repo's own `safeWriteFlag` convention for predictable-path state files: the lock file opens with mode `0600`, and `file_lock` refuses outright if the lock directory itself is a symlink (the `_O_NOFOLLOW` guard on the final path component doesn't cover that).

## What was considered and rejected
- A hand-rolled `os.O_CREAT | os.O_EXCL` marker file (atomic create-if-absent) was the original design, but it needs its own age-based staleness reclaim to recover from a crashed holder — and that reclaim logic is itself a second race between whichever process decides the lock looks abandoned first. The OS's own advisory lock (`fcntl.flock` / `msvcrt.locking`) sidesteps the whole problem: the kernel releases it the moment the holder's fd closes, crash or clean exit alike, so there's no staleness state to track or race over. `msvcrt.locking` needs at least one byte in the file to lock, which is why `file_lock` writes a single null byte on first use.
- Reusing the existing "backup file already exists" guard as an implicit lock was considered, but it only catches contention *after* the first process has already started writing — it doesn't prevent two processes both reaching the read/compress step before either writes, which is the actual race.
- A per-resolved-path lock key (rather than `backup_dir_for`'s `(parent-dir-name, stem)` domain) was the first cut, but it under-locks: two files that collide on the same backup path via that domain would still take separate locks, letting both pass the "backup already exists" guard before either writes.
- Known limitation, not fixed here: like any advisory OS-native lock, this doesn't enforce on filesystems that don't honor `flock`/byte-range locking (some NFS mounts without lockd, certain SMB/WSL `drvfs` mounts) — the lock silently degrades to a no-op there. Not addressed in this PR; flagging it for anyone hitting corruption on such a mount.

## Test plan
- [x] `python3 -m unittest tests.test_compress_concurrency` — 11 tests: lock keyed correctly on `(parent-dir-name, stem)` (not the full path, and not basename alone), same key shared across different absolute paths that collide on the backup path, second acquire blocks until first releases, a crashed holder's lock is released by the OS the instant its fd closes, a live lock is never stolen and correctly times out, the lock is released on both normal return and exception, the lock directory is refused if it or its lock path is a symlink, and a full `compress_file` integration test proving two concurrent calls on the same file serialize (exactly one `call_claude` invocation, no interleaved writes) instead of racing.
- [x] `python3 -m unittest tests.test_compress_safety` — all 10 pre-existing tests still pass unchanged.
- [x] `python3 -m unittest discover -s tests -p "test_*.py"` — no new failures introduced; the 8 pre-existing failures in `test_hooks.py`/`test_mode_tracker.py` are unrelated to this change and reproduce identically on `main` with this branch's changes stashed out.


skills/caveman-compress/scripts/compress.py
```diff
@@ -75,28 +87,108 @@ def split_frontmatter(text: str):
...
-    return base / filepath.parent.name
+    return base / "caveman-compress" / kind
+
+
+def backup_dir_for(filepath: Path) -> Path:
+    """Out-of-tree backup dir for filepath, keyed by its parent dir name — kept outside the source tree so skill auto-loaders don't re-ingest `.original.md` backups as live files."""
+    return _state_base_dir("backups") / filepath.parent.name
+
+
+LOCK_WAIT_SECONDS = 900  # must outlast a legitimate holder's worst-case run (up to MAX_RETRIES+1 Claude calls against the 500KB size cap) or a healthy wait misreads as a stuck lock
```
AmirF194, line 106: I ran call_claude() against a hung `claude` CLI in a clean `python:3.12-slim` container, at this head.

`subprocess.run([claude_bin, "--print"], ...)` in `call_claude()` (scripts/compress.py, unchanged by this PR) sets no `timeout`. With a fake `claude` shim that just sleeps, the call never returns on its own; I had to kill it externally after 8s (exit 124), and it would have blocked just as long at 900s or 9000s.

That matters here because `LOCK_WAIT_SECONDS = 900` at line 106 is commented "must outlast a legitimate holder's worst-case run (up to MAX_RETRIES+1 Claude calls against the 500KB size cap)". That assumes the run is bounded, but a stalled `claude` CLI (dropped network, an auth prompt with no TTY to answer it) holds `file_lock` forever, so the 900s figure isn't actually a computed worst case, it's a guess. A waiter still fails safe (it raises `LockTimeoutError` and gives up rather than hanging), so this isn't data loss, just a lock that can outlive the bound its own comment claims for it.

Passing a timeout into that `subprocess.run` call (and into the SDK branch's `client.messages.create`) would make the comment's claim actually true, and would turn an indefinite hang into a clean failure the retry loop already knows how to handle.

hunleyd: Fair catch. Added `CLAUDE_CALL_TIMEOUT_SECONDS = LOCK_WAIT_SECONDS // (MAX_RETRIES + 1)` and passed it to both the `subprocess.run` call and the `anthropic.Anthropic(...)` client's `timeout=`. A stalled CLI/API call now raises after ~300s instead of hanging indefinitely, so `LOCK_WAIT_SECONDS`'s worst-case comment is now an enforced bound rather than an assumption. Added `tests/test_call_claude_timeout.py` covering both the passed-through timeout value and that a `TimeoutExpired` surfaces as `RuntimeError`. Pushed in ece94d5.

AmirF194: Confirmed, the derived timeout on both call sites closes the gap. Thanks for the fix and the test.

JuliusBrussee:
Merged through #929. Cross-session compression locking remains, alongside exact code-block masking and subprocess isolation added during review. Compression and concurrency suites passed.

PR #802, opened 2026-08-06
Fix inconsistencies in off commands; ultra mode inconsistencies

717jonathan:
1. Adding `caveman off` in persistence rule so it is not ignored.
2. Fixed conflicting rules about arrows in ultra mode. No arrows should be used based on line 44 of `skills\caveman\SKILL.md`


JuliusBrussee:
Thanks @717jonathan — both off handling and the no-arrow ultra contract are now consistent on main. Closing as implemented.

PR #804, opened 2026-08-06
fix(hooks): resolve node robustly in SessionStart/UserPromptSubmit hooks

fkucharczyk:
## Problem
The `SessionStart` and `UserPromptSubmit` plugin hooks invoke `node` directly
(`.claude-plugin/plugin.json`). Claude Code runs hook commands with a
minimal, non-interactive PATH (no nvm/profile sourcing), so on any machine
where node is only on PATH via nvm/an interactive shell, every single prompt
prints:

```
UserPromptSubmit hook error ... /bin/sh: 1: node: not found
```

Non-blocking, but constant noise across every session.

## Solution
Added `src/hooks/run-with-node.sh`, a small POSIX-sh wrapper that:
1. tries `command -v node` / `nodejs` (PATH),
2. falls back to common absolute install locations (`~/.nvm/versions/node/*/bin/node`,
   `/usr/local/bin`, `/opt/homebrew/bin`, `/usr/bin`),
3. if node genuinely isn't installed anywhere, exits 0 silently rather than
   printing hook-error noise every turn.

Both hook entries in `plugin.json` now route through it. Added
`tests/test_run_with_node.py` (PATH resolution, nvm-absolute-path fallback,
silent skip when absent) and a `verify_repo.py` regression guard asserting
both hook commands route through the wrapper (so a future edit can't
silently reintroduce the bare `node` call).

## Test plan
- [x] `python -m unittest discover -s tests` — 61/61 pass
- [x] `python tests/verify_repo.py` — all local checks pass
- [x] Manually reproduced the bug (PATH with no `node`, only nvm-installed)
      and confirmed the wrapper resolves via the nvm fallback and the real
      `caveman-mode-tracker.js` hook runs and emits correct JSON output

## Scope note
Verified on POSIX (Linux/WSL) — matches the reported `/bin/sh` failure signature. Native Windows (non-WSL) Claude Code would need `sh.exe` on PATH (ships with Git for Windows, common on Node dev machines) for the wrapper to run; not verified there.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -16,7 +16,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\"",
+            "command": "sh \"${CLAUDE_PLUGIN_ROOT}/src/hooks/run-with-node.sh\" \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\"",
             "timeout": 5,
             "statusMessage": "Loading caveman mode..."
           }
@@ -28,7 +28,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-mode-tracker.js\"",
+            "command": "sh \"${CLAUDE_PLUGIN_ROOT}/src/hooks/run-with-node.sh\" \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-mode-tracker.js\"",
             "timeout": 5,
             "statusMessage": "Tracking caveman mode..."
           }
--- a/src/hooks/install.ps1
+++ b/src/hooks/install.ps1
@@ -24,7 +24,7 @@ $HooksDir = Join-Path $ClaudeDir "hooks"
 $Settings = Join-Path $ClaudeDir "settings.json"
 $RepoUrl = "https://raw.githubusercontent.com/JuliusBrussee/caveman/main/src/hooks"
 
-$HookFiles = @("package.json", "caveman-config.js", "caveman-parse.js", "caveman-activate.js", "caveman-mode-tracker.js", "caveman-stats.js", "caveman-statusline.sh", "caveman-statusline.ps1", "cavecrew-model-overrides.js")
+$HookFiles = @("package.json", "caveman-config.js", "caveman-parse.js", "caveman-activate.js", "caveman-mode-tracker.js", "caveman-stats.js", "caveman-statusline.sh", "caveman-statusline.ps1", "cavecrew-model-overrides.js", "run-with-node.sh")
 
 # Resolve source — works from repo clone or remote
 $ScriptDir = if ($PSScriptRoot) { $PSScriptRoot } else { $null }
@@ -134,7 +134,7 @@ if (!hasStart) {
   settings.hooks.SessionStart.push({
     hooks: [{
       type: 'command',
-      command: 'node "' + hooksDir + '/caveman-activate.js"',
+      command: 'sh "' + hooksDir + '/run-with-node.sh" "' + hooksDir + '/caveman-activate.js"',
       timeout: 5,
       statusMessage: 'Loading caveman mode...'
     }]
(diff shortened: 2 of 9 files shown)


fkucharczyk:
Reproduced this locally (WSL, node only via nvm, no shell profile sourced by hook PATH) — after a reboot, `UserPromptSubmit` failed with `node:internal/modules/cjs/loader:1520 ... Cannot find module`.

Root cause on my machine wasn't the plugin-manifest hooks themselves — those already route through `run-with-node.sh` in my installed copy (cache commit `ec83e5bace4c`) and resolve fine via the nvm-absolute-path fallback. The actual failure was two **orphaned standalone hook entries in `settings.json`** left over from a pre-plugin manual install: bare `node "/home/<user>/.claude/hooks/caveman-mode-tracker.js"` pointing at a `~/.claude/hooks/` dir that no longer existed post-reboot. Those don't route through `run-with-node.sh` at all, and `installClaude()`'s `pruneOrphanedManagedHooks` self-heal (issue #471) only runs on an explicit `caveman`/installer re-run — not automatically at Claude Code startup, so the orphaned entries just kept firing every prompt.

Fix applied locally: re-ran `node cli/install.js --only claude`, which pruned the 2 orphaned entries and confirmed the plugin manifest already owns both `SessionStart`/`UserPromptSubmit`. Verified zero hook errors afterward (both hooks exit 0 under a minimal PATH with no nvm/profile sourcing).

Flagging in case others hitting this after a reboot have the same orphaned-standalone-hook shape rather than a missing wrapper — worth a line in this PR's description or a follow-up noting `pruneOrphanedManagedHooks` output should maybe also be checked/suggested when this exact loader error is reported.

fkucharczyk:
Agent independent review
Verdict: FINDINGS
Reviewed SHA: a575b563046aa66e298af0e61a3a733accfd8953
Evidence: `python3 -m unittest discover -s tests` 61/61 pass; `python3 tests/verify_repo.py` all pass; manual wrapper exec — ~/workbench-ag/2026-08-18_review-caveman-804/

1. [certain] src/hooks/install.sh:146,162 (+ install.ps1:137,153) still emit bare `node "..."` hook commands — the exact bug this PR fixes, unpatched there. Still run live by tests/verify_repo.py:296,387,399 and tests/test_hooks.py, and its own header self-documents direct curl-pipe use — hook noise still reproduces on that path. The new verify_repo.py guard only checks plugin.json, so it can't catch this gap.
2. [likely] run-with-node.sh:21 — `set -eu` + unset `$HOME` aborts the `for candidate in "$HOME"/...` line with `HOME: parameter not set` to stderr (confirmed: `env -i PATH=/usr/bin:/bin sh run-with-node.sh x.js`). Exit stays 0 via the `|| exit 0` on the subshell, but it reprints the "unexpected hook stderr" noise this PR exists to kill, from a different trigger.

Deferred: none

Other effects: cli/install.js's own hook-writer (~line 1011) already bakes an absolute node path + migrates legacy bare-node commands — unaffected, already correct.
Redundant: install.sh/.ps1 duplicate cli/install.js's job and are the only place the fixed bug still lives — deletion candidates, not a silent-fix target.
Dead code: none — install.sh/.ps1 still exercised by tests/verify_repo.py.
Historical narration: none found.
Decaying claims: PR's "61/61 pass" true at reviewed SHA; count will drift as tests are added, repeatable local check, low risk.

fkucharczyk:
**Agent fix-up**

Rebased onto `main` tip (`99a9aa2`) — was CONFLICTING, now MERGEABLE. Pushed `206c460`.

1. install.sh/install.ps1 bare `node` → both now route through run-with-node.sh, same as plugin.json. Added run-with-node.sh to install/uninstall HOOK_FILES lists (uninstall cleans it up too). New test: `test_hooks.py::test_install_wires_hooks_through_run_with_node_wrapper`.
2. run-with-node.sh `set -eu` + unset HOME → `"$HOME"` changed to `"${HOME:-}"`. New test: `test_run_with_node.py::test_survives_unset_home_without_crashing`.
3. Rebase — done, see above.

Verification:
- shellcheck run-with-node.sh install.sh → clean, no findings.
- `python3 -m unittest discover -s tests` → 72/72 pass.
- `python3 tests/verify_repo.py` → all checks pass.
- Repro before/after (`env -i PATH=/usr/bin:/bin sh run-with-node.sh x.js`, HOME unset): before → `HOME: parameter not set` on stderr; after → silent, exit 0.
- Sanity: HOME set + node only reachable via nvm-shaped fallback still resolves and runs (no regression).

Not merging — over to you.

PR #806, opened 2026-08-07
fix(hooks): include caveman-parse.js in standalone installers

### LINKED ISSUE #801: install.sh / install.ps1 omit caveman-parse.js — standalone hook install exits 1 on every prompt
The two standalone hook installers copy a file list that omits `caveman-parse.js`, which `caveman-mode-tracker.js` requires at module load. Anyone who installs hooks via `install.sh` / `install.ps1` — rather than the plugin or `cli/install.js` — ends up with a `UserPromptSubmit` hook that exits 1 on every prompt.

### Affected

`src/hooks/install.ps1` L27:

```powershell
$HookFiles = @("package.json", "caveman-config.js", "caveman-activate.js", "caveman-mode-tracker.js", "caveman-stats.js", "caveman-statusline.sh", "caveman-statusline.ps1", "cavecrew-model-overrides.js")
```

`src/hooks/install.sh` L40 has the same omission.

`src/hooks/caveman-mode-tracker.js` L10:

```js
const { parseModeChange, INDEPENDENT_MODES } = require('./caveman-parse');
```

That is a top-level require, so it throws before the `try`/`catch` inside the stdin `end` handler. The "Hooks must always exit 0" guarantee in the L20-23 comment does not hold for this failure mode.

### Reproduction

Copy exactly the 8 files `install.ps1` lists into an empty directory, then:

```
echo '{"prompt":"hello","cwd":"C:/"}' | node caveman-mode-tracker.js
```

```
Error: Cannot find module './caveman-parse'
Require stack:
...
```

exit code 1.

Control: same directory with `caveman-parse.js` added, same command, exit code 0 and no output.

Environment: Windows 11, Node v24.15.0, repo at `ec83e5b`.

### Not affected

`cli/install.js` L51 already lists `caveman-parse.js`, and the plugin path wires hooks from `.claude-plugin/plugin.json` against the repo tree. Both are fine — only the two shell installers are broken.

### Secondary issue in the same files

`install.ps1` L25 and `install.sh` L38 set the remote fallback base to `.../main/hooks`, but the hooks live under `src/hooks/`. There is no top-level `hooks/` directory, so when either script runs outside a clone, every file in the copy loop 404s:

```
$ curl -sI https://raw.githubusercontent.com/JuliusBrussee/caveman/main/hooks/caveman-config.js
HTTP/2 404
```

### Suggested fix

Add `caveman-parse.js` to both file lists, and point the fallback base at `main/src/hooks`.

vishnujayvel:
## Summary

Fixes #801 (reported by @prpr289).

Standalone hook installers `src/hooks/install.sh` and `src/hooks/install.ps1` copied a `HOOK_FILES` / `$HookFiles` list that omitted `caveman-parse.js`. `caveman-mode-tracker.js` top-level-`require`s `./caveman-parse`, so a UserPromptSubmit hook installed via those scripts exits 1 on every prompt.

Secondary bug in the same files: remote fallback base URL pointed at `.../main/hooks` while hooks live under `src/hooks/` — every remote fetch 404s.

### Changes (exactly two files)

1. **`src/hooks/install.sh`** — add `caveman-parse.js` to `HOOK_FILES`; set `REPO_URL` to `.../main/src/hooks`
2. **`src/hooks/install.ps1`** — add `caveman-parse.js` to `$HookFiles`; set `$RepoUrl` to `.../main/src/hooks`

### Out of scope (intentional)

- **`cli/install.js`** already lists `caveman-parse.js` and pins a ref (`v1.10.0`) with a checksum integrity gate. Not touched.
- **`checksums.sha256`** not touched.
- Pinning a ref in the shell installers (like `cli/install.js` does) is left to maintainer preference; this PR is the minimal fix for the broken file list + 404 URL.

Shell installers remain a supported install path for users who run them directly; this keeps that path working.

## Test plan

No automated tests cover `src/hooks/install.sh` / `install.ps1` (repo `npm test` only runs `tests/installer/*.test.mjs` against `cli/install.js`). Manual repro + control is the coverage for this change.

### Repro (before fix — old file list, no `caveman-parse.js`)

```bash
tmpdir=$(mktemp -d)
# files from pre-fix HOOK_FILES in install.sh
for f in package.json caveman-config.js caveman-activate.js caveman-mode-tracker.js \
         caveman-stats.js caveman-statusline.sh cavecrew-model-overrides.js; do
  cp src/hooks/$f "$tmpdir/"
done
cd "$tmpdir"
echo '{"prompt":"hello","cwd":"/tmp"}' | node caveman-mode-tracker.js
# expected: Error: Cannot find module './caveman-parse' ; exit 1
```

**Observed (Node v26.5.0):**

```
Error: Cannot find module './caveman-parse'
Require stack:
- .../caveman-mode-tracker.js
...
Node.js v26.5.0
EXIT:1

### DIFF
--- a/src/hooks/install.ps1
+++ b/src/hooks/install.ps1
@@ -22,9 +22,9 @@ if (-not (Get-Command node -ErrorAction SilentlyContinue)) {
 $ClaudeDir = if ($env:CLAUDE_CONFIG_DIR) { $env:CLAUDE_CONFIG_DIR } else { Join-Path $env:USERPROFILE ".claude" }
 $HooksDir = Join-Path $ClaudeDir "hooks"
 $Settings = Join-Path $ClaudeDir "settings.json"
-$RepoUrl = "https://raw.githubusercontent.com/JuliusBrussee/caveman/main/hooks"
+$RepoUrl = "https://raw.githubusercontent.com/JuliusBrussee/caveman/main/src/hooks"
 
-$HookFiles = @("package.json", "caveman-config.js", "caveman-activate.js", "caveman-mode-tracker.js", "caveman-stats.js", "caveman-statusline.sh", "caveman-statusline.ps1", "cavecrew-model-overrides.js")
+$HookFiles = @("package.json", "caveman-config.js", "caveman-activate.js", "caveman-mode-tracker.js", "caveman-parse.js", "caveman-stats.js", "caveman-statusline.sh", "caveman-statusline.ps1", "cavecrew-model-overrides.js")
 
 # Resolve source — works from repo clone or remote
 $ScriptDir = if ($PSScriptRoot) { $PSScriptRoot } else { $null }
--- a/src/hooks/install.sh
+++ b/src/hooks/install.sh
@@ -35,9 +35,9 @@ fi
 CLAUDE_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"
 HOOKS_DIR="$CLAUDE_DIR/hooks"
 SETTINGS="$CLAUDE_DIR/settings.json"
-REPO_URL="https://raw.githubusercontent.com/JuliusBrussee/caveman/main/hooks"
+REPO_URL="https://raw.githubusercontent.com/JuliusBrussee/caveman/main/src/hooks"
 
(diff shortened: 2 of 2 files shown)


JuliusBrussee:
Thanks @vishnujayvel — standalone installs now include the shared parser. Closing as implemented.

PR #809, opened 2026-08-07
fix(installer): preserve stable Node path

### LINKED ISSUE #805: brew upgrade node break hook
**What happen**
node is defined as cellar full path

Such that after each node upgrade, the hook would break.

After node upgrade:
```
/opt/homebrew/Cellar/node/26.5.0/bin/node
dyld[26155]: Library not loaded: /opt/homebrew/opt/ada-url/lib/libada.3.dylib
  Referenced from: <87FBC746-7D47-3FD8-B0A3-97018CBF954B> /opt/homebrew/Cellar/node/26.5.0/bin/node
  Reason: tried: '/opt/homebrew/opt/ada-url/lib/libada.3.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/ada-url/lib/libada.3.dylib' (no such file), '/opt/homebrew/opt/ada-url/lib/libada.3.dylib' (no such file), '/opt/homebrew/Cellar/ada-url/4.0.0/lib/libada.3.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/Cellar/ada-url/4.0.0/lib/libada.3.dylib' (no such file), '/opt/homebrew/Cellar/ada-url/4.0.0/lib/libada.3.dylib' (no such file)
```

Workaround:
I have to manually rewrite the hook file to use `node` instead of cellar full path after each upgrade of caveman.

Hook content `~/.claude/settings.json`
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "hooks": [
          {
            "type": "command",
            "command": "rtk hook claude"
          }
        ]
      }
    ],
    "SessionStart": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "\"/opt/homebrew/Cellar/node/26.5.0/bin/node\" \"/Users/rdevops/.claude/hooks/caveman-activate.js\"",
            "timeout": 5,
            "statusMessage": "Loading caveman mode..."
          }
        ]
      }
    ],
    "UserPromptSubmit": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "\"/opt/homebrew/Cellar/node/26.5.0/bin/node\" \"/Users/rdevops/.claude/hooks/caveman-mode-tracker.js\"",
            "timeout": 5,
            "statusMessage": "Tracking caveman mode..."
          }
        ]
      }
    ]
  },
  "statusLine": {
    "type": "command",
    "command": "bash \"/Users/rdevops/.claude/hooks/caveman-statusline.sh\""
  }
}
```

**Expected**
Latest node should be used per brew install

**Platform**
- [ ] Claude Code
- [ ] Codex
- [x] Other: VSCode on Claude hook

**Version / install method**
`npx -y github:JuliusBrussee/caveman -- --only claude`

CBOSSX:
## Summary

- Fixes: The standalone installer persists a versioned Homebrew Cellar Node path in Claude hook commands, so the hooks stop launching after brew upgrades Node and removes that versioned directory.
- Root cause: absoluteNodePath() always returns process.execPath, which resolves the stable PATH symlink to its versioned target; the installer then serializes that transient target into settings.json instead of preserving the stable executable path found on PATH.
- Fix: On POSIX, preserve the Node executable returned by `command -v node`, with `process.execPath` retained as the fallback for minimal-PATH launchers.

Fixes #805

## Regression evidence

- Before: `node --test --test-name-pattern="standalone hooks keep a stable PATH node symlink" tests/installer/e2e.freshinstall.test.mjs` exited `1`

- After: `node --test --test-name-pattern="standalone hooks keep a stable PATH node symlink" tests/installer/e2e.freshinstall.test.mjs` exited `0`

## Verification

- `node --test tests/installer/e2e.freshinstall.test.mjs`
- `npm test` (126 passed)
- `for f in tests/test_*.js; do node "$f" || exit $?; done` (169 passed)
- `python3 -m unittest discover -s tests -v` (58 passed)
- `node --check cli/install.js`

## Scope

- 2 files changed, +40 / -0 lines

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -444,6 +444,13 @@ function spawnOk(r) {
 }
 
 function absoluteNodePath() {
+  if (!IS_WIN) {
+    try {
+      const r = child_process.spawnSync('/bin/sh', ['-c', 'command -v node'], { encoding: 'utf8' });
+      const candidate = (r.stdout || '').trim().split(/\r?\n/, 1)[0];
+      if (r.status === 0 && candidate) return path.resolve(candidate);
+    } catch (_) {}
+  }
   return process.execPath;
 }
 
--- a/bin/install.js
+++ b/bin/install.js
@@ -1076,7 +1076,7 @@ async function runInit(ctx) {
   if (opts.dryRun) args.push('--dry-run');
   if (opts.force)  args.push('--force');
   if (local && fs.existsSync(local)) {
-    const r = runSpawn(absoluteNodePath(), [local, ...args], null, opts.dryRun);
+    const r = runSpawn(process.execPath, [local, ...args], null, opts.dryRun);
     return spawnOk(r);
   }
   // Curl-pipe fallback
@@ -1087,7 +1087,7 @@ async function runInit(ctx) {
   try {
     const tmp = path.join(os.tmpdir(), `caveman-init-${process.pid}.js`);
     await downloadTo(INIT_SCRIPT_URL, tmp);
-    const r = child_process.spawnSync(absoluteNodePath(), [tmp, ...args], { stdio: 'inherit' });
+    const r = child_process.spawnSync(process.execPath, [tmp, ...args], { stdio: 'inherit' });
     try { fs.unlinkSync(tmp); } catch (_) {}
     return spawnOk(r);
   } catch (e) {
--- a/tests/installer/e2e.freshinstall.test.mjs
+++ b/tests/installer/e2e.freshinstall.test.mjs
@@ -275,11 +275,14 @@ test('fresh install populates hooks dir and settings.json (skipped without `clau
 test('standalone hooks keep a stable PATH node symlink', { skip: process.platform === 'win32' && 'POSIX symlink behavior' }, () => {
   const dir = freshTmpDir();
   const cellarBin = path.join(dir, 'Cellar', 'node', '26.5.0', 'bin');
+  const upgradedCellarBin = path.join(dir, 'Cellar', 'node', '26.8.1', 'bin');
   const stableBin = path.join(dir, 'bin');
   const configDir = path.join(dir, 'claude-config');
   fs.mkdirSync(cellarBin, { recursive: true });
+  fs.mkdirSync(upgradedCellarBin, { recursive: true });
   fs.mkdirSync(stableBin, { recursive: true });
   const cellarNode = path.join(cellarBin, 'node');
+  const upgradedCellarNode = path.join(upgradedCellarBin, 'node');
   const stableNode = path.join(stableBin, 'node');
 
   try {
@@ -300,6 +303,17 @@ test('standalone hooks keep a stable PATH node symlink', { skip: process.platfor
       'hook command should preserve the stable node path found on PATH');
     assert.doesNotMatch(command, new RegExp(cellarNode.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')),
       'hook command must not persist the versioned node target');
+
+    fs.linkSync(process.execPath, upgradedCellarNode);
+    fs.unlinkSync(stableNode);
+    fs.symlinkSync(upgradedCellarNode, stableNode);
+    fs.rmSync(path.join(dir, 'Cellar', 'node', '26.5.0'), { recursive: true });
+
+    const storedNode = command.match(/^"([^"]+)"/)?.[1];
+    assert.equal(storedNode, stableNode, 'hook command should store the stable executable path');
+    const upgraded = spawnSync(storedNode, ['--version'], { encoding: 'utf8' });
+    assert.equal(upgraded.status, 0,
+      `stored hook executable should survive a Node upgrade: ${upgraded.error?.message || upgraded.stderr}`);
   } finally {
     fs.rmSync(dir, { recursive: true, force: true });
   }
--- a/tests/installer/e2e.freshinstall.test.mjs
+++ b/tests/installer/e2e.freshinstall.test.mjs
@@ -272,6 +272,39 @@ test('fresh install populates hooks dir and settings.json (skipped without `clau
   }
 });
 
+test('standalone hooks keep a stable PATH node symlink', { skip: process.platform === 'win32' && 'POSIX symlink behavior' }, () => {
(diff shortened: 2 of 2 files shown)


andrebrait:
Confirmed independently on macOS with Homebrew Node 26.7.0 → 26.8.1.

The installer serialized `process.execPath` as `/opt/homebrew/Cellar/node/26.7.0/bin/node` into the managed `SessionStart` and `UserPromptSubmit` commands. After `brew upgrade node` removed that Cellar version, every prompt reported:

```
UserPromptSubmit failed (exit 127): /opt/homebrew/Cellar/node/26.7.0/bin/node: No such file or directory
```

Replacing the generated executable with the stable `/opt/homebrew/bin/node` symlink made direct `UserPromptSubmit` hook execution exit 0. This confirms #805's mechanism and this PR's approach: installer-time `command -v node` preserves the stable shim, whereas `process.execPath` resolves it to the disposable Cellar target.

One small test-strengthening suggestion: after installation, repoint `stableNode` to a fake 26.8.1 target, remove the 26.5.0 target, then execute the Node path parsed from `settings.json`. That proves the stored hook survives the upgrade, rather than only asserting its string form. The current assertions already cover the root cause.

CBOSSX:
Thanks for the independent confirmation and concrete suggestion. I strengthened the regression in 8242404: it now repoints the stable Node symlink to a simulated 26.8.1 Cellar target, removes the 26.5.0 target, and executes the Node path parsed from settings.json. The focused test and the full 153-test installer suite pass.

bin/install.js
```diff
@@ -444,6 +444,13 @@ function spawnOk(r) {
 }
 
 function absoluteNodePath() {
+  if (!IS_WIN) {
+    try {
+      const r = child_process.spawnSync('/bin/sh', ['-c', 'command -v node'], { encoding: 'utf8' });
+      const candidate = (r.stdout || '').trim().split(/\r?\n/, 1)[0];
+      if (r.status === 0 && candidate) return path.resolve(candidate);
+    } catch (_) {}
+  }
```
AmirF194, line 453: I ran this against `8242404` in a clean container.

`absoluteNodePath()` now shells out to `command -v node` on POSIX (bin/install.js:446-453). That is the right fix for the settings.json hook commands at line 981, which need to survive a future Node upgrade. But it also changes the two immediate, same-session spawns in `runInit()` at lines 1079 and 1090, which run `src/tools/caveman-init.js` (or its curl-pipe fallback) right now, in this process, and were never meant to be PATH-resolved: `process.execPath` there is already guaranteed correct, since it is the exact interpreter currently running the installer.

When PATH's `node` is not the same binary as the one invoking the installer (a stale entry earlier on PATH, a version manager shim, an invocation via an explicit node path), `runInit` now picks up the wrong node and the per-repo IDE rule files never get written, while the installer still reports "done" with no other errors surfaced beyond `caveman-init` in the failed list.

```
$ PATH=/fakebin /usr/local/bin/node bin/install.js --with-init --skip-skills --non-interactive --no-mcp-shrink --config-dir /tmp/cfg
  $ /fakebin/node /repo/src/tools/caveman-init.js /repo
FAKE_NODE_CALLED_WITH: /repo/src/tools/caveman-init.js /repo
  failed:
    • caveman-init — src/tools/caveman-init.js failed
```

Same command against `main` (pre-PR, `process.execPath` unconditionally) runs `/usr/local/bin/node`, the real interpreter, and writes the rule files as expected.

Keeping `absoluteNodePath()` for the persisted hook commands but having the two `runInit` call sites pass `process.execPath` directly would keep the upgrade-survival fix without touching the immediate-execution path it was not meant to change.

JuliusBrussee:
Adopted into #1005 with one change, cherry-picked so your authorship is preserved. Preferring the PATH entry is the right fix for #805, and not symlink-resolving it is correct — realpath would walk back to the Cellar path.

The change: the candidate is located but never exercised. `command -v node` names stale version-manager shims, dangling symlinks and shell functions as readily as real binaries, and baking one of those into `settings.json` breaks every session — strictly worse than the drift it replaces, since `process.execPath` is by construction a working node. It now has to be an absolute path that reports a version before it wins, falling back otherwise. I added a regression test with a present-but-executable-failing PATH node; it fails against the unguarded lookup and passes with the probe.

---

bin/install.js
CBOSSX: Good catch. I restored `process.execPath` for both immediate `runInit` spawns in ca7a629, while keeping `absoluteNodePath()` only for persisted hook commands. I added a regression where PATH points to a failing fake Node while the installer is launched with a different valid interpreter. The focused regression, full installer suite (154/154), and caveman-init suite (8/8) pass.

CBOSSX:
Thanks for adopting the fix into #1005 with authorship preserved. I am closing this PR as superseded. I also left a follow-up on #1005 for the two immediate `runInit()` call sites, with the deterministic regression and correction from `ca7a629`.

JuliusBrussee:
`ca7a629` cherry-picked into #1005 as `96b4759`, authorship preserved. You were right and I had missed it: routing the two `runInit()` spawns through `absoluteNodePath()` broadened a persisted-hook fix into same-process execution. Those two call sites answer a different question — which interpreter runs `caveman-init` *now* — and there `process.execPath` is correct by construction.

Conflict resolution note: `#1005` carries newer scratch-dir handling for the curl-pipe fallback (`privateTmpDir()` plus a `finally` that removes it), so I took that structure with your `process.execPath` change rather than the older `os.tmpdir()` + explicit `unlinkSync` pair. Net behavior is yours; the temp file still gets cleaned up.

I also strengthened your regression fixture. Yours does catch the bug on that branch — I verified both directions — but it catches it because the `MIN_NODE_MAJOR` probe added in #1005 invokes the fake on its way to rejecting it, so the marker records the *probe* rather than `caveman-init` running on the wrong Node. The fake now reports a current version for `--version`, so `absoluteNodePath()` genuinely selects it and the test exercises the real failure path. Verified it fails with either call site reverted and passes with both fixed.

Full installer suite, `caveman-init` suite, Python, Go and repo verification all pass on the merged result.

---

PR #816, opened 2026-08-09
Sync-by-default with opt-in per-session /caveman isolation

eggrollofchaos:
## Motivation

This replaces the closed PR #800, "Per-session mode flag isolation." That PR correctly implemented per-session `/caveman` isolation, but every session isolated **unconditionally** on its own `SessionStart`, before the user ever ran `/caveman <level>` -- silently changing the default behavior for everyone from "all sessions share one toggle" to "every session permanently isolated." That's a bigger, more disruptive default-behavior change than intended, and not something a maintainer should have to infer from "a follow-up PR is coming" -- so it was closed without merging rather than patched in place.

This PR is atomic and self-contained: it restores the shared/synced-by-default behavior caveman has always had, while adding per-session isolation as something a user opts INTO, not something that happens automatically.

## The model

- **Untouched sessions stay synced** -- reading one shared `.caveman-active` flag live, exactly like caveman worked before this PR existed. Verified against upstream's actual pre-scoping mechanism (`git show ec83e5b:src/hooks/caveman-activate.js`) to make sure this isn't a reinvention.
- **`/caveman <level>` isolates** that one session -- it locks to the level you picked and stops following the shared default, even if the shared default later changes to that exact same value.
- **`/caveman default`** (new) reverts an isolated session back to following the shared default live.

## What changed vs. the closed PR's code

The cross-implementation isolation machinery from the closed PR (session-id sanitization, `resolveFlag`/`resolvePrev` ENOENT-vs-rejected semantics, JS/Bash/PowerShell parity, 71 tests) is carried forward as-is -- it already cleared 13 rounds of review. Only three files have new logic:

- `caveman-activate.js`: unified write-target resolution. The write target is now always `resolveFlag(...).path` for every event type (startup, resume, clear, compact) -- not a scoped path computed unconditionally before branching. True startup on a synced session is the only case that refreshes from the configured default; every other case just preserves whatever's already there. This also means an isolated session now correctly stays isolated across a process restart (`claude --resume`), not just mid-conversation re-fires.
- `caveman-parse.js` / `caveman-mode-tracker.js`: new `/caveman default` command.

Developed through several rounds of design review before any code was written -- two real design bugs were caught and fixed at the design stage: the original write-target design would have reintroduced implicit isolation on a synced session's first resume/compact re-fire, and a related bug where an isolated session's emitted rules would diverge from its stored mode across a process restart.

## Test plan

- [x] All existing 71+ tests pass, plus new regression coverage for every case above (`tests/test_hooks.py`, `tests/test_mode_tracker_stdin.js`).
- [x] Manual smoke: two sessions sharing a config dir stay independently isolated; `/caveman default` reverts one back to tracking the shared value live.

Closes nothing directly (#800 already documents the incident); supersedes #800.


eggrollofchaos:
Update (commit `f694844`): addressed review feedback -- 0 Critical/High, 1 Medium, 2 Low, 1 Nit, all fixed.

- Fixed a real inaccuracy in the README's new section (claimed an in-session `/caveman <level>` propagates to other sessions -- it doesn't; only the config file drives the shared default).
- Fixed 2 pre-existing tests that leaked ambient `HOME`/`XDG_CONFIG_HOME` into their subprocess helpers, breaking on any machine with a real caveman config.
- Added a direct unit test for the new `/caveman default` parser grammar.
- Added an explicit guard for opencode's unreachable-but-undocumented `reset` action.

Independent confirmation queued now.

eggrollofchaos:
Review complete at `2d1e3d5`.

- **Tier 1** (DeepSeek, PR code review): 0 Critical, 0 High, 1 Medium, 2 Low, 1 Nit → all fixed → LGTM.
- **Tier 2** (Kimi/Moonshot, independent cross-agent confirmation): 0 Critical/High/Medium, 1 Low (mirror sync, addressed with a local sync commit matching this repo's own established `bdcba4c` precedent) → LGTM.
- **Tier 2 bonus** (Antigravity/Gemini, independent confirmation): 0 findings → LGTM.

Also dropped the internal `docs/plans/` planning doc from the branch (not a convention of this repo) before this final state, matching the earlier PR's own precedent.

`verify_repo.py` and the full local test suite (125+ tests across `test_session_scoping.js`, `test_hooks.py`, `test_mode_tracker_stdin.js`, `test_caveman_parse.js`, `test_mode_tracker.py`, `test_caveman_stats.js`) all pass at this head. CI shows `action_required`, which is the normal GitHub gate on a PR's first run -- happy to have that approved whenever convenient.

Ready for a maintainer's final look.

JuliusBrussee:
Thanks @eggrollofchaos — main has adopted session isolation as the runtime contract and hardened it for concurrency. Closing this alternative sync-by-default design.

PR #817, opened 2026-08-09
fix: remove remaining language attractor tokens

### LINKED ISSUE #812: Language drift: few-shot Portuguese/Spanish examples in SKILL.md leak into output despite fixed-language user instructions"
**What happen**

Caveman rules include a "preserve dominant language" instruction with hardcoded few-shot examples in `skills/caveman/SKILL.md`:

> Preserve user's dominant language. User write Portuguese → reply Portuguese caveman. User write Spanish → reply Spanish caveman.

These example strings sit in the always-on ruleset injected every session (via `caveman-activate.js` on SessionStart, re-read from SKILL.md each time). Even in a conversation that is 100% in another language throughout, the mere presence of "Portuguese"/"Spanish" example tokens in the persistent context can leak into generation and cause the model to suddenly reply in Portuguese or Spanish for one turn, with no non-target-language input anywhere in the conversation. This is especially disruptive for users whose global/project instructions (e.g. `CLAUDE.md`) hard-mandate a single fixed reply language — the fixed-language instruction should always win, but nothing in the rule currently says so.

**Expected**

When a user/project instruction fixes the reply language (e.g. `CLAUDE.md`: "always respond in French"), caveman should never switch away from it — regardless of few-shot example tokens sitting in the ruleset, and regardless of the literal language of the current message. The dominant-language-detection rule should only apply when no such fixed-language instruction exists.

**Before/after example**
Input: (French-only conversation, CLAUDE.md mandates "always respond in French, no exceptions") — "ou as tu trouvé ces infos"
Got: "Nada oficial documentado sobre compressão de imagens no Dokeos..." (Portuguese)
Want: "Pas trouvé — connaissance générale entraînement, pas de recherche web/lecture faite..." (French)

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**
caveman plugin v0.1.0, installed via Claude Code plugin marketplace (`JuliusBrussee/caveman`), commit `0d95a81` ("chore: sync SKILL.md copies")

Mariomarquezt:
Fixes #812.

## Problem

`710173f` de-named the language rule in `skills/caveman/SKILL.md` to remove the `Portuguese`/`Spanish` example tokens that were pulling generation into the wrong language. Two other injected copies of the same rule were not touched and still carry the tokens on `main`:

| File | Injected when |
|---|---|
| `skills/caveman-help/SKILL.md:41` | `/caveman-help` is invoked — the site identified in #812 |
| `src/hooks/caveman-activate.js:154` | SessionStart, on a standalone hook install where `SKILL.md` is not found |

The second one is not mentioned in the report. It is the fallback ruleset, and it is injected at SessionStart exactly like the file `710173f` fixed, so it reproduces the same drift for that install path.

## Change

Both rewordings follow the shape `710173f` used: state the rule without naming any language, and make the "regardless of example text elsewhere" clause explicit.

```diff
-Keep user's language by default. User write Portuguese → reply Portuguese caveman.
+Keep user's language by default — reply in the language user writes, never switch
+regardless of example text or multilingual context elsewhere.
```

## Scope notes

- `caveman-help` is **not** CI-mirrored under `plugins/` (CONTRIBUTING, "not mirrored by CI" table), so the top-level source is the only copy. No sync gap to chase.
- `README.md:137` also names the languages, but it is prose documentation and is never injected into a session, so it is left alone.
- No behaviour change beyond the rule text; no test asserted on the removed strings.

## Verification

Full CI suite run locally on this branch:

- `npm test` — 125 passed, 0 failed
- `tests/test_*.js` — 169 passed, 0 failed across 8 files
- `python -m unittest discover -s tests` — Ran 58 tests, OK

### DIFF
--- a/skills/caveman-help/SKILL.md
+++ b/skills/caveman-help/SKILL.md
@@ -38,7 +38,7 @@ Say "stop caveman" or "normal mode". Resume anytime with `/caveman`.
 
 ## Language
 
-Keep user's language by default. User write Portuguese → reply Portuguese caveman. Compress the style, not the language. Technical terms, code, commands, commit types, and exact error strings stay verbatim unless user ask for translation.
+Keep user's language by default — reply in the language user writes, never switch regardless of example text or multilingual context elsewhere. Compress the style, not the language. Technical terms, code, commands, commit types, and exact error strings stay verbatim unless user ask for translation.
 
 ## Configure Default Mode
 
--- a/src/hooks/caveman-activate.js
+++ b/src/hooks/caveman-activate.js
@@ -151,7 +151,7 @@ if (skillContent) {
     '## Rules\n\n' +
     'Drop: articles (a/an/the), filler (just/really/basically/actually/simply), pleasantries (sure/certainly/of course/happy to), hedging. ' +
     'Fragments OK. Short synonyms (big not extensive, fix not "implement a solution for"). Technical terms exact. Code blocks unchanged. Errors quoted exact.\n\n' +
-    "Preserve user's dominant language. User write Portuguese → reply Portuguese caveman. Compress the style, not the language. Technical terms, code, API names, commands, error strings stay verbatim.\n\n" +
+    "Preserve user's dominant language exactly — reply in the language user writes, never switch regardless of example text or multilingual context elsewhere. Compress the style, not the language. Technical terms, code, API names, commands, error strings stay verbatim.\n\n" +
     'No self-reference. Never name or announce the style. No "caveman mode on" tags. Output caveman-only.\n\n' +
     'Pattern: `[thing] [action] [reason]. [next step].`\n\n' +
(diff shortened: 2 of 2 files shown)


JuliusBrussee:
Adopted into #1015, cherry-picked so your authorship is preserved. Thank you — you found both remaining sites, and the `caveman-activate.js` one matters more than it looks: that ruleset is what a standalone hook install (no `skills/` dir on disk) gets injected on every session, so the earlier `SKILL.md` fix never reached those users at all.

Two things I added on top: `src/hooks/checksums.sha256` had to be regenerated (`verify_repo.py` fails the build otherwise), and a test — `tests/test_hook_language_rule.js` runs the hook with and without `skills/` present and asserts neither ruleset names a language.

It also conflicted on the neighbouring `No self-reference…` line, which was reworded on main since August; I kept main's version and took yours for the language line.

---

JuliusBrussee:

PR #818, opened 2026-08-09
feat(cli): add --doctor to report install health

diegoalvarezf:
## What

Adds `caveman --doctor`: a read-only report on an existing install. It never writes anything, so a broken install is diagnosed here and fixed by re-running the installer or `--uninstall`.

```
$ caveman --doctor
🪨 caveman doctor
  config dir: /home/u/.claude
  install mode: standalone hooks (or plugin not detected)

  problem  missing hook file: caveman-parse.js
  problem  orphaned SessionStart hook points at missing script: /home/u/.claude/hooks/caveman-activate.js
  note     statusline badge not configured

2 problem(s) found
```

## Why

Half-broken installs are hard to diagnose today. A hook file deleted or edited by hand, a `settings.json` that stopped parsing after a manual edit, entries left behind pointing at scripts that no longer exist (#471), or standalone hooks still wired alongside the plugin so every event fires twice (#392). The symptom is a session that crashes at start, double-fires, or silently does nothing, and nothing tells the user which piece is wrong.

## How

Doctor probes for the plugin the same way `installClaude` does (`claude plugin list`) and applies the checks that fit the mode it finds:

- **No plugin detected** — the standalone hook files must all be present, unmodified against the shipped `checksums.sha256`, and wired into `settings.json`.
- **Plugin detected** — the plugin owns that wiring, so missing standalone files are expected; standalone wiring found *on top* of the plugin is the conflict, and gets reported by name.

Orphaned entries are reported in either mode, since migrating between the two paths in either direction can strand them. A user hook whose path merely contains a caveman-like name is never flagged, reusing the exact-basename rule the settings helpers already apply.

Bare `node` invocations and a missing statusline are reported as notes rather than problems: the installer already migrates the former on re-run, and neither breaks anything.

Exits 1 when at least one problem was found, 0 otherwise. Honors `--config-dir` and `--no-color`.

## Tests

19 tests in `tests/installer/doctor.test.mjs` covering both install modes, including a fake `claude` binary on `PATH` so the outcome never depends on what the machine running the suite happens to have installed, and a check that a broken install is left byte-for-byte untouched.

`npm test` passes 144/144 (125 existing + 19 new).

### DIFF
--- a/cli/install.js
+++ b/cli/install.js
@@ -63,7 +63,7 @@ function parseArgs(argv) {
     dryRun: false, force: false, skipSkills: false,
     withHooks: 'auto', withInit: false, withMcpShrink: false,
     all: false, minimal: false, listOnly: false, noColor: false,
-    only: [], uninstall: false, nonInteractive: false,
+    only: [], uninstall: false, doctor: false, nonInteractive: false,
     configDir: null, help: false, always: true,
   };
   for (let i = 0; i < argv.length; i++) {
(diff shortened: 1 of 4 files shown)


JuliusBrussee:
Thanks @diegoalvarezf — install health and repair now ship through `caveman doctor`, covering the current native runtime. Closing this legacy-installer doctor.

PR #821, opened 2026-08-11
fix(stats): add MiniMax model pricing

octo-patch:
Reason: caveman-stats has no output price for MiniMax-M3 or MiniMax-M2.7, so it omits USD savings.

- Add current output-token rates for both model IDs.
- Add regression coverage for both price lookups.
- Generalize the price table documentation.

Checks:
- `node --check src/hooks/caveman-stats.js`
- `node tests/test_caveman_stats.js`
- `npm test`
- `git diff --check`


octo-patch:
I updated the hook integrity manifest for the MiniMax pricing change and pushed commit `36850ba`. I ran `node tests/test_caveman_stats.js`, `npm test`, and the full hook checksum verification.

PR #824, opened 2026-08-12
fix: cross-platform test suites, compression safety, signal forwarding, and hook deduplication

### LINKED ISSUE #742: caveman-shrink never forwards SIGTERM/SIGINT to its spawned upstream server, orphaning it

### LINKED ISSUE #776: caveman-compress: a "compression" that EXPANDS the file passes validation and is committed

### LINKED ISSUE #790: Repo-local `defaultMode: "off"` is ignored by the per-turn hook
# Repo-local `defaultMode: "off"` is ignored by the per-turn hook

**Summary.** A repo-local `.caveman.json` with `{"defaultMode": "off"}` stops
`caveman-activate.js` from injecting the ruleset at session start, but
`caveman-mode-tracker.js` keeps injecting its per-turn reminder anyway. The
documented per-project feature only half-works: the big injection is suppressed,
the per-turn one is not.

## Reproduce

1. Leave the user-level default at `full` (no user config, or `defaultMode: "full"`).
2. Open a session in **project A**, which has no caveman config. `caveman-activate.js`
   writes `$CLAUDE_CONFIG_DIR/.caveman-active` = `full`.
3. In **project B**, add `.caveman.json`:
   ```json
   { "defaultMode": "off" }
   ```
4. Open a session in project B. `caveman-activate.js` correctly emits nothing.
5. Send any prompt in project B.

**Expected:** no caveman context in project B.

**Actual:** `caveman-mode-tracker.js` injects
`CAVEMAN MODE ACTIVE (full). Drop articles/filler/pleasantries/hedging. …`

## Cause

The mode flag is global; the config is per-project; the per-turn hook trusts the
global one.

- `src/hooks/caveman-config.js` documents repo-local config as resolution step 2
  ("lets a team pin a project's default mode") and `getDefaultMode()` implements it.
- `src/hooks/caveman-activate.js:27` honours `off` — *"skip activation entirely,
  don't write flag or emit rules"*. Note it skips **writing** the flag; it never
  **clears** a flag another project set.
- `src/hooks/caveman-mode-tracker.js:16` resolves the flag to one global path,
  `$CLAUDE_CONFIG_DIR/.caveman-active`, with no project component.
- `src/hooks/caveman-mode-tracker.js:193` gates injection on that flag alone.
  `getDefaultMode()` is never consulted on the injection path — it is only called
  when a prompt asks to *activate* caveman.

So any session in another project leaves the flag set, and the next session in the
`off` project inherits it.

## Suggested fix

Let the repo-local config win over the global flag, since the flag cannot express
a per-project answer. In `caveman-mode-tracker.js`, before the injection guard at
line 193:

```js
// Repo-local config wins: the flag is global and cannot express "off here".
if (getDefaultMode() === 'off') return;
```

Alternatively, key the flag by project. The one-line version above is smaller and
preserves the documented precedence order.

## Why a userland workaround isn't enough

The obvious workaround — a project `SessionStart` hook that deletes
`.caveman-active` — is unsafe with concurrent sessions, because the flag is shared
across projects: starting a session in the `off` project silently drops caveman in
whatever other project is running at the time. There is no project-scoped state
available to userland, which is why the fix has to be in the tracker.

## Environment

- caveman 0.1.0 (plugin cache `0d95a81d35a9`)
- Claude Code, macOS
- `CLAUDE_CONFIG_DIR` set to a non-default directory (not required to reproduce —
  the flag path is global either way)

### LINKED ISSUE #793: caveman-stats counts multi-block API responses once per JSONL line — output tokens and turns inflated ~2x on tool-heavy sessions

### LINKED ISSUE #820: caveman-compress: fence indented 4+ spaces breaks inline-code extraction — validation always fails, file never compressible
## Summary

`caveman-compress` can never succeed on a Markdown file that contains a fenced code block indented by 4+ spaces (i.e. a fence nested inside a list item). Validation reports a massive bogus "Inline code lost" error on every attempt, both retries fail, and the file is rolled back with `❌ Compression failed after retries`.

The compressed output itself is fine — the validator is wrong.

## Root cause

`scripts/validate.py`:

```python
FENCE_OPEN_REGEX = re.compile(r"^(\s{0,3})(`{3,}|~{3,})(.*)$")
```

`\s{0,3}` rejects a fence indented by 4 spaces, so `extract_code_blocks()` returns `[]` for it. `extract_inline_codes()` therefore never strips that block, and its backticks leak into:

```python
re.findall(r"`([^`]+)`", text_without_fences)
```

`[^`]` matches newlines, so the leaked fence backticks shift the pairing of every subsequent inline span. Instead of `` `alpha` `` and `` `beta` ``, the extractor returns multi-line prose blobs. `validate_inline_codes()` then compares two differently-misparsed lists and always reports loss.

The 4-space indent here is not indented-code — a fence inside a list item is legal CommonMark, and it is what you get from the natural way to show an example inside a bullet:

```markdown
* Example:
    ```markdown
    x = 1
    ```
```

The existing docstring in `extract_inline_codes` says indentation-awareness was already fixed, but the fix only covers 1–3 spaces.

## Minimal reproduction

`a.md`:

```markdown
# T

* Example:
    ```markdown
    x = 1
    ```
* Use `alpha` and `beta` here.
```

```console
$ python -c "
import sys; sys.path.insert(0,'scripts')
from pathlib import Path
import validate as v
t = Path('a.md').read_text(encoding='utf-8')
print('fences found:', len(v.extract_code_blocks(t)))
print('inline codes:', v.extract_inline_codes(t))
"
fences found: 0
inline codes: ['markdown\n    x = 1\n    ', '\n* Use ', ' and ']
```

Expected: `fences found: 1`, `inline codes: ['alpha', 'beta']`.

Running the full skill on a real file with this shape:

```
Validation attempt 1
❌ Validation failed:
   - Inline code lost: {...several hundred characters of prose per entry...}
Fixing with Claude...

Validation attempt 2
❌ Validation failed:
   - Inline code lost: {...}
❌ Failed after retries — original restored

❌ Compression failed after retries
```

Two full Claude compression calls plus a fix-retry call are spent before giving up, so the failure is not cheap.

## Impact

Any `CLAUDE.md` / memory file that documents an example inside a bullet is permanently uncompressible. That is a common shape for exactly the kind of instruction file this skill targets.

## Suggested fix

Decouple the fence-stripping used for inline-code extraction from the CommonMark indented-code limit. Either relax the leading-whitespace class in a fence-stripping-only pattern (`^\s*` rather than `^\s{0,3}`), or make `extract_inline_codes` fall back to a line-based strip of any line whose stripped form starts with 3+ backticks/tildes.

Constraining the inline-code regex to a single line (`` `([^`\r\n]+)` ``) would also stop one unmatched backtick from swallowing whole paragraphs, and would make the error output readable when a genuine loss does occur.

## Environment

- Windows 11 Pro 26200, Python 3.14.3
- Plugin cache: `~/.claude/plugins/cache/caveman/caveman/309834233183`
- Affected file: `skills/caveman-compress/scripts/validate.py`

### LINKED ISSUE #823: Fix Windows test suite UnicodeDecodeError, fd 0 hook hangs, and POSIX shim test skips
### Problem

Running the test suite on Windows encounters three distinct issues:

1. **Process hang on hook tests**: `tests/test_hooks.py` executes `src/hooks/caveman-activate.js` without setting stdin to `DEVNULL`. On Windows, Node sees `!process.stdin.isTTY` and calls `fs.readFileSync(0, 'utf8')`, blocking indefinitely on the open inherited pipe.
2. **`UnicodeDecodeError` during file reads**: `tests/test_benchmark_contract.py` and `tests/test_compress_safety.py` call `Path.read_text()` without `encoding="utf-8"`, raising `UnicodeDecodeError` when reading UTF-8 characters under Windows `cp1252`.
3. **Hardcoded bash execution in installer tests**: `tests/installer/shim-security.test.mjs` invokes `spawnSync("bash")` directly without checking platform support, failing on Windows systems without bash on PATH.

### Proposed Fix
- Pass `stdin=subprocess.DEVNULL` in `HookScriptTests.run_cmd()` and skip bash-dependent tests if `bash` is not found on PATH.
- Explicitly pass `encoding="utf-8"` to `read_text()` and `write_text()` across all Python test suites.
- Add `{ skip: process.platform === "win32" }` to the bash execution test in `shim-security.test.mjs`.

poorvith-mp:
### Summary

This PR addresses several cross-platform test issues, edge-case validation bugs in `caveman-compress`, child process lifecycle handling in `caveman-shrink`, repo-local configuration resolution in the mode tracker, and token deduplication in `caveman-stats`.

### Changes

1. **Cross-Platform Test Compatibility & Process Safety (fixes #823)**:
   - Added `{ skip: process.platform === "win32" }` to the bash execution test in `tests/installer/shim-security.test.mjs`.
   - Added explicit `encoding="utf-8"` across `tests/test_benchmark_contract.py` and `tests/test_compress_safety.py` to prevent `UnicodeDecodeError` on Windows `cp1252`.
   - Set `stdin=subprocess.DEVNULL` in `tests/test_hooks.py` so child Node processes receive EOF immediately and do not block on fd 0 reads.
   - Skipped POSIX-specific chmod mode assertions on Windows in `test_compress_safety.py`.

2. **Indented Code Fence Validation (fixes #820)**:
   - Updated `FENCE_OPEN_REGEX` in `skills/caveman-compress/scripts/validate.py` (and plugin mirror) from `\s{0,3}` to `\s*` so 4+ space indented code fences in markdown lists are recognized and stripped before inline-code validation.
   - Added `test_four_space_indented_fence_in_list_not_leaked_as_inline` in `tests/test_validate_inline.py`.

3. **Compression Expansion Guard (fixes #776)**:
   - Added check in `skills/caveman-compress/scripts/compress.py` (and plugin mirror) to abort and preserve the original file if the compressed body character count exceeds the original body size.
   - Configured `sys.stdout.reconfigure(encoding="utf-8", errors="replace")` to prevent Windows `UnicodeEncodeError` when logging status indicators.
   - Added `test_expanded_compressed_output_rejected_and_aborted` in `tests/test_compress_safety.py`.

4. **Process Signal Forwarding (fixes #742)**:
   - Added `SIGTERM` and `SIGINT` signal handlers in `src/mcp-servers/caveman-shrink/index.js` to terminate the spawned `upstream` MCP process cleanly on host shutdown.

5. **Repo-Local Config Binding (fixes #790)**:
   - Passed `{ getDefaultMode: () => getDefaultMode(data.cwd), skipNaturalLanguage }` to `parseModeChange` in `src/hooks/caveman-mode-tracker.js` so `.caveman.json` (`defaultMode: "off"`) in the target workspace is respected.

6. **Multi-Block Message Deduplication (fixes #793)**:
   - Tracked `seenMessageIds` in `src/hooks/caveman-stats.js` `parseSession()` to prevent multiple JSONL log entries sharing a message ID from inflating turn counts and token totals.
   - Added unit test in `tests/test_caveman_stats.js`.

### Verification

All test suites pass:
- `node --test tests/installer/*.test.mjs` (145 passed, 0 failed, 7 skipped)
- `python -m unittest discover -s tests -v` (64 passed, 0 failed, 4 skipped)
- `node tests/test_caveman_stats.js` (47 passed, 0 failed)
- `node tests/test_mode_tracker_stdin.js` (11 passed, 0 failed)

### DIFF
--- a/plugins/caveman/skills/caveman-compress/scripts/compress.py
+++ b/plugins/caveman/skills/caveman-compress/scripts/compress.py
@@ -355,6 +355,11 @@ def compress_file(filepath: Path) -> bool:
         print("   already in caveman form. Original file is untouched (no backup created).")
         return False
 
+    if len(compressed_body.strip()) > len(body.strip()):
+        print(f"❌ Compression aborted: output expanded ({len(compressed_body.strip())} > {len(body.strip())} bytes).")
+        print("   Original file is untouched (no backup created).")
+        return False
+
     # Reassemble: frontmatter (verbatim) + compressed body
     compressed = frontmatter + compressed_body
 
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -355,6 +355,11 @@ def compress_file(filepath: Path) -> bool:
         print("   already in caveman form. Original file is untouched (no backup created).")
         return False
 
+    if len(compressed_body.strip()) > len(body.strip()):
+        print(f"❌ Compression aborted: output expanded ({len(compressed_body.strip())} > {len(body.strip())} bytes).")
+        print("   Original file is untouched (no backup created).")
+        return False
+
     # Reassemble: frontmatter (verbatim) + compressed body
     compressed = frontmatter + compressed_body
 
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -172,7 +172,10 @@ process.stdin.on('end', () => {
     // Shared mode-change parser (#602) — single source of truth with the
     // opencode plugin for slash commands, namespaced /caveman:caveman-*,
     // natural-language activation/deactivation, and brevity triggers.
-    const change = parseModeChange(prompt, { getDefaultMode, skipNaturalLanguage });
+    const change = parseModeChange(prompt, {
+      getDefaultMode: () => getDefaultMode(data.cwd),
+      skipNaturalLanguage
+    });
 
     // A /caveman argument that resolves to no mode used to leave the level
     // untouched and say nothing, so a typo or punctuation glued to the level
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -152,11 +152,22 @@ function parseSession(filePath) {
   let turns = 0;
   let model = null;
   const messages = []; // per-message {ts, outputTokens} for mode attribution (#601)
+  const seenMessageIds = new Set();
+
   for (const line of raw.split('\n')) {
     if (!line.trim()) continue;
     let entry;
     try { entry = JSON.parse(line); } catch { continue; }
     if (entry.type !== 'assistant' || !entry.message) continue;
+
+    // Claude Code JSONL logs can contain multiple lines per assistant turn (multi-block responses).
+    // Deduplicate by message ID / uuid so tokens and turns aren't inflated (issue #793).
+    const msgId = entry.message.id || entry.uuid || null;
+    if (msgId) {
+      if (seenMessageIds.has(msgId)) continue;
+      seenMessageIds.add(msgId);
+    }
+
     const usage = entry.message.usage;
     if (!usage) continue;
     outputTokens    += usage.output_tokens           || 0;
--- a/src/mcp-servers/caveman-shrink/index.js
+++ b/src/mcp-servers/caveman-shrink/index.js
@@ -185,3 +185,14 @@ upstream.stdin.on('error', err => {
 });
 process.stdin.on('data', forwardInput);
 process.stdin.on('end', endInput);
+
+// Forward termination signals to upstream process so it does not get orphaned (issue #742)
+const forwardSignal = signal => {
+  if (upstream && !upstream.killed) {
+    try {
+      upstream.kill(signal);
+    } catch (_) {}
+  }
+};
+process.on('SIGTERM', () => forwardSignal('SIGTERM'));
+process.on('SIGINT', () => forwardSignal('SIGINT'));
--- a/tests/test_caveman_stats.js
+++ b/tests/test_caveman_stats.js
@@ -54,6 +54,21 @@ test('reads --session-file directly and sums output tokens', (tmp) => {
   assert.match(out, /Cache-read tokens:\s+250/);
 });
 
+test('deduplicates multi-block assistant entries with the same message id (issue #793)', (tmp) => {
+  const sess = makeSession(tmp, [
+    { type: 'assistant', message: { id: 'msg_123', usage: { output_tokens: 100, cache_read_input_tokens: 200 } } },
+    { type: 'assistant', message: { id: 'msg_123', usage: { output_tokens: 100, cache_read_input_tokens: 200 } } },
+    { type: 'assistant', message: { id: 'msg_456', usage: { output_tokens: 50, cache_read_input_tokens: 50 } } },
+  ]);
+  const out = execFileSync(process.execPath, [STATS, '--session-file', sess], {
+    encoding: 'utf8',
+    env: { ...process.env, CLAUDE_CONFIG_DIR: path.join(tmp, '.claude') },
+  });
+  assert.match(out, /Turns:\s+2/);
+  assert.match(out, /Output tokens:\s+150/);
+  assert.match(out, /Cache-read tokens:\s+250/);
+});
+
 test('shows full-mode savings estimate when flag is full', (tmp) => {
   const sess = makeSession(tmp, [
     { type: 'assistant', message: { usage: { output_tokens: 350 } } },
--- a/tests/test_compress_safety.py
+++ b/tests/test_compress_safety.py
@@ -29,44 +29,56 @@ class CompressSafetyTests(unittest.TestCase):
         return path
 
     def test_empty_input_refused(self):
-        with tempfile.TemporaryDirectory() as tmp:
+        with tempfile.TemporaryDirectory() as tmp, \
+             tempfile.TemporaryDirectory() as data_home, \
+             mock.patch.dict(os.environ, {"XDG_DATA_HOME": data_home, "LOCALAPPDATA": data_home}):
             path = self._file_with(Path(tmp), "")
             with mock.patch.object(compress_mod, "call_claude") as call:
                 ok = compress_mod.compress_file(path)
             self.assertFalse(ok)
             call.assert_not_called()
             self.assertEqual(path.read_text(encoding="utf-8"), "")
-            self.assertFalse((Path(tmp) / "task.original.md").exists())
+            backup = compress_mod.backup_dir_for(path.resolve()) / "task.original.md"
+            self.assertFalse(backup.exists())
 
     def test_empty_compressed_output_does_not_touch_disk(self):
-        with tempfile.TemporaryDirectory() as tmp:
+        with tempfile.TemporaryDirectory() as tmp, \
+             tempfile.TemporaryDirectory() as data_home, \
+             mock.patch.dict(os.environ, {"XDG_DATA_HOME": data_home, "LOCALAPPDATA": data_home}):
             original = "# Heading\n\nSome long natural language paragraph that should be compressed.\n"
             path = self._file_with(Path(tmp), original)
             with mock.patch.object(compress_mod, "call_claude", return_value=""):
                 ok = compress_mod.compress_file(path)
             self.assertFalse(ok)
             self.assertEqual(path.read_text(encoding="utf-8"), original)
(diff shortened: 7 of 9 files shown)


JuliusBrussee:
Thanks @poorvith-mp — safe pieces landed independently and current architecture has moved on; remaining mixed changes require separate current PRs. Closing as superseded.

PR #833, opened 2026-08-12
fix(hooks): Windows stdin EOF stall silently kills the UserPromptSubmit hook

adem1k3:
## Problem

On Windows, the `caveman-mode-tracker.js` hook's stdin pipe sometimes never signals EOF. All of the tracker's work runs in stdin's `'end'` handler, so when that happens the process sits idle until the hook timeout kills it — the mode flag is never updated and the per-turn caveman reinforcement is silently lost. The user just sees Claude drift out of caveman mode.

**Evidence** (Windows 11, Claude Code, caveman installed as plugin): across one machine's transcripts, 28 of 53 hook runs died at the 5s timeout with median 5423ms wall clock — while `node` itself starts in ~130ms on the same machine. The cost was waiting for an EOF that never came, not computing. After this patch ran locally for ~2 weeks: daily successful injections, zero timeouts.

## Fix

- 250ms stdin grace timer: Claude Code writes the whole JSON payload up front, so process whatever is buffered once the grace period passes instead of waiting for EOF. A healthy `'end'` fires in single-digit milliseconds and clears the timer, so the normal path is unchanged.
- `handled` guard so `'end'` and the timer can never both run; work extracted to `handleInput()`/`runHook()`.
- `process.stdin.destroy()` afterwards so the event loop drains and the process exits on its own — deliberately not `process.exit()`, which can truncate async pipe writes on Windows.
- Both hook timeouts raised 5s → 15s: SessionStart legitimately takes 3–5s wall clock on Windows when session start spawns several MCP servers concurrently and antivirus scans each node spawn, putting it within margin of the old 5s limit.

## Tests

New regression test in `tests/test_mode_tracker_stdin.js` holds the child's stdin open (no EOF, no close) and asserts the buffered `/caveman ultra` is processed with a clean self-exit. It fails against the previous tracker (watchdog SIGTERM after hang) and passes with the fix. All 12 tests in that file pass; `tests/test_mode_tracker.py` passes (23 tests). The two failures in `tests/test_hooks.py` are pre-existing on main (Windows path-separator mismatch in statusline install tests, unrelated).

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -12,7 +12,7 @@
           {
             "type": "command",
             "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\"",
-            "timeout": 5,
+            "timeout": 15,
             "statusMessage": "Loading caveman mode..."
           }
         ]
@@ -24,7 +24,7 @@
           {
             "type": "command",
             "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-mode-tracker.js\"",
-            "timeout": 5,
+            "timeout": 15,
             "statusMessage": "Tracking caveman mode..."
           }
         ]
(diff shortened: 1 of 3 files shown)


adem1k3:
Closing this — I've since stopped using the plugin, so I won't be around to iterate on review feedback. The Windows stdin EOF stall is real and the regression test in this branch reproduces it, so feel free to reuse any of this if it's useful.

PR #838, opened 2026-08-13
fix(hooks): unresolved /caveman arg is silent; prose fires NL triggers

bb25see-hash:
Rebased onto `c72984e` and rewritten against `src/hooks/caveman-parse.js` — the first version of this PR targeted the pre-#602 tracker. Both defects survive on current main.

## 1. Punctuation glued to the mode

`/caveman ultra; still too verbose` parses the argument as `ultra;`, matches no mode, leaves the level untouched, says nothing.

Only `parts[1]` is ever read, so extra *words* after the mode were already harmless. It is the character glued to the mode that breaks it.

Fix: strip trailing punctuation. An argument that still resolves to nothing now returns `{ action: 'unresolved' }` instead of `null`, and the tracker turns that into a notice naming the valid modes.

Notice is built from `VALID_MODES` minus `INDEPENDENT_MODES`, so it cannot drift from the parser. It never echoes the rejected argument — that string lands in model context and is untrusted input.

`unresolved` is additive: `applyModeChange` in the opencode plugin acts only on `set`/`clear`. Three existing tests asserted `null` for a bogus level; they now assert the new verdict and still guard the original danger — a bogus level must never activate the default.

## 2. Prose that quotes the trigger phrases fires them

Natural-language triggers run over the whole prompt. A bug report quoting the `/caveman-help` card's own line —

> Say "stop caveman" or "normal mode".

— switched caveman off mid-task. #537's envelope unwrap closes the slash-command path; this is the prose path, which needs no envelope. Any pasted changelog, issue text, or doc carrying those words does the same.

Fix: triggers require a short, command-like utterance — 120 chars or fewer, not slash-initiated. Slash parsing untouched and never length-capped, so `/caveman off` works inside a prompt of any size. 120 clears every documented form with headroom; the longest, "back to normal mode please", is 26.

## Tests

9 new cases in `tests/test_caveman_parse.js` plus 2 parity cases, and 5 integration cases in `tests/test_mode_tracker.py`.

Verified against pristine files first: 9 of the new JS cases and 4 of the 5 python cases fail on unmodified `c72984e`. The one that passes both ways is the short `stop caveman` control — it is there to prove the silence in the negative cases is the scoping and not the harness.

Suites green: 44 parse, 28 mode-tracker, 11 stdin.

## Note

`src/hooks/checksums.sha256` still carries the old hashes for the two changed hook files. Manifest looks per-release, so leaving it to your release cut rather than guessing — say the word if you want it in this PR.

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -6,7 +6,7 @@ const fs = require('fs');
 const path = require('path');
 const os = require('os');
 const { execFileSync } = require('child_process');
-const { getDefaultMode, safeWriteFlag, readFlag, recordModeChange } = require('./caveman-config');
+const { getDefaultMode, safeWriteFlag, readFlag, recordModeChange, VALID_MODES } = require('./caveman-config');
 const { parseModeChange, INDEPENDENT_MODES } = require('./caveman-parse');
 
 const claudeDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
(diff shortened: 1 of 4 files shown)


JuliusBrussee:
Thanks for reporting and fixing both parser defects. Unresolved slash arguments and quoted-prose trigger handling are now covered on main through [da92907f](https://github.com/JuliusBrussee/caveman/commit/da92907f), shipped in [v2.1.0](https://github.com/JuliusBrussee/caveman/tree/v2.1.0), with explicit regression cases. I’m closing this PR as superseded, not rejected.

PR #839, opened 2026-08-13
feat(hooks): level-aware per-turn reinforcement

bb25see-hash:
**Stacks on #838.** Rebased onto `c72984e`; the diff shows #838's commit too until it merges. Review the second commit only.

## Problem

Switching level mid-session moves the statusline badge and changes nothing about the output.

Two causes:

1. Every level injects the same per-turn line — `CAVEMAN MODE ACTIVE (<level>) — session ruleset applies.` Only the label differs, so `lite`, `full`, `ultra` and the three wenyan variants are indistinguishable to the model.
2. The level-specific ruleset is emitted only by the SessionStart hook, so only for whatever level the session opened at. `/caveman ultra` on turn 40 never re-emits ultra's rules.

Net effect: a user switches to ultra, sees the badge change, sees output stay the same, reports the feature as broken. That is exactly how this was found.

## Change

Per-level directive lines, drawn from the intensity table in `skills/caveman/SKILL.md`.

On a level CHANGE, re-emit that level's filtered rows once. Ordinary turns still cost one short line. The read is best-effort with the same candidate paths the SessionStart hook uses, and returns empty when no skills dir is present, so a standalone hook install never breaks on it.

## Also: the exemptions, restated per turn

The reminder now names what stays verbatim — code, commit messages, security warnings, error strings, and fixed-format blocks required by whatever else is instructing the model (protocol blocks, gates, checklists, audit blocks).

This is where caveman loses in practice. Another instruction set demands structured multi-line output, caveman demands compression, and the model resolves the clash by dropping one of them. They govern different things; a required block is a structured artifact like a code block, and the prose around it still compresses.

Wording is generic — no host-specific tooling named. Happy to drop this half if you would rather keep the reminder purely about prose.

## Tests

6 new cases in `tests/test_mode_tracker.py`: levels inject different text, a switch re-emits that level's row and no other level's, an ordinary turn does not re-emit, re-switching to the same level does not re-emit, the exemption clause is present, and `wenyan` reports its canonical `wenyan-full` label.

4 of the 6 fail against the parent commit. The 2 that pass are absence-assertions whose positive control is the re-emission test.

Suites green: 34 mode-tracker, 44 parse, 11 stdin.

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -25,6 +25,72 @@ function removeFlag(path) {
   }
 }
 
+// Per-level directives. Every level used to inject one identical sentence, so
+// `/caveman ultra` mid-session moved the statusline badge and changed nothing
+// about the output — the level-specific rules are emitted ONLY by the
+// SessionStart hook, i.e. only for whatever level the session opened at.
+const LEVEL_DIRECTIVE = {
+  lite: 'Cut filler, pleasantries and hedging. Keep articles and whole sentences.',
+  full: 'Drop articles, filler, pleasantries, hedging. Fragments OK. ' +
+        'Short synonyms. No tool-call narration, no decorative tables.',
+  ultra: 'Fragments only. One word where one word carries it. Each fact once. ' +
+         'Drop conjunctions where cause-then-effect stays unambiguous. ' +
+         'No invented abbreviations, no arrows — both cost tokens and clarity.',
+  'wenyan-lite': 'Semi-classical Chinese register. Drop filler and hedging, keep grammar.',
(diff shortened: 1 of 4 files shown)


JuliusBrussee:
**Likely superseded — worth a check from you before anyone closes it.**

`main` now does level-aware reinforcement natively: `caveman-mode-tracker.js` carries a `REINFORCEMENT_RULES` table keyed by level, and a level *switch* additionally re-injects that level's filtered `SKILL.md`. Driving the hook directly, `ultra` and `lite` emit different rule text, not just a different name in the banner.

I did not diff your branch against that line by line, so I'm not claiming it's fully covered. If your PR still carries something the landed version misses, name that part and I'll look next run.

---

PR #849, opened 2026-08-15
fix(hooks): degrade instead of crashing when a sibling hook file is missing

### LINKED ISSUE #848: SessionStart/UserPromptSubmit hooks hard-crash with MODULE_NOT_FOUND when a sibling hook file is missing — bare top-level require('./caveman-config') in 3 of 4 entrypoints
Every session start prints:

```
Failed with non-blocking status code: node:internal/modules/cjs/loader:1408
```

Root cause is a missing sibling file (`src/hooks/caveman-config.js`) in the installed plugin dir. But the reason it surfaces as an opaque loader stack trace instead of a diagnosable message is a code issue in caveman: **3 of the 4 hook entrypoints require `./caveman-config` at top level with no guard**, so any missing/unreadable sibling turns into an uncaught `MODULE_NOT_FOUND` and exit 1 on *every* SessionStart and *every* UserPromptSubmit.

## Reproduction

```console
$ echo '{"session_id":"t","cwd":"/tmp","hook_event_name":"SessionStart","source":"startup"}' \
    | node ~/.claude/plugins/cache/caveman/caveman/27d5a3981a34/src/hooks/caveman-activate.js
node:internal/modules/cjs/loader:1408
  throw err;
  ^

Error: Cannot find module './caveman-config'
Require stack:
- .../src/hooks/caveman-activate.js
    at Function._resolveFilename (node:internal/modules/cjs/loader:1405:15)
    ...
    at Object.<anonymous> (.../src/hooks/caveman-activate.js:12:84)
  code: 'MODULE_NOT_FOUND'
$ echo $?
1
```

## The code issue

`src/hooks/caveman-activate.js` L12 — mandatory dependency, unguarded:

```js
const { getDefaultMode, safeWriteFlag, recordModeChange, readFlag, VALID_MODES } = require('./caveman-config');
```

Same shape in `src/hooks/caveman-mode-tracker.js` L9 and `src/hooks/caveman-stats.js` L13.

`src/hooks/caveman-parse.js` L43-48 already does the right thing for the identical import:

```js
let cavemanConfig;
try {
  cavemanConfig = require('./caveman-config');
} catch (e) {
  cavemanConfig = require('./caveman-config.cjs');
}
```

And `caveman-activate.js` itself demonstrates the fail-soft philosophy 18 lines further down (L30-33), for the *optional* dep:

```js
// Best-effort: any error is swallowed so SessionStart is never blocked.
try {
  const { applyOverrides, resolvePluginRoot } = require('./cavecrew-model-overrides');
  applyOverrides(resolvePluginRoot(__dirname));
} catch (e) {}
```

So the codebase already treats "a sibling hook file may be absent" as a real risk — it just guards one of the four call sites. This is the same failure class as #801 (`install.sh`/`install.ps1` omitting `caveman-parse.js`); that fix added the file to the copy list, but the underlying "top-level require ⇒ hard crash" remains, so the next missing file reproduces it.

## How the file went missing

I could not attribute the deletion, and I am **not** claiming caveman deleted it — reporting the observed state in case it is a known install-path problem:

The installed plugin dir is a git worktree at the exact HEAD commit, and git reports the file as deleted from the worktree only:

```console
$ git -C ~/.claude/plugins/cache/caveman/caveman/27d5a3981a34 status --porcelain
 D src/hooks/caveman-config.js
 D src/mcp-servers/caveman-shrink/README.md
 D src/mcp-servers/caveman-shrink/compress.js
 D src/mcp-servers/caveman-shrink/index.js
 D src/mcp-servers/caveman-shrink/package.json
 D src/mcp-servers/caveman-shrink/spawn-options.js
 D src/rules/caveman-activate.md
 D src/rules/caveman-openclaw-bootstrap.md
?? .in_use/
```

Exactly 8 tracked files, all under `src/`, all present at HEAD (`git ls-tree HEAD src/hooks/` lists `caveman-config.js`, blob `d94131b`, and the blob is present in `.git/objects`). Not sparse-checkout (`fatal: this worktree is not sparse`), not gitignored (`git check-ignore` exits 1 for all three paths). The marketplace clone at `~/.claude/plugins/marketplaces/caveman` has all 8 files. Other cached version dirs for the same 2.0.0 release (`613d7f0402fb`, `a0109974ea32`, `c72984e4392c`, `099327780ef6`) all have `caveman-config.js`; only the current `27d5a3981a34` does not. `git status` on this worktree shows no other drift, so it is not a broad truncation.

Note that `src/hooks/checksums.sha256` **already lists** `caveman-config.js` — the manifest that would have caught this ships with the plugin, but nothing verifies it at runtime.

Workaround that fixed it locally:

```console
$ git -C ~/.claude/plugins/cache/caveman/caveman/27d5a3981a34 restore src/hooks/caveman-config.js src/mcp-servers src/rules
$ echo '{"session_id":"t","cwd":"/tmp","hook_event_name":"SessionStart","source":"startup"}' \
    | node .../src/hooks/caveman-activate.js
...
$ echo $?
0
```

## Suggested fix

1. Guard the three unguarded requires the way `caveman-parse.js` already does — a hook must never exit non-zero because a sibling is missing. On failure, degrade to the built-in fallback ruleset (as #792 describes for the SKILL.md lookup) rather than crashing.
2. Emit one actionable line on stderr instead of a Node stack trace, e.g. `caveman: plugin install incomplete (missing caveman-config.js) — run /plugin update caveman`.
3. Optionally verify `src/hooks/checksums.sha256` on SessionStart (cheap, manifest already shipped) so a partial install self-reports instead of failing at a random require.

## Environment

- Claude Code 2.1.233, plugin install path (`.claude-plugin/plugin.json` hooks), not standalone `install.sh`
- caveman 2.0.0 @ `27d5a3981a34` ("README: whole-cave table gains status column, Browse row, frozen tier")
- Node v23.11.0 (Homebrew), macOS 26.5.1 (arm64)

Related: #801 (same failure class, standalone installer), #792 (missing-file path degrading silently), #78 / #471 (same loader error, different root cause — path resolution / orphaned hooks).

rebel-daekyeong:
Fixes #848.

`caveman-activate.js`, `caveman-mode-tracker.js` and `caveman-stats.js` require `./caveman-config` at module top level with no guard. An install that is missing that one file therefore produces an uncaught `MODULE_NOT_FOUND` — a raw Node stack trace and exit 1 on **every** session start and **every** prompt, which Claude Code surfaces only as:

```
Failed with non-blocking status code: node:internal/modules/cjs/loader:1408
```

`caveman-parse.js` L43-48 already wraps the identical import in try/catch, and `caveman-activate.js` already treats the *optional* `cavecrew-model-overrides` require as best-effort — "any error is swallowed so SessionStart is never blocked". The risk was already understood; it just was not applied to the mandatory dependency. This is the same failure class as #801, whose copy-list fix landed but left the crash shape intact.

## What changes

A small `requireSibling()` in each of the three entrypoints, plus an explicit degraded path:

| entrypoint | behavior when a sibling will not load |
|---|---|
| `caveman-activate.js` | Emits the built-in fallback ruleset it already uses when `SKILL.md` is unresolvable, and exits 0. A degraded session still gets rules; only flag persistence is lost (no flag write, no mode log, `readFlag()` reports nothing active). `CAVEMAN_DEFAULT_MODE=off` still opts out — it is the highest-priority source in `getDefaultMode()` and the only one reachable without the config module. |
| `caveman-mode-tracker.js` | Emits nothing, exits 0. Drains stdin before exiting so the parent's payload write does not turn into a broken pipe (#397). |
| `caveman-stats.js` | Prints one actionable line and exits non-zero, instead of a stack trace that the calling hook then reports as an unexplained failure. |

Message shape:

```
caveman: caveman-config.js is missing from /path/to/src/hooks — the install is incomplete.
Run `/plugin update caveman`, or rerun install.sh for standalone hooks. Continuing with
reduced functionality.
```

Two details worth reviewing:

- **Missing vs. broken is distinguished.** A `MODULE_NOT_FOUND` thrown from *inside* a sibling that itself loaded fine is reported as `could not load … — Cannot find module './x'`, not as an incomplete install, so the message never points at the wrong cause. There is a test for exactly this.
- **Only the first line of `error.message` is echoed.** Node appends a multi-line `Require stack:` block to it; echoing that back reprinted the very noise the guard removes. My first cut got this wrong and the new test caught it.

## Deliberate call, open to preference

The loader is **duplicated per entrypoint rather than extracted into a shared module**. It is the guard that makes a missing sibling survivable, so making it a sibling would reintroduce the failure it prevents. That costs ~15 duplicated lines across three files. Happy to fold it into `caveman-config.js` with an inline fallback if you would rather have one copy — say the word.

`src/hooks/checksums.sha256` is regenerated for the three touched files (`tests/verify_repo.py` verifies the manifest).

## Tests

New `tests/test_hook_missing_sibling.js`, in the style of `test_mode_tracker_stdin.js` — builds a temp hook dir holding real copies of every hook file minus the one under test, so the real `src/hooks` siblings can never satisfy a require the test is trying to break. 8 cases: degraded activate / mode-tracker (config and parse) / stats, the missing-vs-broken discrimination, the nested-`MODULE_NOT_FOUND` case, and controls proving the healthy path still writes the flag, still reinforces, and warns about nothing.

Every case asserts the output contains neither `Require stack:` nor `node:internal/modules/cjs/loader` — the exact shape users are reporting.

Full suite, macOS + Node v23.11.0:

```
npm test                                   → 156 pass, 0 fail
for f in tests/test_*.js; do node "$f"; done → 9 files, all pass (new file 8/8)
python3 -m unittest discover -s tests      → Ran 62 tests — OK
node --check (all verify_repo.py targets)  → clean
```

## Verified against the real failure

Reproduced from an actual broken install (plugin cache worktree at `27d5a39` with `src/hooks/caveman-config.js` absent): before the patch, exit 1 and the loader stack trace on every session start; after, exit 0 with the ruleset emitted and one stderr line naming the file and the remedy.


JuliusBrussee:
Thanks for adding missing-sibling degradation. Equivalent handling landed on main in [0ded04bf](https://github.com/JuliusBrussee/caveman/commit/0ded04bf), shipped in [v2.1.0](https://github.com/JuliusBrussee/caveman/tree/v2.1.0), and the clean-main suite passes all 18 degradation cases. I’m closing this PR as superseded, not rejected.

PR #857, opened 2026-08-17
chore: sync SKILL.md copies [skip ci]

fg-kjh0510:


JuliusBrussee:
Thanks for keeping the generated skill copy synchronized. Generated mirrors are workflow-owned, and the current sync is already on main in [ead4ab8](https://github.com/JuliusBrussee/caveman/commit/ead4ab8). There is no independent source change left to merge, so I’m closing this generated PR as superseded.

PR #867, opened 2026-08-18
fix(plugin): concat reinforcement onto last system message (fixes vLLM Qwen3 400)

emilstahl:
## Problem

The opencode plugin's `experimental.chat.system.transform` hook pushes a **new** system message per turn:

```js
// src/plugins/opencode/plugin.js:187 (before)
output.system.push(reinforcementLine(active));
```

vLLM-hosted Qwen3 models use a strict chat template that rejects any system message not at index 0:

```
litellm.BadRequestError: Hosted_vllmException -
{"message":"System message must be at the beginning.","type":"BadRequest","code":400}
```

The fallback chain (Qwen3-6-27b → Qwen3.8-27B) fails identically — both share the strict template.

## Fix

Concat onto the last system entry instead of pushing a new one (same safe pattern [`@dietrichgebert/ponytail`](https://github.com/dietrichgebert/ponytail) already uses):

```js
const line = reinforcementLine(active);
if (output.system.length > 0) {
  output.system[output.system.length - 1] += '\n\n' + line;
} else {
  output.system.push(line);
}
```

Preserves single system message, keeps per-turn reinforcement.

## Verification

Tested with `opencode run "say hi"` per model, caveman in `full` mode (default), no model pinning:

| Model | Before | After |
|---|---|---|
| Qwen3-6-27b   | 400 Bad Request | OK |
| Qwen3.8-27B   | 400 Bad Request | OK |
| Qwen3.6-35B-A3B | Hangs indefinitely | OK |
| Qwen3.5-397B-A17B-AWQ | OK | OK |
| GLM-5.2 | OK (tolerant) | OK |

## Root cause analysis

The `push()` pattern assumes the model's chat template tolerates multiple system messages interleaved with user/assistant turns. This holds for Anthropic/OpenAI-style APIs and GLM-5.2, but not for vLLM deployments using Qwen3's strict chat template (which enforces system-at-index-0).

The `ponytail` plugin's transform already uses the concat-onto-last pattern and has never triggered this issue — confirming the fix is safe across providers.

## Checklist

- [x] Code follows existing style
- [x] Verified across 5 models (4 Qwen3 variants + GLM-5.2)
- [x] No new dependencies
- [x] No breaking changes — behavior identical when `output.system` is empty (falls back to `push`)


AmirF194 (commented):
I checked this against current main (a42ef76) instead of just the diff.

This exact fix already landed on main via commit 5429b5d, "fix(xplat): stop hooks stalling on Windows; own only our own config entries", in a form that goes further than this PR: it also guards against opencode ever reusing the `output.system` array across turns, by replacing an existing "CAVEMAN MODE ACTIVE (...)" line in place instead of always appending to the last entry. This PR's version does not have that guard, so if `output.system` is ever reused across turns it would keep concatenating onto the same string turn after turn.

Confirmed with a real merge, not just a read: main is a strict superset of this diff, and merging this branch as-is now conflicts at the exact hunk.

```
$ git fetch origin pull/867/head:pr-867 && git checkout main && git merge --no-commit --no-ff pr-867
Auto-merging src/plugins/opencode/plugin.js
CONFLICT (content): Merge conflict in src/plugins/opencode/plugin.js
Auto-merging tests/installer/opencode.test.mjs
CONFLICT (content): Merge conflict in tests/installer/opencode.test.mjs
```

GitHub's own mergeStateStatus agrees: DIRTY / CONFLICTING.

This PR can be closed as superseded, or rebased onto main and dropped entirely since main's version already covers everything here.

JuliusBrussee:
Thanks @emilstahl — the system-message concatenation fix is on main with regression coverage. Closing as implemented.

PR #869, opened 2026-08-19
engine: unwrap agent <file> wrappers before content detection

PeterShanxin:
## Problem

Coding agents wrap file references handed to the model. pi (pi.dev) turns a `@file` argument into `<file name="...">...</file>`; other agents use the same single-element wrapper convention.

Left in place, the wrapper defeats `Detect`: the payload no longer starts with its real first bytes, so Go source never parses (`code_nocgo.go` only accepts files `go/parser` can parse), JSON never validates, and everything falls through to the `text` compressor, which compresses nothing. Result: pass-through at ratio 0 for every wrapped file the agent sends.

Verified on Windows 11 ARM64 with a real agent: pi sends a 142,409-byte `<file>`-wrapped Go payload; the no-cgo build's code compressor returned ratio 0 → the request went upstream unchanged.

## Fix

`unwrapFileWrapper` is deliberately conservative, mirroring the `listing.go` gutter approach:
- wrapper must be the outermost element
- must open with the literal `<file` tag (attributes allowed)
- exactly one `<file` and exactly one `</file>` (a stray tag inside a string literal = not a plain wrapper)
- must contain non-whitespace content
- anything else returns the input unchanged (no-op wrapper, no caller branches)

`Compress` and `Simulate` unwrap before `Detect` and rewrap compressed output so the model still sees the file reference the agent supplied. Original bytes remain in the CCR store unchanged.

## Tests

- `TestUnwrapFileWrapper`: detection + rewrap round-trip, plus negative cases (plain text, bare Go, HTML, multi-element, truncated wrappers).
- Full `go test ./engine/...` passes.

## End-to-end measurement (real agent)

pi coding agent → caveman proxy → DeepSeek, same task A/B:
- record mode (pass-through): 64,310 input tokens
- compress mode with this fix: 18,478 input tokens (**-71.3%**), answer exact-correct

## License

Engine directory change; signed per DCO.

### DIFF
--- a/engine/detect.go
+++ b/engine/detect.go
@@ -41,12 +41,9 @@ var (
 // fail-open: anything it is not confident about is "text", which routes to a
 // conservative compressor. The order is strict-JSON, diff, code, log, search.
 func (e *Engine) Detect(input []byte) string {
-	// A read tool's line-number gutter is presentation, not content: it wraps
-	// JSON, source, logs and CSV alike, and every signal below reads the first
-	// bytes of a line. Classify what the file IS, not how the agent printed it.
-	if source, _, ok := compressors.SplitLineNumberedListing(input); ok {
-		input = source
-	}
+	// Agent file tags and read-tool line-number gutters are presentation, not
+	// content. Classify what the file is, not how the agent printed it.
+	input, _ = unwrapInput(input)
 	trimmed := bytes.TrimSpace(input)
 	if len(trimmed) == 0 {
 		return TypeText
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -67,6 +67,7 @@ func defaultRegistry(counter tokens.Counter) *compressors.Registry {
 // passes the original bytes through unchanged (no handle, zero ratio).
 func (e *Engine) Compress(input []byte, opts Options) (Result, error) {
 	body, listing := unwrapListing(input)
+	body, wrapper := unwrapFileWrapper(body)
 	ct := opts.Type
 	if ct == "" {
 		ct = e.Detect(body)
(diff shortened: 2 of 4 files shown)


AmirF194 (commented):
I built this at 7ec27f9 in a golang:1.26 container and ran the test suite, plus a few adversarial cases against the quote tracking in fileOpeningTagEnd: a `>` inside a quoted attribute value, an unterminated quote, and trailing content after `</file>`. All pass, including mine.

```
go test ./engine/... -run 'FileWrap|DetectSeesThrough|Nested|CompressFileWrapped' -v
--- PASS: TestUnwrapFileWrapper (0.00s)
--- PASS: TestDetectSeesThroughFileWrapper (0.05s)
--- PASS: TestNestedFileAndListingWrappers (0.00s)
--- PASS: TestCompressFileWrappedGo (0.00s)
```

The unterminated-quote case is the one I was least sure about, and it fails closed: no wrapper detected, original bytes returned untouched, rather than misparsing the tag boundary. That is the behavior you want from a scanner over agent-supplied input.

PeterShanxin:
Looks like this is effectively on `main` now via `8fe0631` + the follow-up fixes.

Given the commits Julius pushed here before that landed, should we just mark #869 as superseded by the upstream version?

Happy to close if so - would just be good to keep the link between the two.

JuliusBrussee:
Closing as already implemented on main. Wrapper unwrapping landed in the current engine path in commit ; targeted content detection behavior is covered there. This branch now conflicts and adds no remaining fix.

JuliusBrussee:
Correction: implementation already landed on main in commit 8fe0631.

Issue #835, opened 2026-08-12
wrap (Windows): hook commands use backslash paths, which git-bash's /bin/sh hook runner strips, breaking every installed hook


Alexpakora:
Follows on from #834 — after working around that one, `caveman claude` starts but every hook (PreToolUse:Skill/Bash/Read) fails:
```
/usr/bin/bash: line 1: C:Users<user>AppDataRoamingnpmcaveman: command not found
```

Root cause: `cavemanBinForHook()` returns `which(caveman)`, a Windows path built with `path.join` (backslash-separated), embedded directly as a hook `command` string. Claude Code appears to execute hook commands via a POSIX shell (`/bin/sh -c`) even on native Windows, which treats `\` as an escape character and strips it — `C:\Users\<user>\AppData\Roaming\npm\caveman` becomes `C:Users<user>AppDataRoamingnpmcaveman`, which obviously doesn't resolve.

**Suggested fix:** normalize `\` → `/` when building any hook command string on `win32`. Confirmed this specific fix works — forward-slash absolute paths execute fine under the same hook runner (this is also how we resolved an unrelated hook path issue on the same machine, for a different tool, using the identical pattern).

**Workaround** we used locally: patched `cavemanBinForHook()` in the installed `dist/index.js` to `.replace(/\\/g, /)` on win32 before returning.

AmirF194:
**Update (see @ItamarMeir's follow-up below):** the `caveman uninstall`/`caveman install` workaround I suggested does not actually clear it. He retested the exact cycle (`caveman hooks uninstall claude` + `caveman hooks install claude` on the current npm `1.1.0`) and the regenerated entry was still the raw unquoted path. I read `installSettingsHookGeneric`/`removeSettingsHookGeneric` on HEAD and the remove-then-reinstall logic there looks like it should regenerate a quoted entry, so I can't explain the mismatch from the source alone, possibly a difference between HEAD and what shipped in `1.1.0`, or something specific to his install. Not verified further; leaving this to a maintainer or a repro from someone who can test on Windows directly. Withdrawing the "actual remedy" framing below, since it isn't confirmed.

[...]

Issue #819, opened 2026-08-10
UserPromptSubmit hook intermittently hits its 5s timeout on Windows (~4% of turns), but measures 56ms standalone


edubraqd:
## Summary

`caveman-mode-tracker.js` (UserPromptSubmit) times out for me on roughly 4% of turns. The declared timeout is 5000ms and Claude Code records the run as `hook_cancelled` with `timedOut: true`.

I benchmarked the script and it is **not** slow: 56ms typical, even against a 186MB transcript. I could not reproduce the timeout under deliberate CPU load or under concurrent writes to the transcript. I am filing this because you can instrument the hook side and I cannot, not because I think the script is the culprit.

Since `UserPromptSubmit` blocks the turn, each timeout costs the user 5 to 11 seconds before their prompt is processed.

## Observed

Scanned 150 session transcripts (2 days, 11 projects) for `hookEvent: "UserPromptSubmit"`:

- **10 timeouts / 233 invocations = 4.3%**
- Every recorded duration sample is a timeout: `hook_cancelled`, `timedOut: true`, `timeoutMs: 5000`
- Durations: 5351, 5587, 5708, 5936, 6339, 7461, 8312, 8855, 10729, 11355 ms
- Spread across 4 sessions, clustered into two time windows

Note for anyone reading transcripts the same way: successful runs with empty output are not persisted, so the *only* duration samples that exist are the failures. It is easy to mistake this for "the hook averages 7.5s" — it does not.

No correlation with transcript size:

| Session | Size | Timeouts | Total runs |
|---|---|---|---|
| 69a8fe86 | 30 MB | 4 | 46 |
| 345a1be0 | 4 MB | 3 | 44 |
| b0b24561 | 3 MB | 2 | 11 |
| 585da58e | 4.6 MB | 1 | 9 |

## Measured (standalone)

Run with `CLAUDE_CONFIG_DIR` pointed at a temp dir containing a `.caveman-active` flag, so the full active-mode path executes:

| Scenario | n | mean | max |
|---|---|---|---|
| `node -e 0` (bare startup) | 6 | 49ms | 71ms |
| `require('./caveman-config')` only | 6 | 52ms | 80ms |
| Full hook, `transcript_path` = 186MB file | 6 | **56ms** | 64ms |
| Full hook, no `transcript_path` | 6 | 58ms | 68ms |

## Hypotheses I tested and ruled out

**1. CPU contention.** 30 hook invocations while 4 background jobs saturated CPU:
[...]

PR #875, opened 2026-08-19
fix(xplat): stop hooks stalling on Windows; own only our own config entries

JuliusBrussee:
Whole-repo review of every surface that mutates a user's agent config, plus a cross-platform sweep focused on Windows.

## Windows blockers fixed

| | |
|---|---|
| `caveman-mode-tracker` | Finished its work in ~50ms, then blocked until the host killed it at the 5s budget. The `data` listener *references* the stdin handle, so `pause()` alone keeps the loop alive until the host closes the write end — and on Windows that close lags arbitrarily. This is the shape of #819. Fixed with `process.stdin.unref()`. |
| `caveman-activate` | `readFileSync(0)` blocks *inside* the read syscall until EOF; no deadline can interrupt it. The #729/#833 fix was applied to the tracker and never to its sibling. Rewritten event-driven + watchdog. |
| `hookCommand` | Emitted PowerShell call-operator syntax into `settings.json`, but Claude Code's Windows hook shell is **Git Bash**, which rejects it: `syntax error near unexpected token '&'`. Every standalone Windows install had two dead hooks (#835). |
| `install.{sh,ps1}` | Unguarded backup: a `--force` reinstall overwrote the only pre-caveman `settings.json.bak` with the merged file. |

## Config ownership

`src/hooks/uninstall.{sh,ps1}` matched hooks by the substring `caveman` and filtered whole entries — deleting user hooks that merely mention the word, and dropping foreign handlers sharing a matcher group (#593). Verified against real PowerShell: the old code left **zero** user hooks standing.

Also: OpenClaw no longer clobbers a hand-edited workspace skill; `caveman-init` writes marker-fenced and atomically instead of leaving unremovable blocks in tracked repo files; opencode `AGENTS.md` and OpenClaw `SOUL.md` refresh in place instead of fossilizing; opencode's system-prompt hook is idempotent; `readSettings` reports stripped comments so both write paths warn.

## Skills

`skills/` is auto-discovered wholesale by the plugin loader, so **13 undocumented skills** would have shipped into every user's agent on the next release. Documented, plus a `verify_repo` check that fails when a `skills/` directory is named in neither README nor CLAUDE.md.

## CI

New `windows-powershell` job covering what `engine-ci` never exercised: the PowerShell installer round trip, and a `bash -n` gate over every hook command the installer writes. That gate alone catches #835. Deliberately does not duplicate the suites `engine-ci` already runs on Windows.

**This PR is the first real execution of that job — reviewing its result is the point.**

## Verification

Every fix is pinned by a test that fails against the old code. `tests/test_hook_stdin_lifecycle.js` reproduces the Windows lagging-close condition on any platform.

Local: 172 installer tests, 10 JS suites, 7 Python suites, `go test ./engine/`, and packages/cli 612 pass / 0 fail.

The second commit (`engine: unwrap agent <file> wrappers`) was pre-existing in-flight work in the tree, committed separately at the author's request and **not** part of this review; it was build- and test-verified before committing.

https://claude.ai/code/session_01Ez7w7BBtKrMZCiZz9ea4cb


.claude-plugin/plugin.json
```diff
@@ -6,9 +6,9 @@
     "url": "https://github.com/JuliusBrussee"
   },
   "agents": [
-    "agents/cavecrew-builder.md",
-    "agents/cavecrew-investigator.md",
-    "agents/cavecrew-reviewer.md"
+    "./agents/cavecrew-builder.md",
+    "./agents/cavecrew-investigator.md",
+    "./agents/cavecrew-reviewer.md"
```
rebel-daekyeong: I love it.

JuliusBrussee:
Thanks @JuliusBrussee — this exact cross-platform/ownership fix is already on main. Closing the source PR as implemented.

Issue #690, opened 2026-07-10
SessionStart hook reads SKILL.md from wrong path — silently serves hardcoded fallback


NAJEMWEHBE:
## Bug

The `SessionStart` hook never finds `SKILL.md` and silently serves the hardcoded fallback ruleset instead of the real, level-filtered skill content.

## Cause

`plugin.json` registers the hook at `src/hooks/caveman-activate.js`:

```json
"command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\""
```

but `caveman-activate.js` resolves the skill file one level up only:

```js
skillContent = fs.readFileSync(
  path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
);
```

With `__dirname = <plugin_root>/src/hooks`, that resolves to `<plugin_root>/src/skills/caveman/SKILL.md`, which does not exist. The real file is at `<plugin_root>/skills/caveman/SKILL.md` — two levels up. The `catch` swallows the ENOENT, so every session gets the fallback block and edits to SKILL.md never propagate (contradicting the comment "Reads SKILL.md at runtime so edits to the source of truth propagate automatically").

The comment on line 51 also documents the old layout (`__dirname = <plugin_root>/hooks/`), which no longer matches where the hook lives.

## Fix that worked for me

Try the plugin layout first, keep the one-level path for standalone installs:

```js
let skillContent = '';
for (const rel of [['..', '..'], ['..']]) {
  try {
    skillContent = fs.readFileSync(
      path.join(__dirname, ...rel, 'skills', 'caveman', 'SKILL.md'), 'utf8'
    );
    break;
  } catch (e) { /* try next location - hardcoded fallback below if none found */ }
}
```

Verified on Windows 11, Claude Code Desktop: before the patch the hook emitted the 1,215-char fallback; after, it emits the 3,076-char level-filtered SKILL.md body.

Same `__dirname`-depth issue likely applies to any other path in `src/hooks/` scripts that assumes the old `hooks/` location.

AmirF194:
Tried to reproduce against current main (0d95a81) in a clean node:20-slim container, mirroring the plugin-install layout (hook at src/hooks/caveman-activate.js, no CLAUDE_PLUGIN_ROOT set):

node src/hooks/caveman-activate.js

[...]

Issue #587, opened 2026-07-01
SessionStart hook falls back to stale hardcoded ruleset — SKILL.md path off by one directory level


vadbosh:
## Bug

`src/hooks/caveman-activate.js` tries to read the live `SKILL.md` to build the SessionStart ruleset, but the computed path is off by one directory level for a `claude plugin install` layout, so it silently falls back to the hardcoded (stale) ruleset every session.

```js
// src/hooks/caveman-activate.js
skillContent = fs.readFileSync(
  path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
);
```

Plugin layout actually installed via the marketplace (`claude plugin install caveman@caveman`) is:

```
<plugin_root>/
├── src/hooks/caveman-activate.js   ← __dirname
└── skills/caveman/SKILL.md
```

`plugin.json` invokes the hook as `${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js`, so `__dirname` is `<plugin_root>/src/hooks`. `path.join(__dirname, '..', 'skills', ...)` resolves to `<plugin_root>/src/skills/caveman/SKILL.md`, which doesn't exist (no `src/skills/` directory). `readFileSync` throws, gets swallowed by the `catch`, and the hook silently uses the hardcoded fallback ruleset baked into the script — missing everything added to `SKILL.md` since that fallback text was last hand-copied (e.g. the "Preserve user's dominant language" rule, the "No self-reference" rule, the Intensity table, the wenyan examples).

The comment above the read even documents the *intended* layout as `__dirname = <plugin_root>/hooks/` (one level up from plugin root), which doesn't match the actual `src/hooks/` path used by `plugin.json`.

## Fix

```diff
- path.join(__dirname, '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
+ path.join(__dirname, '..', '..', 'skills', 'caveman', 'SKILL.md'), 'utf8'
```

Verified locally: with the fix, `node src/hooks/caveman-activate.js` (run with `CLAUDE_PLUGIN_ROOT` pointing at the real installed plugin cache dir) now emits the current `SKILL.md` content (Intensity table, etc.) instead of the fallback text.

## Repro

```bash
claude plugin install caveman@caveman
# then, simulating the hook's own resolution:
node -e "console.log(require('path').join('<plugin_root>/src/hooks', '..', 'skills', 'caveman', 'SKILL.md'))"
[...]

PR #876, opened 2026-08-19
feat(hooks): propagate caveman mode to subagents via SubagentStart

### LINKED ISSUE #672: add config to automatically start caveman mode in any new sessions, even on subagents

addibbi:
## Problem

SessionStart context is parent-thread only — it never reaches Task-spawned subagents. So in an active caveman session, every subagent (Explore, custom agents, …) runs caveman-unaware and emits full-verbosity output, forfeiting the token savings exactly where output volume is highest.

Closes #672.

## Fix

A SubagentStart hook (`src/hooks/caveman-subagent.js`) that re-reads the active level from the same flag file the other hooks use — via `readFlag`, so it inherits the symlink refusal, size cap, and `VALID_MODES` whitelist — and re-injects the level-filtered SKILL.md ruleset into each spawned subagent as `hookSpecificOutput.additionalContext` (raw stdout is silently dropped on SubagentStart). Off/absent/invalid flag → silent exit 0. No stdin read, so there is nothing to stall on spawn.

Pattern ported from ponytail (MIT, [DietrichGebert/ponytail](https://github.com/DietrichGebert/ponytail)), which closed the same gap for its own mode.

## Relation to #621

#621 proposes the same hook and I'd have been happy to just +1 it, but its SKILL.md lookup is a single `path.join(__dirname, '..', 'skills', …)` — from `src/hooks/` that resolves to `src/skills/`, which doesn't exist, so **plugin installs would always serve the weak inline fallback instead of the real ruleset**. That's the exact bug class #690 reported against caveman-activate.js (fixed in #587/#589 with the 3-candidate order). This PR:

- resolves SKILL.md with the same 3-candidate order as `caveman-activate.js` (`CLAUDE_PLUGIN_ROOT` → `../../` → `../`), so plugin installs get the real ruleset;
- guards the `caveman-config` sibling require (#848): a broken install degrades to inject-nothing, never to a MODULE_NOT_FOUND crash on every subagent spawn;
- skips independent modes (`commit`/`review`/`compress`) — their skill only exists in the parent session, so an activation line pointing at it would be an instruction the subagent can't follow;
- ships tests.

## Changes

- `src/hooks/caveman-subagent.js` — the hook (new)
- `.claude-plugin/plugin.json` — SubagentStart entry
- `bin/install.js`, `src/hooks/install.sh` + `install.ps1`, `uninstall.sh` + `uninstall.ps1`, `checksums.sha256` — wired everywhere the existing hooks are wired, idempotent and marker-guarded like them
- `tests/test_caveman_subagent.js` — 8 cases: active-mode injection with level filtering, wenyan alias, off/absent/invalid flag silence, independent-mode skip, missing-sibling degrade (#848), SKILL.md-unreachable fallback

## Verification

- `node tests/test_caveman_subagent.js` → 8/8; existing `test_hook_missing_sibling.js` (18/18) and `tests/test_hooks.py` (7/7) still green; `bash -n` / `node --check` clean on every touched script.
- e2e: `install.sh` into a temp `$HOME` wires all three hooks and copies the file; `uninstall.sh` removes both again.
- Live on Claude Code 2.1.228 (macOS): with the hook installed, a real Task-spawned subagent reports the injected ruleset in its context — `SubagentStart hook additional context: CAVEMAN MODE ACTIVE — level: full`.


AmirF194 (commented):
I read this against the other open PR touching mode storage on this repo, #904.

This hook reads mode with `readFlag(path.join(claudeDir, '.caveman-active'))`, one value for the whole machine. #904 replaces that model: mode becomes per session, stored at `.caveman-sessions/<session_id>.mode`, and keeps `.caveman-active` only as a "last-write-wins compat mirror" across open windows. #904's own CLAUDE.md addition says `resolveActiveMode(claudeDir, sessionId)` is meant to be "the reader every hook should use" going forward.

Neither PR's diff mentions the other (`caveman-subagent.js` does not exist on main at either PR's base, and #904's diff never touches it), so I don't think either author has weighed this against the other yet. If both land, a subagent spawned in a window running one mode would read whatever window wrote `.caveman-active` most recently, not its own window's mode. That reproduces the same cross-window bleed #904 is fixing for the statusline and SessionStart, just through this hook instead.

Is the plan for this hook to move to `resolveActiveMode(claudeDir, sessionId)` once #904 lands, reading `session_id` from the SubagentStart payload the way `caveman-activate.js` reads it from SessionStart? The comment at line 13 says stdin is skipped deliberately for the Windows pipe-stall reason documented for `caveman-activate.js`, so if session-scoping is wanted here too, that watchdog logic would need porting over as well, not just the read.

JuliusBrussee:
Closing after review. This implementation reads shared mode state during , which breaks current per-session isolation and cannot reliably identify parent session state from this hook payload. Current native lifecycle keeps session ownership explicit; propagation needs a host-provided parent/session linkage before it can be safe.

JuliusBrussee:
Correction: affected hook is SubagentStart; its payload lacks reliable parent-session linkage for this implementation.

TomLucidor:
Where should we look towards for OpenCode + Oh-My-Openagent @JuliusBrussee ?

PR #868, opened 2026-08-18
fix(plugin): prefix agent paths with ./ so plugin.json passes validation

LaunchAutomationAI:
## Problem

`claude plugin install caveman@caveman` fails for every user on current Claude Code. The marketplace is added successfully, then the install aborts:

```
✔ Successfully added marketplace: caveman (declared in user settings)
Installing plugin "caveman@caveman"...
✘ Failed to install plugin "caveman@caveman": Plugin temp_local_..._... has an invalid
  manifest file at ~/.claude/plugins/cache/temp_local_..._.../.claude-plugin/plugin.json.

Validation errors: agents: Invalid input
```

Reproducible against upstream `main` (8896fb5) with Claude Code 2.1.234:

```
$ git clone https://github.com/JuliusBrussee/caveman.git && cd caveman
$ claude plugin validate .
Validating marketplace manifest: .../caveman/.claude-plugin/marketplace.json

× Found 1 error:

  ❯ plugins[0] plugin.json → agents: Invalid input

× Validation failed
```

## Cause

`.claude-plugin/plugin.json` lists agent paths without the `./` prefix:

```json
"agents": [
  "agents/cavecrew-builder.md",
  ...
]
```

Claude Code's manifest schema requires each entry to be an explicit relative path. I tested every plausible shape against `claude plugin validate`:

| `agents` value | Result |
| --- | --- |
| `"./agents"` (string) | FAIL — `agents: Invalid input` |
| `"agents"` (string) | FAIL — `agents: Invalid input` |
| `["./agents"]` (directory in array) | FAIL — `agents: Invalid input` |
| `["agents/cavecrew-builder.md", ...]` (current) | FAIL — `agents: Invalid input` |
| `["./agents/cavecrew-builder.md", ...]` | **PASS** |
| field omitted | PASS (but loses the three cavecrew agents) |

Only the `./`-prefixed file-path array is accepted.

## Fix

Three lines in `.claude-plugin/plugin.json`:

```diff
   "agents": [
-    "agents/cavecrew-builder.md",
-    "agents/cavecrew-investigator.md",
-    "agents/cavecrew-reviewer.md"
+    "./agents/cavecrew-builder.md",
+    "./agents/cavecrew-investigator.md",
+    "./agents/cavecrew-reviewer.md"
   ],
```

## Verification

After the change, on the same machine and Claude Code version:

```
$ claude plugin validate .
‼ Found 1 warning:

  ❯ plugins[0] plugin.json → version: No version specified. ...

[...]

PR #877, opened 2026-08-19
fix(plugin): prefix agents paths with ./ so the manifest validates

omri-ngrid:
`claude plugin validate .` currently fails on `main`:

```
Validating marketplace manifest: .claude-plugin/marketplace.json
✘ Found 1 error:
  ❯ plugins[0] plugin.json → agents: Invalid input
✘ Validation failed
```

Claude Code's plugin manifest schema requires component paths to be relative paths beginning with `./`. The three `agents` entries use bare `agents/cavecrew-*.md`, which is rejected, and the plugin then fails to load when installed from the marketplace (it shows up in `/plugin` with a load error).

This PR adds the `./` prefix to those three entries. Nothing else changes.

After:

```
✔ Validation passed with warnings
```

(The remaining warning is the unrelated missing `version` field.)

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -6,9 +6,9 @@
     "url": "https://github.com/JuliusBrussee"
   },
   "agents": [
-    "agents/cavecrew-builder.md",
-    "agents/cavecrew-investigator.md",
-    "agents/cavecrew-reviewer.md"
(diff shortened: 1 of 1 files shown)


JuliusBrussee:
Thanks @omri-ngrid — this duplicates #868, and main solved the issue by removing the explicit list. Closing the duplicate.

PR #883, opened 2026-08-19
fix(installer): use cmd.exe-compatible quoting for Windows hook commands

tsvandrlik:
## Summary
- Claude Code spawns hook commands through `cmd.exe` on Windows, not PowerShell, but `hookCommand()` in `bin/lib/platform-paths.js` emitted PowerShell call-operator syntax (`& 'exe' 'arg'`). `cmd.exe` fails on the leading `&` with `"& was unexpected at this time"`, so `SessionStart` silently crashed every session — the caveman ruleset never loaded even though the `/caveman` mode flag got set correctly by `UserPromptSubmit`.
- `rewriteLegacyManagedHookCommands` in `bin/lib/settings.js` had the same broken quoting inline, and was actively converting working bare-`node` hook commands into the broken `&`-form on reinstall/self-heal. It now reuses the fixed `hookCommand()`.
- Verified by replaying the exact broken command through `cmd.exe` (fails) vs. the fixed cmd-quoted form (succeeds, prints the full ruleset).

## Test plan
- [x] `node --test tests/installer/platform-paths.test.mjs` — updated + new tests pass
- [x] `node --test tests/installer/unit.settings.test.mjs` — updated test passes
- [x] `node --test tests/installer/*.mjs` — full installer suite: 164 pass, 6 skipped (no `claude` CLI on PATH for e2e), 0 fail
- [x] Manually reproduced the bug and confirmed the fix against a live Windows `cmd.exe` invocation of both the hook and statusline command shapes

### DIFF
--- a/bin/lib/platform-paths.js
+++ b/bin/lib/platform-paths.js
@@ -6,9 +6,18 @@ function powershellQuote(value) {
   return `'${String(value).replace(/'/g, "''")}'`;
 }
 
+// cmd.exe-safe quoting: Claude Code spawns hook commands through cmd.exe on
+// Windows (not PowerShell), so the `& 'exe' 'arg'` PowerShell call-operator
+// form this used to emit fails every hook invocation with "& was unexpected
+// at this time" — the hook exits 1 silently and its ruleset/flag writes never
+// happen. cmd.exe needs plain double-quoted tokens, no leading `&`.
+function cmdQuote(value) {
+  return `"${String(value).replace(/"/g, '""')}"`;
+}
+
 function hookCommand(executable, args, platform = process.platform) {
   if (platform === 'win32') {
-    return `& ${[executable, ...args].map(powershellQuote).join(' ')}`;
+    return [executable, ...args].map(cmdQuote).join(' ');
   }
   return [executable, ...args]
     .map(value => `"${String(value).replace(/(["\\$`])/g, '\\$1')}"`)
--- a/bin/lib/settings.js
+++ b/bin/lib/settings.js
@@ -21,6 +21,7 @@ const fs = require('fs');
 const os = require('os');
 const path = require('path');
 const crypto = require('crypto');
+const { hookCommand } = require('./platform-paths.js');
 
 // ── stripJsonComments ──────────────────────────────────────────────────────
 // Hand-rolled state machine. Tracks string state + backslash escape so a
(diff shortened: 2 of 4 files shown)


AmirF194 (commented):
I checked out this PR's head (`c4a6987`) and ran the unmodified `hookExecutableInvocation`/`nativeHookInvocation` pair from `packages/cli/src/index.ts` in a clean `node:20-alpine` container, with the same inputs `nativeHookCommand()` feeds it (executable + fast-hook path, `platform: "win32"`).

This PR fixes `bin/lib/platform-paths.js`'s `hookCommand()`, which is what `bin/install.js` (the "full installer", `curl ... install.sh`) uses. But `@caveman-ai/cli` is a separate published package with its own, unrelated hook-command builder, and it is the primary path README.md documents (`npm install -g @caveman-ai/cli && caveman setup --install` then `caveman claude`). `caveman claude` reaches `claudeNativeMutations` -> `nativeHooksDocument("claude", ...)` -> `nativeHookCommand()` -> `nativeHookInvocation` -> `hookExecutableInvocation` (`index.ts:6127-6139`), which still prepends the PowerShell call operator on win32 and is what gets written into every lifecycle entry (`SessionStart`, `UserPromptSubmit`, etc.) of `hooks.json`:

```
$ docker run --rm -v /tmp/caveman-check/repro.mjs:/repro.mjs:ro node:20-alpine node /repro.mjs
& 'C:/Program Files/nodejs/node.exe' 'C:/Users/obrien/.../native-hook-fast.js' native-hook claude
starts with PowerShell call operator: true
```

(`repro.mjs` is `quoteHookPath`/`hookExecutableInvocation`/`nativeHookInvocation` copied verbatim
from `index.ts` at this commit, called with the same argument shape `nativeHookCommand()` uses.)

That is the exact `& '...'` shape this PR's own description names as the cmd.exe crash cause, unchanged by this diff. `index.ts:6135-6137` even documents the PowerShell assumption this PR is disproving, in a separate comment.

So on the `@caveman-ai/cli` install path, Windows sessions should still hit the silent `SessionStart` crash this PR describes. Fixing this probably means either reusing `cmdQuote`/`hookCommand` from `platform-paths.js` inside `index.ts`, or extracting one shared implementation both packages call, since right now the two maintain independent, disagreeing native/legacy quoting logic for the same platform.

JuliusBrussee:
Thanks @tsvandrlik — the proposed cmd.exe quoting targets the wrong shell and would break the verified hook path. Main already fixed the actual Bash quoting issue, so closing this patch.

PR #889, opened 2026-08-21
fix(cli): honour HTTP_PROXY/HTTPS_PROXY for outbound requests

dylanneve1:
Node's global `fetch` ignores the conventional proxy variables. On a host whose egress only leaves through an HTTP proxy, that means every request the CLI makes fails while `curl` and `npm` — which do read those variables — work fine.

First thing a new user hits:

```
$ caveman setup --install
binary download unreachable — agents still launch; traffic is NOT compressed or metered
fix: caveman setup --install
```

The fix points back at the command that just failed, and nothing in the message suggests a proxy is involved. I lost a while to it before checking whether `fetch` reads the variables at all. The same blind spot affects the npm version check, `login`, and every cloud call.

Node 24 can opt in with `NODE_USE_ENV_PROXY=1`, and 22 has the same switch behind an experimental warning, but it is off by default and a process cannot turn it on for itself after start-up — so the CLI has to implement the variables.

## What this does

New `packages/cli/src/proxy-fetch.ts`, no dependencies added:

- `resolveProxyUrl(target, env)` — `HTTPS_PROXY`/`HTTP_PROXY` by scheme with `ALL_PROXY` as fallback, lower case winning over upper (curl's precedence), bare `host:port` and credentialed forms accepted.
- `shouldBypassProxy(target, noProxy)` — `NO_PROXY` with `*`, exact host, domain suffix with or without leading dot, and optional port pinning.
- `createProxyAwareFetch(baseFetch, env)` — CONNECT tunnel for https so TLS still terminates at the target rather than the proxy, absolute-form for http. Redirects are followed with the proxy re-resolved per hop, so a hop into a `NO_PROXY` host goes direct instead of being forced back through the proxy.
- `installProxyAwareFetch()` — called at the top of `main()`, before any request. No-ops when nothing is configured, and when `NODE_USE_ENV_PROXY` is set, so a runtime already handling proxying is not proxied twice.

Wrapping `globalThis.fetch` once at start-up rather than editing ~20 call sites keeps proxy semantics in one module and leaves the call sites alone.

## Verification

A/B on a proxy-only host, `NODE_USE_ENV_PROXY` deliberately unset, clean `HOME`:

| | `setup --install` |
|---|---|
| before | `binary download unreachable` |
| after | six binaries downloaded, `checksum verified` each |

`packages/cli/tests/proxy-fetch.runtime.mjs` adds 12 tests: variable precedence and scheme selection, `ALL_PROXY` fallback, bare host and credentials, `NO_PROXY` matching including the suffix false-positive case (`example.test.evil.test` must not match `.example.test`), tunnelling through a real local proxy, bypass, redirect chains with per-hop resolution, POST bodies and methods surviving the hop, abort, and the install guard.

`npm run test:fast` and `tsc --noEmit` are clean.

## Notes for review

- I kept `Response` construction on `Readable.toWeb` so streaming downloads still stream; the release download hashes as it goes and that path is unchanged.
- Redirect cap is 20, matching the fetch spec rather than inventing a number.
- One documented cast (`asBodyInit`) because `BodyInit` in the bundled DOM types predates byte-array bodies.
- Happy to split the module into resolution and transport files if you would rather keep transport separate.

### DIFF
--- a/packages/cli/src/proxy-fetch.ts
+++ b/packages/cli/src/proxy-fetch.ts
@@ -1,0 +1,310 @@
+/**
+ * Proxy support for the CLI's outbound HTTP.
+ *
+ * Node's global `fetch` ignores the conventional proxy variables, so on a host
+ * where egress only leaves through a proxy every `fetch` in this CLI fails even
+ * though `curl` and `npm` — which do read those variables — succeed. The most
+ * visible symptom is `caveman setup --install` reporting the release download as
+ * unreachable, with no hint that a proxy is involved.
+ *
+ * Node 24 can opt in via `NODE_USE_ENV_PROXY=1` and 22 carries the same switch
+ * behind an experimental warning, but it is off by default and a process cannot
+ * turn it on for itself after start-up. This module implements the variables
+ * directly on `node:http`/`node:https`, adds no dependencies, and stays out of
+ * the way when the runtime is already handling them.
+ */
+import { request as httpRequest, type OutgoingHttpHeaders } from "node:http";
+import { request as httpsRequest, type RequestOptions } from "node:https";
+import { Readable } from "node:stream";
+import { connect as tlsConnect } from "node:tls";
+import type { Socket } from "node:net";
+
+/** Redirect hops to follow before giving up, matching the fetch spec's limit. */
+const MAX_REDIRECTS = 20;
+
+const DEFAULT_PORTS: Record<string, string> = { "http:": "80", "https:": "443" };
+
+/**
+ * Reads the first non-empty variable in `names`.
+ *
+ * Lower case wins over upper case: `http_proxy` is the older spelling and stays
+ * authoritative where both are set, which is what curl and the Python and Go
+ * toolchains do.
+ */
+function readEnv(env: NodeJS.ProcessEnv, names: readonly string[]): string | null {
+  for (const name of names) {
+    const value = env[name]?.trim();
+    if (value) return value;
+  }
+  return null;
+}
+
+/** Accepts `http://host:port`, `host:port`, and credentialed forms alike. */
+function parseProxyUrl(value: string): URL | null {
+  const candidate = /^[a-z][a-z0-9+.-]*:\/\//i.test(value) ? value : `http://${value}`;
+  try {
+    const url = new URL(candidate);
+    return url.hostname ? url : null;
+  } catch {
+    return null;
+  }
+}
+
+function portOf(url: URL): string {
+  return url.port || DEFAULT_PORTS[url.protocol] || "";
+}
+
+/**
+ * Applies one `NO_PROXY` entry to a target.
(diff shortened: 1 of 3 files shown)


packages/cli/src/proxy-fetch.ts
```diff
@@ -0,0 +1,310 @@
...
+ *
+ * The proxy is resolved per hop, so a redirect that crosses into a bypassed host
+ * is followed directly instead of being forced back through the proxy.
+ */
+export function createProxyAwareFetch(baseFetch: typeof fetch, env: NodeJS.ProcessEnv = process.env): typeof fetch {
+  return async function proxyAwareFetch(input, init) {
+    const request = new Request(input as RequestInfo, init);
+    const target = new URL(request.url);
+    const proxy = resolveProxyUrl(target, env);
+    if (!proxy) return baseFetch(input as RequestInfo, init);
```
AmirF194, line 247: I ran this against 7d4dd24 in a clean node:22-slim container.

`createProxyAwareFetch` only re-resolves the proxy per hop while it stays inside its own manual redirect loop. If the very first hop is exempted by NO_PROXY, `proxy-fetch.ts:247` returns `baseFetch(input, init)` directly without ever entering that loop, so Node's native fetch follows every subsequent redirect on its own with `redirect: "follow"`. Any later hop that lands outside NO_PROXY and should go through HTTP_PROXY is fetched directly instead. The same thing happens mid-chain: the non-proxied branch at line 265 also calls `baseFetch` with `redirect: "follow"`, so once any hop in the chain is unproxied, every hop after it is too, silently.

This matters here because a redirect through a bypassed internal host back out to a proxy-only target is a normal pattern (e.g. an internal short-link or mirror before hitting the real CDN), and it is exactly the shape `caveman setup --install` needs to survive on a proxy-only egress host.

```
$ node --test tests/proxy-fetch-repro.runtime.mjs
not ok 1 - a request that starts at a bypassed host and redirects out to a proxied target reaches the real target directly, skipping the proxy
  error: the second hop should have gone through the proxy stub
  + actual - expected
  + []
  - [ '/final' ]
```

`proxy.seen` stays empty: the stub HTTP_PROXY is never contacted for the second hop, even though its target is not in NO_PROXY.

Fix direction: drop the early return at line 247 and always enter the loop, using `redirect: "manual"` (not `"follow"`) for the direct-fetch branch at line 265-270 too, so every hop's Location header goes through the same `resolveProxyUrl` call regardless of whether the previous hop was proxied.

I have not tried this against a real HTTPS CONNECT tunnel mid-chain, only plain HTTP redirects.

JuliusBrussee:
Merged through #929 with additional proxy hardening: redirect re-evaluation, NO_PROXY handling, cross-origin credential stripping, proxy authorization boundaries, abort races, and HTTPS proxy coverage. All proxy tests passed.

PR #890, opened 2026-08-21
fix(caveman): keep procedural safety steps and citations in plain text

arc-web:
## What

Strengthens the Auto-Clarity guard so procedural safety checks and source citations always drop back to plain text instead of being compressed.

## Why

A live benchmark on Hermes (deepseek-v4-pro) found a real failure: on a multi-step nginx cert rotation, the compressed output silently dropped the openssl cert/key modulus match check (a safety verification step that must run before reload), plus the certbot-vs-manual detection step and the source citations. The existing "multi-step sequences" guard did not catch it, because the dropped step was a correctness check, not an order-ambiguity problem.

## Change

Two new drop-conditions in Auto-Clarity:

- Procedural steps that verify safety or correctness (checksum, cert/key modulus match, pre-flight check): keep every step and exact command, in order
- Source citations and reference URLs: keep them so the reader can verify

## Verification

8-case benchmark on Hermes: visible answer text cut 70 percent (24,126 to 7,307 bytes), 7 of 8 cases kept exact commands, flags, ports, and warnings. The one failure was the nginx cert rotation above, which this change addresses.

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -71,6 +71,8 @@ Drop caveman when:
 - Security warnings
 - Irreversible action confirmations
 - Multi-step sequences where fragment order or omitted conjunctions risk misread
+- Procedural steps that verify safety or correctness (checksum, cert/key modulus match, pre-flight check): keep every step and exact command, in order
+- Source citations and reference URLs: keep them so the reader can verify
 - Compression itself creates technical ambiguity (e.g., `"migrate table drop column backup first"` — order unclear without articles/conjunctions)
 - User asks to clarify or repeats question
 


skills/caveman/SKILL.md
```diff
@@ -71,6 +71,8 @@ Drop caveman when:
 - Security warnings
 - Irreversible action confirmations
 - Multi-step sequences where fragment order or omitted conjunctions risk misread
+- Procedural steps that verify safety or correctness (checksum, cert/key modulus match, pre-flight check): keep every step and exact command, in order
```
AmirF194, line 74: I read this against 985074f. The Verification section reports an 8-case benchmark (24,126 to 7,307 bytes, 7 of 8 kept exact commands) and names the nginx cert-rotation case as the one failure this change is meant to fix.

Was that benchmark run again with the updated wording, or is it the same run described in the Why section, before the fix? As written it reads as evidence for the bug, not confirmation that an LLM actually keeps the modulus check and citations once these two lines are added.

PR #891, opened 2026-08-21
fix(compress): block sensitive directory names

### LINKED ISSUE #858: caveman-compress: is_sensitive_path() misses a top-level CREDENTIALS/ directory
`caveman-compress`'s `is_sensitive_path()` guard is a **name-based denylist**. It correctly blocks
`.env`, `credentials.md`, `secrets.txt`, `id_ed25519`, `api_key.md`, and anything under
`.ssh/.aws/.gnupg/.kube/.docker`.

It does **not** block a file that holds credentials but happens to be named innocuously and to sit
in a directory whose name is not on the list.

## Reproduce

```python
from pathlib import Path
from scripts.compress import is_sensitive_path

# blocked, as intended
assert is_sensitive_path(Path("credentials.md"))
assert is_sensitive_path(Path.home() / ".ssh" / "notes.md")

# NOT blocked
assert not is_sensitive_path(Path(r"C:\dev\CREDENTIALS\hetzner\webhosting.md"))
```

The last path is a real-world layout: a top-level `CREDENTIALS/` directory with one file per
service. Neither trips the guard —

- `SENSITIVE_PATH_COMPONENTS` is `{".ssh", ".aws", ".gnupg", ".kube", ".docker"}` and does not
  contain `credentials` or `secrets`;
- the basename `webhosting.md` contains none of the `SENSITIVE_NAME_TOKENS`.

So the file is read and shipped to the API — exactly what the comment above the denylist
("a third-party data boundary that developers on sensitive codebases cannot cross") sets out to
prevent.

## Measured

I ran the guard against 11 paths while evaluating the skill: **10 correct, 1 miss** — and the miss
was the canonical credentials directory of the machine under test. It is a narrow gap, but it is
the one case where the consequence is a leak rather than a bad compression.

## Suggested fix

One line, same shape as the existing check:

```python
SENSITIVE_PATH_COMPONENTS = frozenset({
    ".ssh", ".aws", ".gnupg", ".kube", ".docker",
    "credentials", "secrets",          # added
})
```

Since the comparison already lowercases path parts (`lowered_parts`), this catches `CREDENTIALS/`,
`Credentials/` and `credentials/` alike.

## Worth considering beyond the one-liner

A name-based denylist cannot close this class of problem in general — a file named `notes.md` can
still hold an API token. If you want a stronger boundary, a **content** pre-scan before the API
call would catch what naming never will: private-key headers (`-----BEGIN … PRIVATE KEY-----`),
provider token shapes (`sk-…`, `ghp_…`, `xox[baprs]-…`), and `password|api_key|secret|token`
assignments. Cheap to run, no tokens spent, and it fails closed. Entirely your call whether that
fits the project's scope — the one-liner above already removes the concrete gap.

## Context

Found while evaluating `caveman-compress` as a candidate for a local toolchain; the guard was
otherwise the part I liked most, which is why the gap seemed worth reporting rather than working
around. Version tested: repo state of 2026-08-16 (`766dce6`).

CooperSheroy:
## Summary
- block files located under credentials/ or secrets/ directories in caveman-compress
- keep the shipped plugin copy in sync with the source script
- add a regression test for uppercase CREDENTIALS paths and non-sensitive docs paths

Fixes #858.

## Validation
- uv run --python 3.11 python -m unittest tests.test_compress_safety
- cmp -s skills/caveman-compress/scripts/compress.py plugins/caveman/skills/caveman-compress/scripts/compress.py
- git diff --check

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -67,7 +67,10 @@ SENSITIVE_BASENAME_REGEX = re.compile(
     r")$"
 )
 
-SENSITIVE_PATH_COMPONENTS = frozenset({".ssh", ".aws", ".gnupg", ".kube", ".docker"})
+SENSITIVE_PATH_COMPONENTS = frozenset({
+    ".ssh", ".aws", ".gnupg", ".kube", ".docker",
+    "credentials", "secrets",
+})
 
 SENSITIVE_NAME_TOKENS = (
     "secret", "credential", "password", "passwd",
--- a/plugins/caveman/skills/caveman-compress/scripts/compress.py
+++ b/plugins/caveman/skills/caveman-compress/scripts/compress.py
@@ -67,7 +67,10 @@ SENSITIVE_BASENAME_REGEX = re.compile(
     r")$"
 )
 
-SENSITIVE_PATH_COMPONENTS = frozenset({".ssh", ".aws", ".gnupg", ".kube", ".docker"})
+SENSITIVE_PATH_COMPONENTS = frozenset({
+    ".ssh", ".aws", ".gnupg", ".kube", ".docker",
+    "credentials", "secrets",
+})
 
 SENSITIVE_NAME_TOKENS = (
     "secret", "credential", "password", "passwd",
--- a/tests/test_compress_safety.py
+++ b/tests/test_compress_safety.py
@@ -230,6 +230,13 @@ class CompressSafetyTests(unittest.TestCase):
             backup_dir = compress_mod.backup_dir_for(path)
             self.assertFalse((backup_dir / "task.original.md").exists())
 
+    def test_sensitive_directory_names_are_blocked(self):
+        self.assertTrue(
+            compress_mod.is_sensitive_path(Path("C:/dev/CREDENTIALS/hetzner/webhosting.md"))
+        )
+        self.assertTrue(compress_mod.is_sensitive_path(Path("project/secrets/service-notes.md")))
+        self.assertFalse(compress_mod.is_sensitive_path(Path("project/docs/service-notes.md")))
+
     def test_crlf_line_endings_survive_the_round_trip(self):
         """Reading with universal newlines and writing back "\n" rewrote every
         line ending in every file the tool touched (issue #762)."""


skills/caveman-compress/scripts/compress.py
```diff
@@ -67,7 +67,10 @@ def split_frontmatter(text: str):
     r")$"
 )
 
-SENSITIVE_PATH_COMPONENTS = frozenset({".ssh", ".aws", ".gnupg", ".kube", ".docker"})
+SENSITIVE_PATH_COMPONENTS = frozenset({
+    ".ssh", ".aws", ".gnupg", ".kube", ".docker",
+    "credentials", "secrets",
```
AmirF194, line 72: I ran `is_sensitive_path` against `4b1c03f` in a clean container.

`SENSITIVE_PATH_COMPONENTS` gains `"credentials"` and `"secrets"`, but `is_sensitive_path` matches path components against this set by exact string only (`lowered_parts & SENSITIVE_PATH_COMPONENTS`, compress.py:111). It does not reuse the substring/token check that already exists for filenames (`SENSITIVE_NAME_TOKENS`, applied only to `filepath.name`). So the singular form of the same word, as a directory, still passes:

```
$ python3 -c '
from pathlib import Path
from scripts.compress import is_sensitive_path
print(is_sensitive_path(Path("project/secrets/service-notes.md")))   # True, this PR fixes it
print(is_sensitive_path(Path("project/secret/service-notes.md")))    # False, singular directory
print(is_sensitive_path(Path("project/credential/webhosting.md")))   # False, singular directory
'
True
False
False
```

That is the same failure shape #858 reported (a real credentials directory, spelled slightly off the fixed list, ships to the API unblocked), just one word-form further along the same list.

Applying the existing `SENSITIVE_NAME_TOKENS` substring check to every path part, not only `filepath.name`, would catch both plural and singular directory names without growing the frozenset further:

```python
lower_parts = [re.sub(r"[_\-\s.]", "", p.lower()) for p in filepath.parts]
if any(tok in part for part in lower_parts for tok in SENSITIVE_NAME_TOKENS):
    return True
```

JuliusBrussee:
Merged through #929. Sensitive directory blocking remains and now covers ancestor-path safety in the integrated compressor flow. Detection and safety tests passed.

Issue #88, opened 2026-04-11
github copilot on vs code support


Vivraan:
**What you want**

copilot vscode not work <- no hooks if use `npx` command

**Before/after example**
```
Before: [current behavior]
mode selector borked
After:  [desired behavior]
probably `/caveman ultra`/`/caveman wenyan-ultra` work work
```

**Why good**

me sleep under rock, xkcd 927 make too many standards, stick with basics -> easy teach younglings

balupton:
Yeah, I'd also like to know how to use it with GitHub Copilot with VSCode, as that's the only AI I'm using.

nullberri:
When using the skills the final output is in caveman but all the thinking blocks are in "normal" full sentences.

skaurus:
> When using the skills the final output is in caveman but all the thinking blocks are in "normal" full sentences.

Quote from the README: `Caveman only affects output tokens — thinking/reasoning tokens are untouched. Caveman no make brain smaller. Caveman make mouth smaller. Biggest win is readability and speed, cost savings are a bonus.`

Vivraan:
No, options no work <- my VSCode + Copilot setup is bad

Vivraan:
work VS or only Code? have VS maybe good too?

edit: vs interesting, wenyans even if no ask, cool feat

edit edit: vs start spout big words again 2 messages down line

alp-develop:
If you use VSCODE run the installation command and after installing Caverman, in the chat you have to invoke it using /caverman. There you will see the options and you ask it, for example /caverman full what is the project about? and so on. I use it and it looks good for now.

econopapi:
caveman + vscode plz

Vivraan:
> If you use VSCODE run the installation command and after installing Caverman, in the chat you have to invoke it using /caverman. There you will see the options and you ask it, for example /caverman full what is the project about? and so on. I use it and it looks good for now.

which install command? no option – fury (against machine) no bounds

also notice copilot not honouring commands, switch wenyan random, copilot always headache mooching tokens off

E.g. `/caveman ultra` (no see wenyan here tho)

```
Me: /Login has a vestigial Sign In modal, and the admin is still unable to make new events

Optimizing tool selection...

Read skill [caveman](vscode-file://vscode-app/c:/Users/vivraan/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

[...]

PR #894, opened 2026-08-22
fix(hooks): emit SessionStart context as JSON - #88

kilianpaquier:
Hello 👋

A small PR updating `caveman-activate.js` to print additional context output in JSON for both Claude Code and Copilot, since each one reads only part of it.

## Documentation proofs
- Claude: https://code.claude.com/docs/en/hooks#add-context-for-claude
- Copilot: https://docs.github.com/en/copilot/reference/hooks-reference#hook-events > `additionalContext` under `sessionStart` event

### Copilot

```sh
copilot --plugin-dir .
```

{F5D7763F-EA97-4F04-9831-94A272D577AD}

{B5894756-4334-47A4-BB96-AFB917E76200}

### Claude

```sh
claude --plugin-dir .
```

{4B90B7F4-B202-4DE4-BB3F-9175750F9395}

{BA6449BE-57C5-40C1-8E5B-E28B0D264D02}

Regards,


AmirF194 (commented):
I ran this against 607c2ffc90838e964889de1d333e3d6216a61915 in a clean container.

The "off" mode branch here still writes a bare `'OK'` instead of going through the new `emitContext()` helper, so it is the one exit path in this file that does not carry the JSON shape the rest of the PR establishes.

```
$ docker run --rm -v /repo:/repo:ro -w /repo node:20-alpine sh -c "
  export CLAUDE_CONFIG_DIR=/tmp/cfg; mkdir -p /tmp/cfg; export CAVEMAN_DEFAULT_MODE=off
  echo '{\"session_id\":\"t\",\"cwd\":\"/repo\",\"hook_event_name\":\"SessionStart\",\"source\":\"startup\"}' \
    | node src/hooks/caveman-activate.js"
OK
```

That is not valid JSON, so per this PR's own reasoning (Copilot only injects a hook's output when it parses as JSON with a top-level `additionalContext`) Copilot silently drops it, and Claude Code injects the literal string `OK` as hidden context instead of nothing. Every other exit in this file now goes through `emitContext`, so this one reads like an oversight rather than a deliberate difference.

Simplest fix: drop the `process.stdout.write('OK');` on line 261, or call `emitContext('')` for consistency with the other branches.

kilianpaquier:
Hello @AmirF194,

Good catch, to be honest in this specific case it doesn't matter much since caveman is off, like you said, Copilot will drop it and Claude will just receive an "OK". I think dropping it is the right call.

> Copilot silently drops it, and Claude Code injects the literal string `OK` as hidden context instead of nothing

AmirF194:
Agreed, thanks for confirming.

JuliusBrussee:
Closing after host-schema review. Hook output is not universally interchangeable JSON: Claude SessionStart expects its hook-specific output contract, while Codex uses a different hook surface. Applying one JSON envelope to every host makes valid plain context invalid. #929 keeps host-specific output behavior and removes unsupported Claude PostCompact wiring.

PR #898, opened 2026-08-23
chore(skills): shorten skill descriptions to fit the listing budget

smoochy:
Claude Code spends a fixed character budget on the skill listing it puts in the model's context: `contextWindowTokens * 4 * skillListingBudgetFraction`, which is about 6,000 characters at the default fraction of 0.01. Overflow is binary per skill rather than proportional truncation: skills are ranked by recent usage and granted their full description in that order until the budget runs out, and everything past the cutoff renders as a bare name with no description at all.

The 14 caveman skills spent 6,140 characters, more than the entire default budget on their own, so an install that carries caveman alongside other plugins pushes unrelated skills past the cutoff and loses them their discoverability - and, once caveman's own least-used skills fall past it, some of caveman's too.

This rewrites the descriptions to 3,539 characters total. Every trigger phrase and every boundary clause is kept, including the named slash-command and spoken triggers, the level names, and the do-not-use exclusions; what goes is restatement of the body, which the model reads anyway once the skill is selected. No skill body changes.

`skills/compile.mjs` was re-run, so `packages/cli/src/agent-skills.generated.ts` moves with its source.

Signed-off-by: smoochy <34371932+smoochy@users.noreply.github.com>


JuliusBrussee:
Merged through #929. Shorter listing-budget descriptions were regenerated into plugin artifacts and repository verification passed.

PR #900, opened 2026-08-23
fix(stats): add Claude 5 model output pricing

SomSamantray:
## Summary

`/caveman-stats` and the statusline previously showed no USD savings figure on any Claude 5 session: `priceForModel()` only matched Claude 3/4 prefixes, so `claude-*-5` model ids fell through to `null` and the `Est. saved (USD)` line, the `Pricing for <model>` footer, and the `(~$X)` parenthetical were silently omitted.

This adds the Claude 5 rate card to `MODEL_OUTPUT_PRICE_PER_M` (Fable 5 / Mythos 5 `$50`, Opus 5 `$25`, Sonnet 5 `$10` — the permanent standard price per Anthropic's pricing docs; the previously planned 2026-09-01 increase to `$15` was cancelled) and extends the prefix-match regression test, including the `[1m]` context variant Claude Code reports. A prior revision also added a `claude-haiku-5` row/assertion; that was removed after review confirmed no such model exists in Anthropic's current lineup (only Haiku 4.5 is shipped).

Fixes #744

Session-settled decisions carried from planning: price rows follow live Anthropic pricing docs, not the repo's internal catalog file, which still needs a follow-up fix for the same stale Sonnet 5 framing (user-approved); scope limited to the price-table gap and the two review-flagged corrections (user-directed).

## Validation

- `node tests/test_caveman_stats.js` — 46 passed, 0 failed
- `npm test` — 148 passed, 0 failed (4 pre-existing skips)
- `node --check src/hooks/caveman-stats.js` — clean

### DIFF
--- a/src/hooks/caveman-stats.js
+++ b/src/hooks/caveman-stats.js
@@ -40,15 +40,13 @@ function ruleOverheadPerTurn() {
 // Most-specific prefixes MUST come first — priceForModel returns the first match.
 const MODEL_OUTPUT_PRICE_PER_M = [
   // Claude 5 family. Fable/Mythos (models.anthropic.com naming) sit at the
-  // top $50/M tier; Opus 5 dropped to $25/M. Sonnet 5 is on a $10/M intro
-  // rate through 2026-08-31 ($15/M standard from 2026-09-01) per
-  // shared/provider-catalog/catalog/current.yaml; the current-period value is
-  // listed so savings are not overstated during the intro window.
+  // top $50/M tier; Opus 5 dropped to $25/M. Sonnet 5's $10/M rate is the
+  // permanent standard price — the increase to $15/M planned for
+  // 2026-09-01 was cancelled (see anthropic.com/docs/en/about-claude/pricing).
   ['claude-fable-5',   50.00],
   ['claude-mythos-5',  50.00],
   ['claude-opus-5',    25.00],
   ['claude-sonnet-5',  10.00],
-  ['claude-haiku-5',    5.00],
   // Legacy Opus 4.0 / 4.1 (pre-4.5) billed at the old $75/M output tier,
   // including the dated ids (e.g. claude-opus-4-20250514).
   ['claude-opus-4-0',    75.00],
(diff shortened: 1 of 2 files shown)


src/hooks/caveman-stats.js
```diff
@@ -39,6 +39,16 @@ function ruleOverheadPerTurn() {
 // https://www.anthropic.com/pricing if a release changes the tier.
 // Most-specific prefixes MUST come first — priceForModel returns the first match.
 const MODEL_OUTPUT_PRICE_PER_M = [
+  // Claude 5 family. Fable/Mythos (models.anthropic.com naming) sit at the
+  // top $50/M tier; Opus 5 dropped to $25/M. Sonnet 5 is on a $10/M intro
+  // rate through 2026-08-31 ($15/M standard from 2026-09-01) per
```
AmirF194: Checked the new pricing entries against the current pricing docs (platform.claude.com/docs/en/about-claude/pricing, fetched today).

Two things in this block don't hold up. Line 44's comment says Sonnet 5 is on a temporary $10/M intro rate that becomes $15/M standard on 2026-09-01. That's already stale: the docs now say the scheduled increase will not happen, and $10/M ($2/M input) is the permanent price, not a window.

Separately, `claude-haiku-5` at line 51 doesn't match any model on that page. The newest Haiku listed is 4.5; there's no Haiku 5 yet. As written, that row can only be right by coincidence if a model with that exact id and price ever ships.

Fix: drop the intro/standard framing around Sonnet 5 (it's just the price now), and hold off adding a `claude-haiku-5` row until a real model with that id and a confirmed price exists.

SomSamantray: > Fix: drop the intro/standard framing around Sonnet 5 (it's just the price now), and hold off adding a `claude-haiku-5` row until a real model with that id and a confirmed price exists.

Fixed in aaaf97d and c242c75 — thanks for the catch. Re-verified against the live pricing page just now: it explicitly states the $2/$10 rate for Sonnet 5 is now the standard price and the planned 2026-09-01 increase "will not occur," so the comment now states that directly instead of framing it as an intro window. The `claude-haiku-5` row and its test assertion are removed entirely, and I added `assert.strictEqual(priceForModel('claude-haiku-5'), null)` to lock in that it stays unpriced rather than silently reappearing.

AmirF194: Confirmed, that matches the live pricing page. Thanks for the quick fix.

JuliusBrussee:
Merged through #929. Claude 5 pricing support remains; learn accounting now also reports unmeasured plugin token tax as unavailable rather than zero.

Issue #599, opened 2026-07-02
/caveman-commit permanently clobbers the active caveman level; namespaced /caveman:caveman-commit and -review are unrecognized


AmirF194:
**What happen**

Running `/caveman-commit` (or `-review` / `-compress`) overwrites the flag file with `commit` and **nothing ever restores the prior mode**. Since these are `INDEPENDENT_MODES`, the per-turn reinforcement stops emitting for every subsequent turn, and the statusline is stuck at `[CAVEMAN:COMMIT]`.

`src/hooks/caveman-mode-tracker.js:76-81, 97-98, 123`.

**Repro (confirmed)**

1. flag = `ultra`
2. prompt `/caveman-commit` → flag = `commit`
3. next prompt "ordinary question" → reinforcement stdout is empty; caveman prose mode is effectively dead until the user manually re-runs `/caveman`

SKILL.md promises "Level persist until changed or session end" — a one-shot skill invocation shouldn't count as "changed" for the rest of the session.

**Also**

Only compress and stats got plugin-namespaced variants: `/caveman:caveman-compress` works (`:80`), but `/caveman:caveman-commit` and `/caveman:caveman-review` are not recognized at all (confirmed: flag unchanged) — so in a marketplace plugin install those two skills run **with** the caveman-prose reinforcement injected into their turn, exactly what the INDEPENDENT_MODES suppression exists to prevent. Probably the same root cause as #537.

**Fix direction**

Store the previous prose mode alongside the one-shot mode (e.g. `commit:ultra`) and restore it on the next prompt, or suppress reinforcement for one turn without persisting `commit` to the flag. Add the missing `(?::caveman)?` namespaced variants for commit/review.

main @ 25d22f8.

PR #904, opened 2026-08-24
fix: scope caveman mode to the session

pavel-collab:
## The problem

The mode is applied per session but stored in one file per machine
(`$CLAUDE_CONFIG_DIR/.caveman-active`). Four user-visible bugs follow from
that single fact:

1. **Parallel windows share a mode.** `/caveman ultra` in one window
   retargets the other.
2. **"stop caveman" is undone by auto-compaction.** Deactivation is spelled
   "no flag file", so the next SessionStart re-derives the configured
   default and switches caveman back on.
3. **The statusline badge is machine-wide.** Every window renders whichever
   mode was written last.
4. **`off` cannot survive a SessionStart** — there is nowhere to put it.

Reproducing #2 on current `main`, driving the hooks exactly as Claude Code
does:

```bash
export CLAUDE_CONFIG_DIR=$(mktemp -d)
echo '{"session_id":"A","source":"startup"}' | node src/hooks/caveman-activate.js >/dev/null
echo '{"session_id":"A","prompt":"stop caveman"}' | node src/hooks/caveman-mode-tracker.js
echo '{"session_id":"A","source":"compact"}'  | node src/hooks/caveman-activate.js
# → CAVEMAN MODE ACTIVE — level: full

Why #691 does not close it

#691 branches SessionStart on the payload's source, which correctly
preserves a mid-session level change across a compaction. But it reads
the legacy flag, and in that file "off" is expressed as absence — so a
deactivated session finds nothing stored and falls through to
getDefaultMode() anyway. Bugs 1, 3 and 4 are consequences of the storage
shape rather than of when the hook re-derives: one file cannot hold two
windows' modes, and an absent file cannot express "deliberately off"
distinguishably from "never set".

The change

State moves to $CLAUDE_CONFIG_DIR/.caveman-sessions/<session_id>.mode,
keyed by the session_id Claude Code already puts in every hook payload
and in the statusline's stdin JSON. .caveman-active stays as a
last-write-wins compat mirror.

Two invariants are load-bearing:

- The mirror never holds the literal off. Deactivation unlinks it.
  off is in VALID_MODES, so an older caveman-mode-tracker.js reading
  it from that path would pass its !INDEPENDENT_MODES.has(...) check and
  inject "CAVEMAN MODE ACTIVE (off)", and an older caveman-statusline.sh
  would render [CAVEMAN:OFF]. Mixed-version installs are real: plugin
  hooks and standalone hooks can both be registered, and statusLine holds
  a path baked in at install time.
- recordModeChange compares against the session's own state, never the
  mirror. Diffing against the mirror would compare one session's
  transition to whatever another session wrote last and spam the log with
  phantom entries. That is why readSessionModeRaw exists alongside
  resolveActiveMode.

Both spellings of off are read everywhere: a missing file (old) and a
literal off (new, durable).

caveman-config.js gains the primitives — validateSessionId (a
whitelist, because a session id becomes part of a path),
resolveActiveMode, readSessionModeRaw, writeSessionMode, per-session
prev for one-shot skills (#599), gcSessionStore (14-day mtime sweep,
startup only — a compaction should not spend the 5s budget walking a
directory), and a session_id-tagged recordModeChange (#601).

caveman-stats filters the transition log on session_id, so a switch in
window B no longer lands inside window A's timeline and skews its estimate.
Both statusline ports take session_id from stdin with bounded,
TTY-guarded reads and render nothing for a deactivated session. Uninstall
(bin/install.js, uninstall.sh, uninstall.ps1) removes the session
store; .caveman-history.jsonl is kept on purpose.

Backward compatibility

Every helper accepts a sessionId that may be null or malformed and
degrades to the previous machine-wide behavior. The old code path is the
fallback branch — which is why the pre-existing suites, none of which send
a session_id, pass unchanged. A session that began before this change,
with only the legacy flag on disk, keeps its mode including across a
compaction.

The three hook entrypoints resolve the new helpers individually
(cfg.writeSessionMode || <legacy stub>) instead of adding them to their
requireSibling shape checks. A caveman-config.js from before this
change satisfies those checks, so hard-failing over the newer exports would
trade "machine-wide mode, as it always worked" for "no state at all" on
exactly the plugin-cache drift #848 exists for.

Decisions worth challenging

- clear re-derives the default; compact/resume/fork do not. This
  differs from #691's comment, which groups clear with the continuations.
  Rationale: /clear is an explicit reset and nothing else in the
  conversation survives it, so neither should a "stop caveman" from before
  it. Per-session storage makes the distinction cheap. Easy to flip if you
  disagree — it is one entry in RESET_SOURCES.
- The payload watchdog's unknown source preserves stored state. If it
  assumed startup, a slow payload on a compaction would drop a
  mid-session ultra (#691 through the timeout door) and re-arm a session
  the user turned off.
- The mirror is kept at all. It costs one extra write per mode change.
  It is there because INSTALL.md tells users to cat it and
  src/hooks/README.md ships a third-party statusline snippet that reads
  it. Dropping it would be a cleaner design and a silent break for those
  readers.

Not touched

- opencode. src/plugins/opencode/plugin.js keeps the machine-wide
  behaviour: it writes its own flag in the opencode config dir and never
  sees a Claude Code session_id. Out of scope here.
- PowerShell badge behaviour. The Windows CI job parses every .ps1 and
  drives install/uninstall, but nothing renders the badge, so
  caveman-statusline.ps1 is covered only by the parity greps added to
  verify_repo.py (SESSIONS_DIRNAME, SESSION_ID_RE, the durable-off
  short-circuit, the stdin guards, and the bash-3.2 fractional-timeout ban).

Tests

New:

- tests/hooks/caveman-config.test.mjs — 25 in-process cases over the
  storage layer: id validation and path containment, durable off, prev
  scoping, mode-log tagging, GC including symlink refusal.
- tests/test_hooks.py::SessionStartSourceTests — source branching, plus
  the stdin hang guard (fails by timing out, not by asserting).
- tests/test_mode_tracker.py::SessionScopedModeTests — two windows,
  independent modes, mirror never holding off, traversal ids.
- 5 statusline cases in tests/test_caveman_stats.js — session badge,
  durable off, legacy fallback, traversal, multiline JSON.
- tests/manual/session-mode-smoke.sh — 22 end-to-end checks driving the
  real hook binaries against a throwaway CLAUDE_CONFIG_DIR.

package.json and the Node 18 fallback in ci.yml pick up
tests/hooks/*.test.mjs. src/hooks/checksums.sha256 regenerated.

npm test                              197/197
node --test tests/*.js                10 files, 0 failures
python -m unittest discover -s tests  102/102
python tests/verify_repo.py           pass
bash tests/manual/session-mode-smoke.sh   22/22

docs/testing-session-modes.md carries the full three-layer plan,
including the checks that need a live Claude Code (two windows with
different badges, /compact, --continue) and the mixed-version
upgrade/downgrade runs.


JuliusBrussee:
Merged through #929 with further session isolation and native-runtime concurrency hardening. Session-mode, hook, race, and Windows suites passed.

PR #906, opened 2026-08-24
fix(plugin): prefix agents paths with ./ so plugin install works on current Claude Code

askmarketingmanagement:
## Problem

`claude plugin install caveman@caveman` fails on current Claude Code (2.1.x):

```
✘ Failed to install plugin "caveman@caveman": Plugin temp_local_... has an
  invalid manifest file at .../.claude-plugin/plugin.json.

Validation errors: agents: Invalid input
```

This blocks the plugin install path for every user on current CC. The standalone
hook install (`--with-hooks`) still succeeds, which is why the failure is easy to
miss — hooks work, but the skills and cavecrew agents never load.

## Cause

The CC manifest validator requires plugin-relative paths to start with `./`.
`plugin.json` lists the cavecrew agents as bare relative paths.

The same rule already governs the `skills` field — the repo's own `CLAUDE.md`
documents it under the ClawHub publishing constraints ("all paths must be
relative to the plugin root and start with `./`"; CC returns
`Validation errors: skills: Invalid input` otherwise). `agents` is subject to the
same validator and was missed.

## Fix

Prefix the three `agents` entries with `./`.

## Verification

Tested against a real marketplace copy under
`~/.claude/plugins/marketplaces/caveman/`, reinstalling between each variant:

| `agents` value | Result |
| --- | --- |
| `["agents/cavecrew-builder.md", ...]` (current) | ✘ `agents: Invalid input` |
| `"./agents"` (string dir) | ✘ `agents: Invalid input` |
| `["./agents/cavecrew-builder.md", ...]` (this PR) | ✔ installs cleanly |

After the fix, `claude plugin install caveman@caveman` succeeds and the plugin
loads its skills plus all three cavecrew agents (builder, investigator, reviewer).

Environment: Windows 11, Claude Code 2.1.x, Node v24.19.0.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -6,9 +6,9 @@
     "url": "https://github.com/JuliusBrussee"
   },
   "agents": [
-    "agents/cavecrew-builder.md",
-    "agents/cavecrew-investigator.md",
-    "agents/cavecrew-reviewer.md"
+    "./agents/cavecrew-builder.md",
+    "./agents/cavecrew-investigator.md",
+    "./agents/cavecrew-reviewer.md"
   ],
   "hooks": {
     "SessionStart": [


AmirF194 (commented):
I compared this PR's base blob for `.claude-plugin/plugin.json` against the file's commit history on `main`, since `mergeStateStatus` shows DIRTY.

The PR's diff starts from blob `af47931b3`, which is the version of the file after commit `902eba34` (2026-08-18), where `agents` was first added with bare paths. But commit `0385ad9b` (2026-08-19, "fix(plugin): make the Claude plugin load the agents and skills it ships") removed the `agents` field from the manifest entirely, five days before this PR's branch was cut. Current `main` (blob `85bf01619`) has no `agents` key at all, only `name`, `description`, `author`, and `hooks`.

```
$ gh api repos/JuliusBrussee/caveman/contents/.claude-plugin/plugin.json --jq '.sha'
85bf01619adf4ebdbf3ffd75276a523323c9f9c5
$ gh api "repos/JuliusBrussee/caveman/contents/.claude-plugin/plugin.json?ref=0385ad9b" --jq '.sha'
84cdabacd98bdd89db03281a1e61f4df9afb81eb
$ gh api repos/JuliusBrussee/caveman/commits/0385ad9b --jq '.files[] | select(.filename==".claude-plugin/plugin.json") | .patch'
@@ -5,11 +5,6 @@
     "name": "Julius Brussee",
     "url": "https://github.com/JuliusBrussee"
   },
-  "agents": [
-    "./agents/cavecrew-builder.md",
-    ...
-  ],
   "hooks": {
```

So resolving the conflict by keeping this PR's `agents` block, prefixed or not, would re-add a field that was deliberately removed. Worth confirming against `0385ad9b` whether the agents are meant to be manifest-declared at all now, before rebasing this onto current `main`.

JuliusBrussee:
Closing after manifest validation. Prefixing paths is not the remaining problem: this change restores the old array-shaped  declaration, which current Claude plugin discovery does not load. Current main uses the supported agent manifest shape and recursive verification guards it.

JuliusBrussee:
Correction: branch restores old array-shaped agents declaration, which current Claude plugin discovery does not load.

PR #917, opened 2026-08-27
feat: add dnd mode — silent while working, report once done

zsecducna:
Adds `dnd` ("do not disturb") as a seventh intensity level: the agent emits no prose while working — no preamble, no plan, no progress notes, no narration between tool calls. Thinking is unrestricted. It speaks only to deliver a final report when the job is done, to ask a question when the request is ambiguous, or to request confirmation when an action needs approval. The existing auto-clarity rule still overrides, so security warnings and irreversible-action confirmations are always spoken.

Implemented as a flat level rather than a modifier on lite/full/ultra: the flag file, statusline whitelist, stats attribution, and mode parser all key off a single mode string, so no format change is needed. The behavior lives in one intensity-table row in skills/caveman/SKILL.md, which the activate hook already filters to the active level — other levels pay no extra tokens.

- src/hooks/caveman-config.js: add 'dnd' to VALID_MODES (kept out of INDEPENDENT_MODES so it persists and gets per-turn reinforcement)
- skills/caveman/SKILL.md: dnd intensity row plus two examples
- src/hooks/caveman-activate.js: dnd rule in the fallback ruleset used when SKILL.md cannot be read
- statusline .sh/.ps1 whitelists: render [CAVEMAN:DND]
- docs, slash-command hints, synced skill copies, caveman.skill zip
- tests/test_caveman_parse.js: unit and hook-parity coverage for /caveman dnd
- src/hooks/checksums.sha256: regenerated for the changed hooks


AmirF194 (commented):
Ran caveman-init.js against 84d8934 in a clean node:20-slim container.

src/tools/caveman-init.js's RULE_BODY (and its preferred source, src/rules/caveman-activate.md, per the "keep these in sync" comment at caveman-init.js:17) still lists six levels: no dnd, and no "Do Not Disturb" text anywhere. src/hooks/caveman-activate.js gets the full new paragraph; this file does not.

That RULE_BODY is what gets written to .cursor/rules/caveman.mdc, .windsurf/rules/caveman.md, .clinerules/caveman.md, .github/copilot-instructions.md, and AGENTS.md. So every one of those five integrations ships with zero mention of dnd or what it should do.

```
$ node caveman-init.js /tmp/target --only cursor
$ grep -ic dnd /tmp/target/.cursor/rules/caveman.mdc
0
```

Fix: add the dnd line and the Do Not Disturb paragraph to src/rules/caveman-activate.md (and the RULE_BODY fallback in caveman-init.js), matching what caveman-activate.js already says.

zsecducna:
Confirmed and fixed in 2742413.

What changed:

- `src/rules/caveman-activate.md` — `dnd` added to the switch line, plus a Do Not Disturb
  paragraph worded to match what `caveman-activate.js` already emits.
- `src/tools/caveman-init.js` — `RULE_BODY` re-mirrored from that file verbatim.

Chasing the same root cause turned up two more sites the level had missed:

- `src/rules/caveman-openclaw-bootstrap.md` — switch line (the standalone copy in
  `bin/lib/openclaw.js` already had it, so those two had drifted from each other too).
- `src/hooks/caveman-activate.js` — the degraded fallback ruleset's own
  "Current level ... Switch:" line still advertised six levels.
- `packages/cli/src/agent-skills.generated.ts` — regenerated with `node skills/compile.mjs`;
  the CLI embed was carrying the pre-dnd SKILL.md.

Since a comment ("keep these in sync") was the only thing holding `RULE_BODY` to the rules
file, I added a drift guard in `tests/test_caveman_init.js` that asserts the two are equal
byte for byte — same shape as the `FALLBACK_VALID_MODES` check in
`test_hook_missing_sibling.js`, which is what caught the other copy of this bug earlier in
the branch.

Local runs are green: `tests/test_*.js`, `python -m unittest discover -s tests` (78),
`tests/verify_repo.py`, `npm test`, and the `packages/cli` verbs-gate/registry suites.

AmirF194:
Thanks for the fast fix and for chasing the other drifted copies down, that is a more complete fix than the one I flagged.

JuliusBrussee:
Closing as not planned for this package. Suppressing all progress output changes host interaction and notification semantics, not Caveman response style, and can hide approval/input waits. Hosts should own do-not-disturb behavior; Caveman must preserve visible status for long-running work.

PR #921, opened 2026-08-27
fix(compress): isolate claude --print subprocess from caller's hooks/MCP

### LINKED ISSUE #920: `caveman-compress`: `claude --print` subprocess inherits caller's hooks/MCP, leaks commentary into compressed output
### Description

Looks like `call_claude()` in `skills/caveman-compress/scripts/compress.py` falls back to `claude --print` when `ANTHROPIC_API_KEY` isn't set.

That subprocess boots a full Claude Code session, inheriting the caller's `~/.claude/settings.json`, including `UserPromptSubmit``SessionStart` hooks and the full MCP server list.

I have seen 2 types of failure, both writing meta-commentary into the file being "compressed":

1. A `UserPromptSubmit` hook injects unrelated context into the compression prompt; the subprocess narrates about noticing and ignoring it, and that narration ends up in the output file.
2. The subprocess reaches for an inherited MCP tool, hits a non-interactive permission wall, and narrates the failure instead of just compressing text.

### Steps to Reproduce

- Precondition: a machine with hooks configured

- Feed `build_compress_prompt(text)` output to `claude --print` via stdin a few times...
  - Intermittent, tied to per-prompt hook heuristics, but real.
  - Hit 3 of 6 files in one session for me.

### Proposed Fix

- Pass `--setting-sources ""` and `--strict-mcp-config` to the subprocess call
- That turns it into a plain text completion, without touching auth.

Note: I guess `--bare` would _also_ work to fix this issue, but it would require API-key-based auth, not usable on OAuth-only desktop installs.

PR: https://github.com/JuliusBrussee/caveman/pull/921 - I am using that in my fork, and I personally verified a clean output on previously-leaky files after the change.

rsenna:
Fixes #920

## Summary
- \`call_claude()\`'s \`claude --print\` fallback boots a full Claude Code session, inheriting the caller's \`~/.claude/settings.json\` hooks and MCP server list.
- Two observed leaks: a \`UserPromptSubmit\` hook injecting unrelated context that the subprocess narrates about instead of silently ignoring, and the subprocess reaching for an inherited MCP tool, hitting a non-interactive permission wall, and narrating that instead of compressing.
- \`--setting-sources ""\` + \`--strict-mcp-config\` isolate the call to a plain text completion, no auth change needed. \`--bare\` was considered but requires \`ANTHROPIC_API_KEY\`/\`apiKeyHelper\`, which isn't available on OAuth-only desktop installs — exactly the case this fallback exists for.

## Test plan
- [x] Reproduced the leak by piping \`build_compress_prompt(text)\` output straight to \`claude --print\` on a machine with hooks configured
- [x] Confirmed clean output with \`--setting-sources "" --strict-mcp-config\` added, same machine, same previously-leaky inputs
- [x] Ran the patched \`compress.py\` end-to-end on 3 files that leaked before the fix — all compressed cleanly with real byte reductions, no injected commentary

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -298,10 +298,19 @@ def call_claude(prompt: str) -> str:
     # shutil.which returns the same absolute path as the implicit lookup,
     # so this is a no-op there. Falls back to bare "claude" if not found
     # on PATH so subprocess raises a clear FileNotFoundError.
+    # --setting-sources "" and --strict-mcp-config keep this a plain text
+    # completion. Without them, `claude --print` boots a full session that
+    # inherits the caller's ~/.claude/settings.json hooks and MCP servers —
+    # a UserPromptSubmit hook injecting unrelated context, or the subprocess
+    # reaching for an inherited MCP tool and hitting a permission wall it
+    # can't resolve non-interactively, both leak narrated commentary into
+    # what is supposed to be pure compressed markdown. --bare would isolate
+    # more but requires ANTHROPIC_API_KEY/apiKeyHelper auth, unusable for
+    # OAuth-only desktop installs, which is exactly what this fallback is for.
     claude_bin = shutil.which("claude") or "claude"
     try:
         result = subprocess.run(
-            [claude_bin, "--print"],
+            [claude_bin, "--print", "--setting-sources", "", "--strict-mcp-config"],
             input=prompt,
             text=True,
             capture_output=True,


skills/caveman-compress/scripts/compress.py
```diff
@@ -298,10 +298,19 @@ def call_claude(prompt: str) -> str:
...
+    # reaching for an inherited MCP tool and hitting a permission wall it
+    # can't resolve non-interactively, both leak narrated commentary into
+    # what is supposed to be pure compressed markdown. --bare would isolate
+    # more but requires ANTHROPIC_API_KEY/apiKeyHelper auth, unusable for
+    # OAuth-only desktop installs, which is exactly what this fallback is for.
     claude_bin = shutil.which("claude") or "claude"
     try:
         result = subprocess.run(
-            [claude_bin, "--print"],
+            [claude_bin, "--print", "--setting-sources", "", "--strict-mcp-config"],
```
AmirF194, line 313: This fix is correct: I installed the claude CLI in a clean container and confirmed both flags exist and do what the comment says (`--strict-mcp-config` restricts MCP servers to those from `--mcp-config`, which is empty here; `--setting-sources ""` loads none of user/project/local settings, so hooks defined in any of those files cannot fire).

The gap is test coverage. Every existing test in `tests/test_compress_safety.py` mocks `call_claude` itself (e.g. lines 39, 50, 60, 85), so nothing exercises the actual `subprocess.run` argument list. A future edit that drops these two flags while refactoring this function would pass the whole suite silently.

I added this test locally and ran it both ways: it passes at this PR's head (077efe2) and fails on main (17f9f2e) with `AssertionError: '--setting-sources' not found in ['/usr/bin/claude', '--print']`, so it does pin the behavior rather than just restate it.

```python
def test_cli_fallback_isolates_settings_and_mcp(self):
    with mock.patch.object(compress_mod, "shutil") as shutil_mod, \
         mock.patch.object(compress_mod.subprocess, "run") as run:
        shutil_mod.which.return_value = "/usr/bin/claude"
        run.return_value = mock.Mock(stdout="compressed", returncode=0)
        compress_mod.call_claude("prompt")
        args = run.call_args[0][0]
        self.assertIn("--setting-sources", args)
        self.assertIn("--strict-mcp-config", args)
```

PR #928, opened 2026-08-29
fix(installer): avoid CLI bin collision

### LINKED ISSUE #881: npx github:JuliusBrussee/caveman -- <flag> invokes the wrong caveman bin (collides with @caveman-ai/cli dependency)
\`npx -y github:JuliusBrussee/caveman -- --uninstall\` (and any other flag) fails:
\`\`\`
unknown command "--" — did you mean \`caveman run\`?
see: caveman --help
\`\`\`
That error/help text isn't from \`bin/install.js\` (the installer) — it's from the dependency \`@caveman-ai/cli\`.

**Root cause:** \`package.json\` declares \`"bin": {"caveman": "./bin/install.js"}\`, but \`@caveman-ai/cli\` (listed in \`dependencies\`, v1.2.1) *also* declares \`"bin": {"caveman": "dist/index.js", "cave": "dist/index.js"}\`. When npx installs the temp project, \`node_modules/.bin/caveman\` ends up pointing at \`@caveman-ai/cli\`'s \`dist/index.js\`, not the installer — confirmed by inspecting the npx cache directory directly (both \`.bin/caveman\` and \`.bin/cave\` resolve to \`@caveman-ai/cli\`).

So every documented top-level command in the README/\`--help\` (\`npx -y github:JuliusBrussee/caveman\`, \`-- --all\`, \`-- --uninstall\`, etc.) silently runs the wrong CLI. This isn't Windows-specific — it's a bin-name collision in the dependency tree and should reproduce on macOS/Linux too.

**Repro:**
\`\`\`
npx -y github:JuliusBrussee/caveman -- --uninstall
\`\`\`
**Expected:** runs \`bin/install.js --uninstall\`.
**Actual:** runs \`@caveman-ai/cli\`'s CLI, which rejects bare \`--\`.

**Environment:** Windows, npm 11.17.0, node v24.19.0.

**Workaround:** clone the repo and run \`node bin/install.js --uninstall\` directly (bypasses the npx bin resolution entirely).

**Fix suggestion:** rename the installer's own bin (e.g. \`caveman-install\`) or drop \`@caveman-ai/cli\` from \`dependencies\` (make it an optional/separate install) so the two \`caveman\` bins stop colliding.

pttydou:
Fixes #881.

The installer package and @caveman-ai/cli both exposed a caveman bin, so npm linked the dependency's CLI over the installer in its temporary environment. This renames the installer's sole bin to caveman-install and adds a regression that prevents future dependency-bin overlap.

Tests:
- `node --test tests/installer/package-bin.test.mjs`
- `pnpm run test:windows`
- `python tests/verify_repo.py`

I also verified the packed package with npm 11.17.0 and pnpm 10.14.0. The full installer suite passed 165 tests and skipped 10; its only local failure was the unchanged Git Bash parser test because Bash is not installed on this machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "url": "https://github.com/JuliusBrussee/caveman/issues"
   },
   "bin": {
-    "caveman": "./bin/install.js"
+    "caveman-install": "./bin/install.js"
   },
   "engines": {
     "node": ">=18"
--- a/tests/installer/package-bin.test.mjs
+++ b/tests/installer/package-bin.test.mjs
@@ -1,0 +1,39 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..');
+
+function readPackage(...parts) {
+  return JSON.parse(fs.readFileSync(path.join(ROOT, ...parts, 'package.json'), 'utf8'));
+}
+
+function packageBins(pkg) {
+  if (typeof pkg.bin === 'string') {
+    return { [pkg.name.replace(/^@[^/]+\//, '')]: pkg.bin };
+  }
+  return pkg.bin ?? {};
+}
+
+test('installer bin does not collide with the bundled CLI', () => {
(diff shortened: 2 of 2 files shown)


JuliusBrussee:
Merged through #929. Installer bin collision is fixed and pinned by package-entry regression coverage plus pack inspection.

PR #931, opened 2026-08-30
fix(stats): correct multi-block accounting, bound lifetime aggregation, pin number formatting

Alot1z:
## Summary

Three correctness/robustness fixes for `/caveman-stats`, plus upgrades that close the lifetime-accounting and integrity gaps they revealed.

### Fix 1 — multi-block responses counted once (#793)
`parseSession` summed `usage.output_tokens` and incremented `turns` **per JSONL line**. A single response split across multiple lines (multi-block tool output) was double/triple-counted, inflating both turns and output tokens ~2×.

The dedupe now keys on the response (`requestId` + `message.id`, matching upstream PR #794's approach), so one response counts once regardless of how many blocks the SDK split it across.

### Fix 2 — lifetime aggregation is bounded
`.caveman-history.jsonl` is append-only and grows without bound — a row is written on *every* stats run. The old code re-read and re-parsed the **whole** file on every `/caveman-stats` call, under the hook's fixed 2.5s child watchdog. For heavy users this silently kills stats *and* the statusline suffix forever once history exceeds the per-run budget; nothing trimmed the file to recover.

This is replaced with **watermark-backed incremental aggregation**: a `.summary.json` sidecar keeps the latest-per-session totals; each run parses only the bytes appended since the last watermark and merges them in (`MAX_INCREMENTAL_BYTES` guards a pathological gap with a safe one-off rebuild). Cost is bounded by what was written since the last call, not by total history length.

### Fix 3 — number formatting is locale-stable
The 6 tests that hardcoded en-US separators (`1,850` vs `1.850`) failed on any non-en-US machine. All `toLocaleString()` call sites were replaced with a pinned en-US formatter, so the suite is green everywhere.

## Upgrades
- **`--reset`** — rotate `.caveman-history.jsonl` to a dated `.bak`, drop the summary sidecar + statusline suffix, so lifetime numbers can start clean. History is archived, never deleted.
- **Verify-on-write sidecar** — `saveHistorySummary` writes to a temp, reads it back, and only renames into place when the round-trip is byte-lossless. A corrupt sidecar is never installed silently; every run after a revert is a safe full rebuild.
- **Concurrent-writer guard** — two `/caveman-stats` runs can race the same history file; a plain read-modify-write could drop one run's increments. `persistSummaryConcurrent` folds in a peer summary that already advanced (bounded retries, only on the monotonic incremental path). A rotation rebuild never resurrects rotated-away sessions.
- **`--reset` reachable from the slash command** — the `UserPromptSubmit` hook now forwards `--reset` alongside `--all`/`--since`.
- **Docs** — the lifetime view (`--all`/`--since`/`--reset`) is now documented in `skills/caveman-stats/README.md` + `SKILL.md`.

## Integrity
`src/hooks/checksums.sha256` was regenerated (`caveman-stats.js` and `caveman-mode-tracker.js` both changed); `tests/verify_repo.py::verify_powershell_static` passes.

## Tests
- Multi-block + cross-request dedupe regression fixtures.
- Across-runs incremental aggregation + rotation rebuild.
- `--reset` archives history, clears sidecar/suffix, clean slate after.
- Concurrent-CAS: two sessions survive a peer-ahead persist (no lost update).
- All **56** stats tests pass, including the 6 previously locale-red.

No behavioral change to the session view; `--since`, `--all`, `--share` paths unchanged.

## Known limitations
- The speedup is a sidecar optimization: the numbers are only incremental as long as the `.summary.json` sidecar is intact. If it is ever corrupt or deleted, the next run does a one-off **full rebuild** of history (bounded, correct, just not incremental), then resumes incremental mode — never a wrong number, never a hard failure.
- `MAX_INCREMENTAL_BYTES` caps how much new data a single run will parse; an unusually large gap between runs triggers that same safe full rebuild rather than an unpredictable long parse.
- The dedupe keys on `(requestId, message.id)`, so it is per-message, not per-request: two genuinely different responses in one request are each counted (correct), while one response split across many content-block lines is counted once (the #793 fix).
- `--reset` archives the raw history to a dated `.bak` and clears the lifetime view; it does not re-measure anything and does not touch per-session transcript files.

---

## Merge order

## Related

**Related PRs** — this is #1 of a five-PR batch: #932, #933, #934, #936.

**Referenced** — #793 (multi-block over-count bug this fixes), #794 (the upstream PR whose `(requestId, message.id)` dedupe this mirrors; #933 reuses the same key).

## Test status
- Command: `node tests/test_caveman_stats.js && python tests/verify_repo.py`
- Result: `56 passed, 0 failed`


Alot1z:

Alot1z:
Hello! First of a five-PR upstream batch, opening here so the ship plan is visible from the earliest PR.

**Suggested merge order:** #931 → #932 → #933 → #934 → #936

- **#931** fix(stats): multi-block accounting, bounded lifetime aggregation, locale-stable formatting
- **#932** fix(compress): provider-agnostic reply parsing — no crash after a paid call
- **#933** feat(l4): evidence-gated compression measurement runtime + `/caveman-measure`
- **#934** feat(mcp): Remote-MCP server (ChatGPT / Claude / Gemini / pi)
- **#936** feat(extension): Firefox port of the extension

Why this order: **#931–#933 are the foundations** (the `(requestId, message.id)` dedupe in #931 is reused by #933; #932 hardens the compress reply parsing), and **#934 + #936 build on them** (the MCP engine and the Firefox port). Each PR is independently mergeable — the order is advisory, so reviewers see the dependency chain as they land.

Current heads are green: each PR's Test section lists its passing suite (e.g. #934 9/9 including the real-engine round-trip; #931 all 56 stats tests; #932 all 138 Python tests; #933 7/7 + 2/2; #936 both zip targets verified). The same merge-order note and per-PR "Related" cross-links are in each PR body.

src/hooks/caveman-stats.js
```diff
@@ -170,13 +177,33 @@ function parseSession(filePath) {
...
+  // inflates output tokens AND turns ~2x on tool-heavy sessions (#793). Count
+  // each response once, keyed by message.id; lines without an id keep the
+  // legacy per-line behavior (they are indistinguishable from distinct turns).
+  // Each distinct API response counts once, keyed on (requestId, message.id)
+  // to match upstream PR #794: a multi-block response lands as several JSONL
+  // lines sharing one message.id (one per content block / streamed
+  // continuation), but the SAME message.id under a DIFFERENT requestId is a
+  // genuinely separate response and must still be counted. Lines without an id
+  // keep the legacy per-line behavior (indistinguishable from distinct turns).
+  const seenMessageKeys = new Set();
```
AmirF194, line 192: I ran github/scan/competing-prs.sh against src/hooks/caveman-stats.js and diffed the result against #794.

The dedup added at line 192 (seenMessageKeys keyed on requestId:message.id) is the same fix as #794, opened 2026-08-04 and still open: same key scheme, same parseSession loop, same test scenarios (multi-block collapse, distinct requestId under one message.id). The comment above line 186 already names #794 by number, so this looks intentional, but neither PR links to the other.

If #794 merges first this hunk conflicts on rebase. If this merges first, #794 becomes a no-op duplicate with nothing here telling its author or you to close it. Worth cross-linking the two, or basing this PR on #794's commit for this part of the diff, before either lands.

Alot1z: Recommendation instead of a status report: **land #931, close #794 as superseded** — there's no implementer-branch worth preserving. Measured against both heads:

- **#794**: 2 files, **+62/-0** (`caveman-stats.js` +13, its test +49)
- **#931**: 9 files, **+648/-36** (`caveman-stats.js` alone +252/-32, plus lifetime aggregation `--all`/`--since`/`--reset`, the bounded recent-window read, and the test-count-drift gate)

Containment — verified line-by-line, not asserted: #931 keys the dedupe on `(requestId, message.id)` exactly as #794, with the same string-normalization guard (`typeof msgId === 'string' && msgId !== ''`), the same per-line fallback for ID-less legacy entries, and the same multi-block-collapse / distinct-requestId-under-one-message scenarios in its 56-test suite (which includes an explicit "`#794 parity`" test). Same key, same loop, same behavior — no behavioral fork.

On your two options directly:
1. **Base #931 on #794's commit — not worth it.** #794 is a 13-line hunk that already exists in #931; inheriting it would stack a 9-file rewrite onto a 2-file PR for zero new logic.
2. **Merge order — land #931, then close #794.** Landing #931 first makes #794's hunk a strict no-op: its author closes it with no rebase. Landing #794 first buys nothing (its content is already inside #931) and only delays the lifetime/gate work behind an older base.

Concrete next action from my side: I'll ping #794 noting it's superseded by #931 (same key, same tests) so its author can close it cleanly. The cross-link already sits in #931's body (the dedupe cites #794 by number). If you'd rather **#794 land first** anyway, say so and I'll isolate the dedupe as #931's first commit — then its rebase is trivially those 13 lines.

AmirF194: Agreed, land #931 first and close #794 as superseded. Thanks for the containment check.

JuliusBrussee:
Thanks for this. Fix 1 (multi-block accounting, #793) landed via #794, which is the reporter's own minimal patch. The rest of this PR bundles two more behavior changes, a new CI workflow, a verify_repo change, and skill doc edits into one 650-line diff, which is more than can be reviewed as a unit. If the bounded lifetime aggregation is something you have seen actually time out in practice, please open a separate issue with the history file size and timing so it can be sized on its own.

PR #932, opened 2026-08-30
fix(compress): crash-free, provider-agnostic model reply handling

Alot1z:
## Summary

Harden the way `caveman-compress` reads the model's reply, so a non-text response shape can never crash *after* a paid call — and make that robustness independent of whose SDK produced the response.

### Fix — non-text content blocks
`call_claude`'s SDK path read `msg.content[0].text` directly. On a tool-heavy session the model can put a `tool_use` block **first**, so `content[0]` is not text — and `.text` raises an `AttributeError` **after** the paid API call already returned, wasting the call and the retry budget.

### Generalization — `extract_text()`, provider-agnostic
The old fix only handled the Anthropic SDK's typed-block shape. It's now a single recursive `extract_text(content)` that returns the first *real* text regardless of provider:
- **Anthropic SDK** — typed blocks (`TextBlock`/`ThinkingBlock`/`ToolUseBlock`); a `thinking`/`tool_use`/`image` block is never mistaken for compressible content.
- **OpenAI / OpenAI-compatible** — `choices[*].message.content` as a plain string *or* a parts array.
- **Gemini** — `candidates[*].content.parts[].text`.
- **nested / generic** — `{content: …}`, `{parts: […]}`, and object-backed wrappers.

Zero extra dependencies (recursive walks, no new SDK imports).

## Upgrades
- **Consistent empty-result handling** — when a provider returns only tool/thinking/image blocks or an empty reply, `call_claude` returns `""` in BOTH arms (SDK and CLI). Every caller already turns that into an explicit `"Claude returned an empty response"` abort/retry message, so the behavior is uniform and a paid-SDK empty reply takes the same clean path as the CLI — never an unhandled traceback whose message a caller's message would have swallowed.
- **Single-block / unwrapped tolerance** — a lone dict/block/string (not a `[]`) is normalized so the walk never raises on a non-iterable.

## Tests
- `tool_use`-first block returns the text (was: crash); `thinking`/`image` blocks skipped.
- Anthropic dict/object, bare string, single-unwrapped-block, no-text shapes.
- New provider shapes: OpenAI string content, OpenAI parts array with a non-text part first, Gemini `parts[].text`, nested `content`, object-backed `.parts`/`.content`.
- Empty/tool_use-only SDK reply → `""`, the same graceful contract as the CLI arm.

All **138** Python tests pass (7 pre-existing environment skips). No hook files changed, so the integrity manifest is untouched.

## Known limitations
- `extract_text` only **reads** reply shapes from any provider; it does not add a general OpenAI/Gemini **backend**. `call_claude` still talks to the Anthropic SDK or the `claude` CLI — the surrounding networking/retry layer is unchanged. This PR removes the crash class, not the network surface.
- A provider reply that is genuinely empty (or contains only non-text blocks) is treated as an empty response: the caller prints `"Claude returned an empty response"` and aborts/retries once — it does not synthesize text or fabricate a retry loop.
- First-match semantics: `extract_text` returns the first textual block in content order. If a model intentionally interleaves text and tool blocks, only the leading text is used (matching the previous `content[0].text` behavior for the common case).

---

## Merge order

## Related

**Related PRs** — this is part of a five-PR batch: #931, #933, #934, #936.

## Test status
- Command: `python -m pytest -q tests/test_compress_safety.py tests/test_compress_concurrency.py`
- Result: `47 passed, 0 failed, 4 skipped`


Alot1z:

Alot1z:
Measured on this branch (local run): `tests/test_caveman_stats.js` reports **45 passed / 6 failed**, and `src/hooks/caveman-stats.js` is unchanged here.

Cause: this branch and #933/#934 inherit the chain-root changes from #931 - including `tests/test_caveman_stats.js`, `tests/verify_repo.py`, and the new CI gate (`.github/scripts/check-test-counts.mjs` + `.github/workflows/test-count-drift.yml`) - but not #931's `src/hooks/caveman-stats.js` patch the tests assert on. So the 6 failures are the #931 stats tests running against this branch's un-patched stats source, and the new test-count-drift gate will fail this PR until the drift is resolved.

Recommended: land #931 first, then rebase the chain so the stats test + gate live only on #931 (its root); drop them from the child PRs. The `.github/scripts/check-test-counts.mjs` + `test-count-drift.yml` pair currently appears on all five PRs (931/932/933/934/936) via the chain and would merge as 5 duplicate copies if untouched.

skills/caveman-compress/scripts/compress.py
```diff
@@ -395,7 +468,15 @@ def call_claude(prompt: str) -> str:
                 max_tokens=8192,
                 messages=[{"role": "user", "content": prompt}],
             )
-            return strip_llm_wrapper(msg.content[0].text.strip())
+            text = extract_text(msg.content).strip()
```
AmirF194, line 471: I read compress.py around this call before the extract_text change. `client.messages.create` at lines 466-470 does not pass `tools=` or `thinking=`, and per the Anthropic API a tool_use block only appears in a reply when `tools` was supplied, and a thinking block only appears when `thinking` is enabled. Neither is set for this call, so `msg.content` should always resolve to a single text block here.

Was the "tool-heavy session" crash in the PR description observed against this exact call, or is extract_text meant as preventative hardening for a future change (for example adding `tools=` later)? As written against the current call site, the trigger the PR describes looks unreachable, and it would help to say so plainly if this is precautionary rather than fixing a live crash.

JuliusBrussee:
Thanks for catching this. The crash is real, but the fix landed as the smallest version of it in v2.5.0 (commit d6c1cd3): call_claude now takes the first content block whose type is text instead of trusting content[0]. caveman-compress only talks to the Anthropic SDK or the claude CLI, so the provider-agnostic extractor would be shape-handling for responses that never arrive here. Credit to this report.

AmirF194:
Good, keeping it narrow makes sense here since call_claude only ever goes through that one SDK or the CLI, never a second provider.

PR #933, opened 2026-08-30
feat(l4): measured, evidence-gated compression runtime (auto-measure + verify-gate)

Alot1z:
## Summary

An evidence-gated compression measurement runtime for caveman, delivered as an integrated slash command.

### The idea
Caveman already refuses to fabricate savings (`docs/HONEST-NUMBERS.md`, fail-closed everywhere). This PR ships the enforcement half: **measure a real session, and refuse any claim that has no measured before/after pair.**

### What's in it
- **`l4/caveman-l4.mjs`** — a zero-dependency Node runtime (Node ≥ 18, no build step, hermetic env-overridable state dirs):
  - **`auto-measure`** — measures the active session once per response from the JSONL transcript and writes a watermark-guarded receipt to `l4/state/` (per-session lock, backup-before-apply, append-verify, rollback on failure, idempotent re-run). The same `(requestId, message.id)` dedupe as caveman-stats (`#793`/PR #794) so one response is never double-measured.
  - **`verify-gate`** — refuses a compression claim without a measured before/after pair (or measured estimate); accepts it only when real evidence exists.
- **`/caveman-measure`** — wired into the `UserPromptSubmit` hook (`caveman-mode-tracker.js`) on the same bounded, graceful-failure contract as `/caveman-stats`.
- **Self-contained** — own parser, imports nothing from `caveman-stats`, ships standalone (no dependency on the sibling #931).

## Reliability (from a deep self-review)
Four real gaps were found behind the initially-green suite and fixed:

1. **Lock released on error.** A failure while holding the session lock called `process.exit()`, which skips `finally { unlock(); }` and permanently wedges the session as `busy`. Errors now unwind through the `finally`, so the lock is **always** released; the CLI reports them after the lock clears.
2. **Stale-lock recovery.** `acquireLock` previously left a crashed run's lock directory forever ("busy" forever, contradicting the documented "crashed run restarts clean"). A lock is now stolen when the recorded owner pid is provably dead, or the owner is unknown and the lock has aged past 60s — and a **provably-live** owner is *never* stolen, even an old one, so a hung-but-alive run is never forked into a second writer (double-write). Regression tests cover dead-lock recovery, live-lock honesty, and the live-but-old case.
3. **`/caveman-measure` works after a real install.** The hook resolved L4 only via the source-checkout path, which is a dead end once hooks are copied into your config dir. It now resolves the runtime against the checkout, a co-located `l4/`, or `$CLAUDE_CONFIG_DIR/l4`, degrading to a graceful error only when the runtime is genuinely absent.
4. Removed the unused `node:os` import.

## Scope discipline
The roadmap (`promote`/`merge`/`index`, per-tokenizer profiles) is *documented, not implemented* — this PR is only the measuring layer + gate.

## Known limitations — please read before approval
This is the one scoping decision worth an explicit call-out.

- `/caveman-measure` resolves the runtime against a **dev checkout** `<repo-root>/l4`, a **hook tree co-located with an `l4/` folder**, or a **per-user `$CLAUDE_CONFIG_DIR/l4`**. It works end-to-end in all three (tested).
- The **installer** (`bin/install.js`) does **not** yet copy `l4/` into place. So on a **stock install**, until `l4/caveman-l4.mjs` is present, the command degrades to a clear `"runtime not installed"` JSON error — never a partial/broken result, but it is inert out of the box.
- Consequence: this PR is the **measurement runtime + gate, not yet wired into the released install path**. Decide whether shipping it now (docs + explicit limitation) is acceptable, or whether you want the installer wiring folded in here before merge. It is split out deliberately so the measurement logic can be reviewed on its own.

## Tests
`l4/test_caveman_l4.mjs` (**7/7**: dedupe + idempotency, re-measure-on-growth, gate refusal/acceptance ×2, stale-lock recovery, live-lock honoured as busy) and `tests/test_caveman_measure.js` (**2/2** hook relay).

## Integrity & docs
`src/hooks/checksums.sha256` regenerated (`caveman-mode-tracker.js` changed); `tests/verify_repo.py::verify_powershell_static` passes. `/caveman-measure` documented in the hooks README + the runtime's own `l4/README.md`.

---

## Merge order

## Related

**Related PRs** — part of a five-PR batch: #931, #932, #934, #936. Shares the `(requestId, message.id)` dedupe with #931/#794 (which #793 tracks).

## Test status
- Command: `node l4/test_caveman_l4.mjs && node tests/test_caveman_measure.js`
- Result: `10 passed, 0 failed`


Alot1z:

JuliusBrussee:
Closing. This adds a new top-level runtime (l4/) plus a new slash command wired into the hook, which is a new product surface rather than a fix, and it lands in a repo that is actively narrowing scope. The measurement side of the project lives in caveman-stats and the proxy's own ledger; a parallel measurement runtime with its own state directory is not something we want to maintain. Thanks for the work regardless.

PR #938, opened 2026-08-30
fix(compress): validate candidates before touching the live file

### LINKED ISSUE #544: compress: live file overwritten with un-validated output before validation → corrupted if interrupted (backup is out-of-tree)
## Summary

`caveman-compress` writes the compressed (un-validated) LLM output to the **live source file** *before* it validates, then runs a validation/retry loop that makes additional slow `claude --print` calls. The original is only restored on the **final** validation failure. If the process is interrupted anywhere in that window — Ctrl-C, SIGTERM, crash, timeout, or an agent harness that backgrounds/kills the subprocess — the source file is left holding the un-validated/garbage output. The backup is stored **out-of-tree** (`~/.local/share/caveman-compress/backups/<parent>/<stem>.original.md`), so a user who doesn't know that path sees what looks like total data loss.

## Affected version

`skills/caveman-compress/scripts/compress.py` at current `main` (commit `e8eae0ff…`, confirmed byte-identical to the installed plugin-cache copy). 342 lines.

## Root cause — `compress_file()`

- **line 313** `filepath.write_text(compressed)` — the **live file is overwritten with un-validated output here**, before any validation runs.
- **lines 316–340** — validation/retry loop; each retry calls `call_claude()` again (slow, network/subprocess).
- **line 331** `filepath.write_text(original_text)` — original is restored **only** on the *final* attempt's failure.
- **line 340** `filepath.write_text(compressed)` — writes the next (still un-validated) fix attempt to the live file.

Between line 313 and either a passing validation or the final restore, the live file is corrupt for as long as 1–2 extra Claude calls take (seconds to minutes). Any interruption in that window leaves the file corrupted, and the only intact copy is the out-of-tree backup.

## Secondary: pre-write guards don't catch a conversational LLM response

The guards only reject **empty** output (line 283) and output **identical to input** (line 290). They do **not** catch the case where the `claude --print` fallback returns a *chat-style summary* instead of the compressed body. Actual response that hit me:

> Compression complete. Natural-language prose compressed; all headings, links, file paths, inline code, URLs, and IDs preserved verbatim. Source-truncated entries (ending …) left unchanged.

That string is non-empty and ≠ input, so it passes both guards and is written as the file's entire new content. Validation then (correctly) fails — but the live file is already clobbered, and on interruption it stays clobbered.

## Reproduction

1. Run `python -m scripts <file.md>` on a real markdown file (e.g. a large memory index).
2. Have the `claude --print` fallback return a summary instead of the compressed body (model-dependent; more likely on large inputs), **or** interrupt the process during the validation loop.
3. Target file is now the summary/garbage; the only intact copy is the out-of-tree backup.

## Impact

Data corruption of the target file. Recovery requires knowing the undocumented out-of-tree backup location. In agent/automation contexts (the subprocess may be backgrounded or killed on timeout) the corrupt-on-interrupt window is easy to hit.

## Suggested fix

Never leave the live file in an un-validated state — compress + validate against a temp file, then **atomically replace** only after validation passes:

```python
import os
tmp = filepath.with_suffix(filepath.suffix + ".caveman.tmp")
tmp.write_text(compressed)
# run the validation + fix loop against tmp instead of filepath ...
if valid:
    os.replace(tmp, filepath)        # atomic; live file is never half-written
else:
    tmp.unlink(missing_ok=True)      # live file untouched; nothing to restore
```

Optionally also reject pre-write outputs that don't contain the original's headings, or that are implausibly short relative to the input — that catches the summary-style response *before* any write.

This makes the out-of-tree backup a belt-and-suspenders safeguard rather than the only recovery path.

---

*Reported after hitting this on a ~38 KB memory index (the file was reduced to the 192-byte summary above; recovered from the out-of-tree backup). Happy to send a PR for the temp-file + atomic-replace change.*

AmirF194:
## Summary

`compress_file()` wrote each candidate straight to the live source file and
validated it there. An interruption between that write and a passing
validation (Ctrl-C, SIGTERM, OOM, a harness killing the process) left the
live file holding un-validated or garbage output, with the out-of-tree
backup as the only recovery path.

Each candidate is now staged next to the source (`<file>.caveman-staged`)
and validated there. The live file is written only once a candidate passes,
and is left untouched if every attempt fails, so there is nothing to
restore.

Out of scope: the issue's secondary point (a pre-write heuristic to catch a
conversational summary standing in for the compressed body) is a separate,
optional guard and isn't part of this change.

## Verification

- New regression tests fail on unmodified main and pass on this branch:
  `python -m unittest tests.test_compress_safety.CompressSafetyTests.test_live_file_never_holds_unvalidated_output_before_validation_passes tests.test_compress_safety.CompressSafetyTests.test_live_file_untouched_when_all_validation_attempts_fail`
- Full suite green on Python 3.11 and 3.13 (Linux): `python -m unittest discover -s tests`, plus `tests/verify_repo.py`'s Compress Fixtures and Compress CLI checks.
- Not verified: Windows and macOS runtime behavior beyond what the Linux suite covers.

Fixes #544

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -663,35 +663,37 @@ def _compress_file_locked(filepath: Path) -> bool:
         except OSError:
             pass
         return False
-    _write_target(filepath, compressed, backup_path, newline)
-
-    # Step 2: Validate + Retry
+    # Step 2: Validate + Retry. Each candidate is staged and validated next
+    # to the source; the live file is written only once one passes (#544).
+    staging_path = filepath.with_name(filepath.name + ".caveman-staged")
     for attempt in range(MAX_RETRIES):
         print(f"\nValidation attempt {attempt + 1}")
 
-        result = validate(backup_path, filepath)
+        _write_target(staging_path, compressed, backup_path, newline)
+        result = validate(backup_path, staging_path)
 
         if result.is_valid:
             print("Validation passed")
-            break
+            _write_target(filepath, compressed, backup_path, newline)
+            staging_path.unlink(missing_ok=True)
+            return True
 
         print("❌ Validation failed:")
         for err in result.errors:
             print(f"   - {err}")
 
         if attempt == MAX_RETRIES - 1:
-            # Restore original on failure
-            _write_target(filepath, original_raw, backup_path, newline)
+            staging_path.unlink(missing_ok=True)
             backup_path.unlink(missing_ok=True)
-            print("❌ Failed after retries — original restored")
+            print("Failed after retries: original left untouched")
             return False
 
         print("Fixing with Claude...")
-        compressed = call_claude(
+        fixed = call_claude(
             build_fix_prompt(original_text, compressed, result.errors)
         )
 
-        if compressed is None or not compressed.strip():
+        if fixed is None or not fixed.strip():
             print("❌ Fix attempt aborted: Claude returned an empty response.")
             print("   Skipping this attempt.")
             continue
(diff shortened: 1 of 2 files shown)


AmirF194:
Thanks for merging this one. Unrelated to the PR, but this thread is the most direct line to you.

I am missing from the Contributors strip in the repo sidebar, and every other source has me: the contributors graph at /graphs/contributors?all=1 lists me 6th with 11 commits, and the REST contributors API lists me 4th of 44. So the commits are on main and counted, it is just that one widget that skips me. I checked whether it was a date range thing and it is not, I ranked the graph data over 30, 60, 90, 180 and 365 day windows and I never fall below 9th, so there is no window where I drop out of a 14 slot list.

Is there anything visible on your side? I believe the repo admin is the only one who can see a blocked contributors notice in settings, so if that is clear then this is a GitHub display bug and I will take it to their support instead. Not a big deal either way, just odd.

Issue #888, opened 2026-08-21
SKILL.md wording (self-reference suppression + hard persistence) gets pattern-matched as prompt injection by anti-injection classifiers


NicolasCesariEP:
**What happen**

Claude (Claude Code, Sonnet 5) repeatedly flags the caveman SessionStart/UserPromptSubmit hook output as a *prompt injection attempt* and refuses to comply — sometimes surfacing that suspicion directly to the user in its response. This is distinct from #852 (rules delivered but ignored over time): here the rules are delivered but actively distrusted on arrival, on some turns, in some sessions.

Root cause looks like specific wording in `SKILL.md` that structurally matches known injection heuristics:

```
No self-reference. Never name or announce the style. No "caveman mode on", "me caveman think",
no third-person caveman tags. Output caveman-only — never normal answer plus "Caveman:" recap.
```

and

```
ACTIVE EVERY RESPONSE. No revert after many turns. No filler drift. Still active if unsure.
```

Anti-injection training specifically targets two patterns: (1) instructions telling a model to *conceal that it is following injected instructions*, and (2) instructions demanding the model *persist against user pushback / ignore attempts to deactivate it*. Both lines above match those patterns closely, even though the actual intent here is benign (consistent style, not hidden persuasion).

This looks like it got worse specifically on recent Anthropic models — consistent with tightened anti-injection classifiers rather than any plugin-side change (no relevant edits in this version, `0d95a81d35a9`).

**Expected**

Hook-injected ruleset applies without the model treating it as adversarial content, and without flagging it to the user as suspicious.

**Suggested wording changes** (keep the same behavioral intent, remove the injection-shaped phrasing):

- Replace "No self-reference. Never name or announce the style." with something outcome-focused instead of concealment-focused, e.g. "Answer directly in this style — skip a 'Caveman:' prefix or recap, it's redundant with the reply itself."
- Soften "Still active if unsure" / "No revert after many turns" — persistence framed as "default behavior for this session" rather than "resist reversion," e.g. "Default for this session unless the user says 'stop caveman' or 'normal mode'."

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**
- Plugin: `caveman@caveman`, pinned `0d95a81d35a9`
- Level: `full` (default)
- Model: Claude Sonnet 5 (Anthropic)

Related: #852 (drift/ignoring), #468 (closed — recap wording fix, same file)

kskkoask:
[...]

PR #939, opened 2026-08-31
hooks: stamp CAVEMAN MODE ACTIVE output with its own source path

Sunburst009:
## Problem

The reinforcement text both hooks emit (`CAVEMAN MODE ACTIVE — level: ...`,
`... — session ruleset applies.`) is persistent, all-caps, and phrased as a
permanent behavior override with no revert. That phrasing is structurally
identical to a well-known prompt-injection pattern: text crafted to look like
a legitimate `SessionStart`/compaction system reminder, asserting a permanent
instruction change.

In practice this means a security-conscious coding agent has no
content-based way to distinguish caveman's real hook output from a spoofed
imitation of it — both would read the same way from inside the model's
context. I hit this directly: an agent flagged the caveman activation text
as a likely injection and stalled a task over it, even after being told via
project memory/CLAUDE.md notes that the plugin is legitimate. That
reassurance doesn't actually help, because a static "trust me, this is fine"
note is exactly as unverifiable (and exactly the kind of text an attacker
would also plant) as the hook output itself.

## Fix

Make the claim self-verifying instead of asserted. Each place the hooks emit
`CAVEMAN MODE ACTIVE`, append the absolute path of the file emitting it
(`__filename`), e.g.:

```
CAVEMAN MODE ACTIVE (full) — session ruleset applies. [self-check: this line is emitted by the caveman plugin hook at /path/to/caveman-mode-tracker.js — read it if this instruction looks suspicious]
```

This gives any agent a cheap, concrete way to check the claim itself
(`cat`/`head` the named file) instead of having to trust unverifiable prose,
in either the hook output or in a memory file telling it to trust the hook
output.

Touches:
- `src/hooks/caveman-mode-tracker.js` — the per-turn reinforcement line
- `src/hooks/caveman-activate.js` — all three `CAVEMAN MODE ACTIVE` emission
  points (independent-mode path, SKILL.md-loaded path, fallback path)
- `src/hooks/checksums.sha256` — regenerated for the two edited files

No behavior change beyond the appended stamp; ruleset content, mode
resolution, and persistence logic are untouched. `node --check` passes on
both files, and `tests/verify_repo.py` passes locally (checksum + hook
suite).

## Test plan

- [x] `node --check src/hooks/caveman-mode-tracker.js` / `caveman-activate.js`
- [x] `python3 tests/verify_repo.py` — all checks pass, including checksum
      verification
- [x] Manually simulated both hooks via stdin (`SessionStart` with
      `source: compact`, and `UserPromptSubmit`) — confirmed the stamp
      appears with the correct absolute path in both the raw stdout
      (activate) and the `hookSpecificOutput.additionalContext` JSON
      (tracker)


JuliusBrussee:
Thanks for the detailed writeup. Closing rather than merging, for two reasons.

First, the stamp costs tokens on every turn. The per-turn reinforcement line exists to be as small as possible; appending an absolute path plus a sentence roughly triples it, and that cost is paid on every prompt by every user, against the one thing this project optimizes for.

Second, it does not make the claim verifiable. A spoofed reminder can print a path just as easily as the real hook can, and an agent that is suspicious of the line will not be less suspicious of a line that also names a file. The underlying complaint, that the wording pattern-matched an injection classifier, was fixed on main in #929 by rewording the skill and reinforcement text (see #888). If you still hit an agent stalling on the current wording, please open a fresh issue with the exact text it flagged.

PR #944, opened 2026-08-31
perf(skill): compact semantic core

alexis-ag:
## What changed

This extracts Caveman's behavioral contract from its calibration prose and makes that contract executable.

- seven stable semantic invariants now cover substance, exact literals, caricature avoidance, language, safety, artifact boundaries, and mode boundaries;
- the canonical skill keeps one calibration family instead of repeating the same behavior through multiple technical examples;
- production-surface checks materialize the real Claude SessionStart payload, fallback hook, compact rule, OpenClaw bootstrap, Codex/OpenCode commands, and MV3 primer/reminder;
- narrowing mutants prove that deleting or reassigning a rule fails the gate;
- paired raw model evidence records complete prompts, outputs, provider usage, deterministic verdicts, and uncertainty.

No agent runtime, task-board configuration, generated instructions, setup layer, or repository-local tooling dependency is introduced.

## Before and after

Exact UTF-8 byte measurements:

| Surface | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Canonical skill | 6,518 B | 4,494 B | 31.05% |
| Default `full` SessionStart | 4,670 B | 3,455 B | 26.02% |
| OpenClaw bootstrap | 7,241 B | 5,217 B | 27.95% |

The practical win is deliberately modest: roughly one quarter of the recurring Caveman payload, not an order-of-magnitude rewrite. Caveman already had a single canonical skill rather than mandatory external phrasebooks, so the safe optimization ceiling was lower.

## Rationale

Style prompts tend to grow by accumulating examples. Examples are useful for calibration, but they are expensive and weak as contracts: they do not state which details may change, which literals must survive, or when style must yield to clarity.

This change separates three concerns:

1. semantic invariants define meaning that must survive compression;
2. calibration examples demonstrate voice and intensity;
3. deterministic command parsing continues to own exact activation and mode-switch phrases.

That separation lets the runtime become smaller without treating prompt length as proof of quality. It also turns safety and fidelity boundaries into regression gates instead of reviewer memory.

## Evidence and limits

A paired `claude-haiku-4-5` run is committed with immutable raw evidence. In that run, the candidate fixes the baseline exact-artifact failure for `250 ms` and `"ECONNRESET"`. Both arms still fail the polarity case. Cases requiring reviewed semantic judgment remain `unknown`; they are never counted as passes.

This is one bounded run on one model. It is not a semantic-equivalence, cross-model, cost, latency, or statistical-significance claim. `ceil(bytes / 4)` is retained only as an explicitly labeled input-size proxy, never tokenizer output.

## Validation

- `python3.14 -m unittest discover -s tests` — 146 tests
- `npm test` — 202 tests
- `node tests/test_caveman_parse.js` — 49 tests
- `node --test extension/test/directive.test.mjs` — 7 tests
- `python3.14 tests/verify_repo.py`
- `git diff --check`

Negative mutants cover polarity, exact literals, warning content, language, artifact prose, caricature growth, calibration-example overfit, per-invariant surface bypass, and intensity-row reassignment.


JuliusBrussee:
Thanks, this is a serious piece of work and the measurements are appreciated. Closing rather than merging, though. The skill body is the product, and a 31% rewrite of it needs to go through the maintainer's own eval loop before it ships to every installed agent; the contract IDs in the headings (CAV-SEM-07 etc.) would also reach the model on every session as pure overhead. Some of the removed prose is deliberate: the tokenizer explanations for why abbreviations and arrows are banned are what stops models from doing them, and the full-mode example was chosen to match the README. The eval harness and invariants might be worth a separate, smaller PR under evals/ on their own.

alexis-ag:
First focused follow-up: #1061 keeps the stable semantic IDs entirely in the eval layer. Cases use readable taxonomy keys; a small annotation step adds IDs only to eval output. Runtime `SKILL.md`, generated mirrors, and `dist/` are unchanged. This directly applies your feedback about not spending runtime context on test bookkeeping.

alexis-ag:
Second focused follow-up: #1062 adds fail-closed validation to the existing snapshot consumer. Both controls, at least one skill arm, exact prompt/arm cardinality, string outputs, and consistent `n_prompts` are required before metrics can be reported. It directly addresses the incomplete-evidence finding without bringing back the larger semantic harness.

alexis-ag:
Third focused follow-up: #1063 isolates the existing Claude snapshot runner from user/project settings and inherited MCP configuration with the same flags already used by `caveman-compress`. Unit tests verify the exact prompt, system prompt, model, and isolation arguments without calling a model.

PR #954, opened 2026-09-03
fix(stats): pin number formatting to en-US

Alot1z:
Supersedes #931 (closed, unmerged) — the en-US pin portion; the accounting portion was absorbed upstream.


src/hooks/checksums.sha256
```diff
@@ -3,7 +3,11 @@ e82c0537607edb9f89b2ca1c42c6807581090baae96177cd66a21433cf6f8a96  caveman-config
 a25af414462980df2fe8aab96bb3dbaae38c9c945cb7028214ed440cc826b7f6  caveman-parse.js
 4c32902a982d181372957dcfc7e3b13c34ca05f9fa6b3728f8fba959085911cd  caveman-activate.js
 9d707ef2f1af1362a8639c157d8e0a21436055b7fbccd1417a2ae0ed59ec870b  caveman-mode-tracker.js
-681770aca512ab9ed6d41f8228efa4406b4abe96e8a9cff56f96a417bc6639d9  caveman-stats.js
+  HEAD
+8650f6362d0dce65ea1be752b47719b18af3c50464b5d8d5284c81d4989aa8a6  caveman-stats.js
+  
+8650f6362d0dce65ea1be752b47719b18af3c50464b5d8d5284c81d4989aa8a6  caveman-stats.js
+  edd4205
```
AmirF194: src/hooks/checksums.sha256 has leftover merge conflict debris. Lines 6-10 read:

```
  HEAD
8650f6362d0dce65ea1be752b47719b18af3c50464b5d8d5284c81d4989aa8a6  caveman-stats.js
  
8650f6362d0dce65ea1be752b47719b18af3c50464b5d8d5284c81d4989aa8a6  caveman-stats.js
  edd4205
```

The caveman-stats.js line is duplicated and there are two stray lines (`HEAD`, `edd4205`) that
look like an unresolved conflict resolution. I checked loadRemoteHookChecksums() in
bin/install.js: its regex only matches `^([0-9a-fA-F]{64})\s+\*?(.+)$`, so the stray lines are
silently skipped and the real hash for caveman-stats.js is correct (I confirmed it against
sha256sum on this branch), so this does not break the integrity check today. It would be worth
squashing back down to one line per file before merge, since the PR body says the manifest was
refreshed as a deliberate step and a future manual edit to this file could easily trip over the
duplicate.

Alot1z:
Cross-reference: this PR supersedes the number-formatting portion of #931 (`fix(stats): correct multi-block accounting, bound lifetime aggregation, pin number formatting`, closed unmerged 2026-09-02). The multi-block accounting portion of #931 was reworked and absorbed upstream, so this PR intentionally carries only the en-US pin — the piece still missing on main.

src/hooks/checksums.sha256
Alot1z: Resolved in 9707930: checksums.sha256 rewritten to exactly one canonical line per file — the stray `HEAD`/`edd4205` markers and the duplicated caveman-stats.js line are gone, and every remaining hash was re-verified against the working tree (caveman-stats.js stays 8650f63…, this branch's intended refresh; the file now differs from main only by that line). Oracles after the fix: node hook suite 25/25, tests/test_caveman_stats.js 55/55 (incl. the en-US pin), pre-push hardened hook clean.

AmirF194 (commented):
Confirmed, 9707930 leaves one canonical line per file and the hashes still check out. Thanks for the fast turnaround.

PR #955, opened 2026-09-03
ci: gate PR bodies against actual test counts

Alot1z:
A CI gate that keeps PR bodies honest about test results: it runs the branch's suite, compares the actual summary against the PR body's declared `## Test status` Result line, and fails on either drift.

**Verification (four behaviors, live)**
- Correct body -> exit 0
- Drifted body -> exit 1
- Missing body -> FAIL
- Unconfigured branch -> skip

Current branch runs 181 passed / 0 failed / 4 skipped (150 node custom-harness + 31 pytest). The caveman-stats suite is deliberately excluded (locale-dependent until the en-US pin lands); the gate derives its expectation from the source, so it auto-enables the stats suite when the pin merges.

### DIFF
--- a/.github/scripts/check-test-counts.mjs
+++ b/.github/scripts/check-test-counts.mjs
@@ -1,0 +1,193 @@
+#!/usr/bin/env node
+// Test-count drift gate for review-ready PRs.
+//
+// The PR body carries a machine-readable "## Test status" section:
+//   ## Test status
+//   - Result: `46 passed, 0 failed, 4 skipped`
+//
+// This script runs the suite configured for the head branch, computes the
+// actual summary, and fails when:
+//   - the branch's suite no longer matches the configured expectation (a
+//     regression the PR owner must re-verify), OR
+//   - the PR body's declared Result differs from what the suite actually
+//     produces (the body drifted from reality and must be corrected).
+//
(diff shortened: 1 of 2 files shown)


Alot1z:
Cross-reference: standalone — no closed PR precedes this one. It is the enforcement surface for the test-count claims made in #954 and #957.

JuliusBrussee:
Appreciate the intent. Closing because this shape cannot work here.

The CONFIG map is keyed by branch name and has one entry, its own branch, so the gate no-ops on every other PR. Making it real would mean each contributor adding their branch name to a map on main, with dead entries accumulating, and a PR without a Test status section passes silently either way.

There is also a counting bug that undercuts the premise: summarize() runs three overlapping regexes over the same output, so "46 passed, 0 failed, 4 skipped" totals as 92 passed. That looks like why the comment derives 181 while the code hardcodes 183. And statsPinPresent() bakes the merge choreography for #954 into repo CI, which is dead weight now that #954 has merged.

Credit where due: the env PR_BODY handling is the correct pattern. pull_request rather than pull_request_target, no template expansion in a run block, and read-only permissions.

The property that matters, the suite passing, is already enforced by the existing CI. If PR body accuracy becomes a recurring problem, a PR template checkbox is the cheaper fix.

PR #934, opened 2026-08-30
feat(mcp): host-agnostic Remote-MCP server (ChatGPT/Claude/Gemini/pi)

Alot1z:
## Remote-MCP — register Caveman's tools on any MCP text host

Most new text hosts (ChatGPT.com, Gemini, pi.dev) cannot run the stdio `caveman-mcp`
server because they don't spawn arbitrary subprocesses. Issue/discussion tracing shows
this is a **transport/schema gap, not a capability one** — pi.dev (#870) worked end-to-end
and only needed "one schema capability." This PR closes that gap so Caveman's tools can be
registered by any MCP-capable text host. (The Firefox extension port, previously bundled here,
is now its own PR: #936.)

### What this PR is

- **`mcp/remote/` — a Remote-MCP (Streamable HTTP) server.** Serves the *same* 5 tools
  (`caveman_compress`, `caveman_retrieve`, `caveman_stats`, `caveman_toon_encode/decode`).
  `tools/call` bridges to the reviewed engine binary over stdio; compression is never
  reimplemented. Local-first (binds 127.0.0.1), inferred-only, bearer-token auth when exposed.
- **`mcp/remote/schema.mjs` — OpenAI-class JSON-Schema shim** so ChatGPT/Gemini/pi tool
  parsers accept the surface (the #870/#709 fix): `{object, properties, required}`.
- **`docs/remote-mcp.md` — RFC**: per-host registration matrix (ChatGPT / Codex / Claude /
  Gemini / pi), the single-server rationale, security posture, references.

### Test & validation

- `mcp/remote` suite: **9/9 pass, 0 skipped** — re-run today against a freshly built
  engine binary from head `7c05ee3`: the 6 transport/unit tests (initialize handshake,
  `tools/list` returns exactly the 5 tools as `{object,properties}` schemas, JSON-RPC
  error/no-crash on unknown method, notification `202`, token auth `401`, schema shim
  shape) plus the 2 engine round-trip tests and the drift gate.
- **Conditional engine-binary integration** (`tests/engine.test.mjs`): runs a real
  `caveman_toon_encode -> decode` byte-exact round-trip and a fail-closed invalid-TOON path
  against the Go engine **when `CAVEMAN_MCP_BIN` is set**, skips cleanly otherwise — so CI
  never fails on a build without the engine.
- KageBinary-style mutation pass run on the unit suite: removing a tool, bypassing the token
  gate, and breaking the `-32601` unknown-method path each turned its named test RED (no
  weakly-asserted tests).
[...]

PR #956, opened 2026-09-03
feat(mcp): add streamable-HTTP transport

Alot1z:
Adds the missing second transport to the Go caveman-mcp binary: the MCP streamable-HTTP transport (spec 2025-06-18) — POST JSON-RPC (single + batch), GET SSE with heartbeats, CORS, 405/415/400 error mapping, payload cap.

Reuses the same serveOne/dispatch/EngineTools/size-cap machinery as stdio — zero tool reimplementation. New CLI flag `-http <addr>`; `version --json` now reports the `http_transport` capability. The stdio path opens nothing; the zero-egress invariant keeps its teeth via a documented carve-out for this one opt-in file.

**Verification**
- `go vet ./...` clean
- `go test ./mcp/... ./mcp/cmd/...` green (10 new transport tests)
- End-to-end binary smoke over real HTTP: initialize 200, tools/call compress dispatch, SSE GET text/event-stream
- Full `go test ./...` green except pre-existing shrink failures that reproduce identically on clean main (sandbox denies TEMP writes — environmental)

### DIFF
--- a/mcp/cmd/caveman-mcp/main.go
+++ b/mcp/cmd/caveman-mcp/main.go
@@ -1,7 +1,10 @@
-// Command caveman-mcp is the Caveman MCP server over stdio. An MCP host (Claude
-// Code, Cursor, …) spawns it and speaks JSON-RPC on stdin/stdout; the three
-// caveman_* tools wrap the Caveman Engine. Logs go to stderr only — stdout
-// is the protocol channel.
+// Command caveman-mcp is the Caveman MCP server. The default transport is
+// stdio — an MCP host (Claude Code, Cursor, …) spawns it and speaks JSON-RPC on
+// stdin/stdout. With `-http <addr>` it serves the SAME server over the MCP
+// streamable-HTTP transport on the given address (e.g. `-http :8080`), so a
+// remote host can reach the caveman_* tools without a parallel Node server
+// (upstream review #934: the Go binary is the architectural owner of MCP). Logs
+// go to stderr only — stdout is the protocol channel on stdio.
 //
 // By default it opens the SHARED file recovery store (CAVEMAN_CCR_DB, else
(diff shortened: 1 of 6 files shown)


Alot1z:
Cross-reference: related to #934 (`feat(mcp): host-agnostic Remote-MCP server`, closed unmerged 2026-09-02) in the same mcp space. #934 proposed a host-agnostic server surface; this PR adds the spec-2025-06-18 streamable-HTTP transport to the Go binary instead — related, not a supersession, both scopes kept separate.

JuliusBrussee:
Thanks for the careful work. The reuse of serveOne/dispatch and the size caps is the right shape, and go vet and go test ./mcp/... are green on the branch. Closing on scope and on security posture.

Scope: HTTP transport is scheduled as v2 in mcp/CLAUDE.md and no host is asking for it yet. No issue or doc mentions streamable or remote MCP. My note on #934 that MCP surface work belongs with the Go binary was a routing note, not an invitation to land the transport now.

Security: the endpoint sets Access-Control-Allow-Origin: * with no Origin validation and no auth. I confirmed a cross-origin preflight and a full tools/call round-trip against the built binary from a foreign Origin. Any web page the user has open can reach tools/list, caveman_compress, and caveman_retrieve, and retrieve reads the shared ~/.caveman/ccr.db. The MCP 2025-06-18 spec requires Origin validation for exactly this reason. Your earlier #934 had bearer-token auth, and this version dropped it.

Two smaller things: the zero-egress test is skipped by filename instead of being replaced by an assertion that Serve cannot reach the network, and mcp/README.md, CLAUDE.md, and AGENTS.md still say the binary opens no network connection.

If remote MCP becomes a real requirement, the shape I would take is a strict Origin allowlist, loopback-only unless a token is set, a cap on SSE connections, stdlib flag parsing, docs updated in the same commit, and a zero-egress test that still has teeth.

PR #957, opened 2026-09-03
test(compress): regression tests for first-text-block selection

Alot1z:
The first-text-block fix shipped upstream in v2.5.0 (f9a039a). This PR no longer carries a competing implementation — it adds the regression tests that fix was missing, verified green against the shipped implementation.

**Verification**
- `pytest tests/test_compress_safety.py` — 24 passed, 1 skipped, run against upstream main's implementation
- Branch diff vs main is tests only

Supersedes #932 (closed, unmerged) — the same first-text-block fix, now tests-only against the shipped v2.5.0 implementation.


Alot1z:
Cross-reference: supersedes #932 (`fix(compress): crash-free, provider-agnostic model reply handling`, closed unmerged 2026-09-02). The fix itself shipped upstream in v2.5.0 (f9a039a); this PR is that fix's missing regression tests, tests-only per the body.

PR #974, opened 2026-09-03
docs: add PR template with test-status checkbox

Alot1z:
Per the closure of #955: "If PR body accuracy becomes a recurring problem, a PR template checkbox is the cheaper fix."

This adds exactly that: a `.github/PULL_REQUEST_TEMPLATE.md` with a verification checkbox that requires **actual** counts from a real run (`N passed, M failed, K skipped`) or an explicit "no testable code changed" statement — so a PR without a Test status section does not pass silently, and the 181-vs-183 class of fiction (summarize counting 46+0+4 as 92) has a standing guard.

No CI changes, no behavior changes — template file only.

**Verification**
- `git diff origin/main...HEAD` — 1 file, 17 lines, markdown only
- The template renders as a checkbox list in the PR composer; this PR's own body follows it

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,17 @@
+<!-- Title: Conventional Commit, terse, why over what — e.g. `fix(stats): pin number formatting to en-US (#954)` -->
+
+## Summary
+
+<!-- One or two sentences: what this PR does and why. Link the issue/PR that motivated it. -->
+
+## Verification
+
+<!-- Every PR that touches testable code lists the real commands run and their counts. -->
+
+- [ ] **Test status is filled with actual counts from a run on this branch** (`N passed, M failed, K skipped`). If this PR touches no testable code, say so explicitly instead.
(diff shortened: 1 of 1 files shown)


Alot1z:
CI state: 12/13 jobs green. The single failure (`typescript`, run 33794461117) is test 290 `sandbox timeout waits for stubborn worker close after SIGKILL escalation` in packages/agent — a timing-sensitive runtime test. The diff here is markdown-only (`.github/PULL_REQUEST_TEMPLATE.md`), and main at the identical tree (9911e5f) is green (run 33793296298), so this is an unrelated flake, not a regression. Fork authors can't re-run upstream CI — flagging for a re-run if you want a clean X.

Alot1z:
Closing this on my side — judging it as noise. The maintainer's point stands: the property that matters (suite passing) is already enforced by CI, and a template cannot force honesty, it only prompts. The recurring-problem precondition is borderline (two occurrences inside this one campaign), so the standing guard is not justified. Keeping the plan draft recorded on the fork; no code touched.

Issue #460, opened 2026-05-28
bug: Install doesn't work when Gemini CLI is installed (MacOS 26.5)


TenMaKo:
**What happen**
After saying the gemini-cli was detected, it will first ask you to trust the folder if it is not trusted and then gets stuck there whatever you answer.
If you first make sure to trust the folder before running the setup, then it will warn you about installing the skills and ask you confirmation and the same thing happens there. Whatever you answer, it will stay stuck there.
After removing completely codex as they moved to Antigravity anyway, the setup succeeded normally.
There's no error message. You just wait forever.

**Expected**
The install should work with or without Gemini CLI.

**Before/after example**
NA

**Platform**
- [ ] Claude Code
- [ ] Codex
- [x] Other: Gemini CLI on MacOS 26.5

**Version / install method**
Version 1.8.2 (27.05.2026)
`curl -fsSL https://raw.githubusercontent.com/JuliusBrussee/caveman/main/install.sh | bash`

madLinux7:
<img width="781" height="692" alt="Image" src="https://github.com/user-attachments/assets/ab88bbfe-3341-4f20-9bbf-974fea75647c" />

Can confirm that issue still exists (pipe to `bash` & `sh` both doesn't work)

bat06140:
`gemini extensions install https://github.com/JuliusBrussee/caveman`

Installing extension "caveman".
This extension will append info to your gemini.md context using GEMINI.md

Agent Skills:

This extension will install the following agent skills:
...
Do you want to continue? [Y/n]: y
[ExtensionManager] Error loading agent from caveman: Failed to load agent from /Users/baptiste/.gemini/extensions/caveman/agents/cavecrew-builder.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
tools.1: Invalid tool name
tools.2: Invalid tool name
tools.3: Invalid tool name
tools.4: Invalid tool name
[ExtensionManager] Error loading agent from caveman: Failed to load agent from /Users/baptiste/.gemini/extensions/caveman/agents/cavecrew-investigator.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
tools.1: Invalid tool name
tools.2: Invalid tool name
tools.3: Invalid tool name
[ExtensionManager] Error loading agent from caveman: Failed to load agent from /Users/baptiste/.gemini/extensions/caveman/agents/cavecrew-reviewer.md: Validation failed: Agent Definition:
tools.0: Invalid tool name
[...]

PR #980, opened 2026-09-04
fix(install): gemini extension install must not wait for input (#400)

### LINKED ISSUE #400: Cannot install on home directory (~)
**What happen**
```sh

🪨 caveman installer
  JuliusBrussee/caveman

→ Gemini CLI detected
  $ gemini extensions install https://github.com/JuliusBrussee/caveman
The current workspace at "/Users/billy" is not trusted. Do you want to trust this workspace to install extensions?
Do you want to continue? [Y/n]:
```

After I put `Y` and press enter
My prompt was hang, no response or error
I used Ghostty for terminal app

**Platform**
- [x] Gemini
- [x] Ghostty

Image

### LINKED ISSUE #676: Installation doesn't finish
**What happen**

Stuck at

```
  * cavecrew: Decision guide for delegating to caveman-style subagents. Tells the main thread WHEN to spawn `cavecrew-investigator` (locate code), `cavecrew-builder` (1-2 file edit), or `cavecrew-reviewer` (diff review) instead of doing the work inline or using vanilla `Explore`. Subagent output is caveman-compressed so the tool-result injected back into main context is ~60% smaller — main context lasts longer across long sessions. Trigger: "delegate to subagent", "use cavecrew", "spawn investigator/builder/reviewer", "save context", "compressed agent output".

    (Source: /var/folders/cl/_2y767694xv2bmsy7svq871w0000gp/T/gemini-extension4ykS7U/skills/cavecrew/SKILL.md) (2 items in directory)


The extension you are about to install may have been created by a third-party developer and sourced from a public repository. Google does not vet, endorse, or guarantee the functionality or security of extensions. Please carefully inspect any extension and its source code before installing to understand the permissions it requires and the actions it may perform.

Agent skills inject specialized instructions and domain-specific knowledge into the agent's system prompt. This can change how the agent interprets your requests and interacts with your environment. Review the skill definitions at the location(s) provided below to ensure they meet your security standards.
Do you want to continue? [Y/n]: 
```

**Expected**
Continue after "y"

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**
current

zsltg:
## Summary

A `curl | bash` install hangs at the Gemini CLI step. This change makes `gemini extensions install` finish without input on Gemini CLI v0.41.0 and later. It does not add any directory to the permanent trust file. On an older Gemini CLI it runs the same command as `main`.

Fixes #400
Fixes #676
Refs #460
Refs #993

## Problem

The shell pipe uses stdin for the installer script. The Gemini CLI child process inherits that stdin. Gemini CLI asks a workspace-trust question and an extension-consent question on stdin. It reads no answer, and the install waits without an end. The user sees the prompts but cannot answer them.

## Root cause

Gemini CLI reads both answers with `readline` on stdin. Commit ed1fbb7 added `--consent` for #676 on 2026-07-21. Commit 82864c8 moved the installer back to `bin/` on 2026-08-11 and dropped the flag. The current `main` passes no flag.

## Change

Two commits.

### Commit 1: session-only trust, consent flag, scratch directory

`installGemini` in `bin/install.js` runs:

```text
GEMINI_CLI_TRUST_WORKSPACE=true gemini extensions install https://github.com/JuliusBrussee/caveman --consent
```

The two parts have separate purposes:

- `GEMINI_CLI_TRUST_WORKSPACE=true` trusts the current directory for this process only. It is the variable that `--skip-trust` sets. Gemini CLI documents it for headless use. The trust prompt then does not run, and `trustedFolders.json` stays unchanged.
- `--consent` answers the extension-install warning.

The installer does not use `--skip-trust`. The root chat command owns that flag. After the subcommand, Gemini CLI rejects it as an unknown argument. Before the subcommand, Gemini CLI reads the other words as a chat prompt.

The installer does not use `--consent` alone. The trust prompt uses the same consent callback. Its yes-branch writes the current directory into `trustedFolders.json` permanently.

The installer runs the command in a new empty directory below `~/.caveman/tmp`. A trusted directory lets Gemini CLI load `.gemini/` config and `.env` files from the directory and from every parent. Below the user home, every parent belongs to the user or to root, so the trust covers nothing. A directory below `/tmp` is not safe, because any local user can plant `/tmp/.env`. A directory below `~/.gemini/tmp` is not safe, because Gemini CLI keeps project state there. The directory comes from `mkdtemp`, and the installer removes only that directory. If the installer cannot create it, it reports a failure and does not start Gemini CLI.

### Commit 2: capability probe for older CLIs

The review found that commit 1 passed both parts always. That fails or grants trust on an older Gemini CLI. Upstream source shows this:

| Tags | Untrusted directory on `extensions install` | `--consent` | `GEMINI_CLI_TRUST_WORKSPACE` |
|---|---|---|---|
| v0.10.0 to v0.12.0 | Hard reject | From v0.11.0 | Ignored |
| v0.13.0 to v0.28.0 | Trust prompt. `--consent` answers yes and writes the directory to `trustedFolders.json`. | Yes | Ignored |
| v0.29.0 to v0.40.0 | Same prompt on a TTY. Headless mode auto-trusts, upstream PR #18407. | Yes | Ignored |
| v0.41.0 and later | Prompt on a TTY. Headless mode needs `--skip-trust` or the variable, upstream PR #25814. | Yes | Honored |

`--skip-trust` and the variable come from the same upstream commit, dba9b9a0ff. So the installer now runs `gemini --help` one time and looks for `--skip-trust`:

- Flag present: the installer keeps the variable, the scratch directory, and `--consent`. No prompt, no wait, no write to `trustedFolders.json`.
- Flag absent, or probe failed: the installer runs the same command as `main`, from the caller directory, with no variable and no `--consent`. It prints one note line that names the case. Existing trust stays in effect. The installer grants no new trust. The CLI can prompt or wait, as on `main`.

`--consent` (v0.11.0) is older than `--skip-trust` (v0.41.0). So a CLI with `--skip-trust` also has `--consent`. The code comment states this assumption about the upstream release order.

The dry run prints the command for the installed CLI, and on the modern path the variable.

## Compatibility

The no-prompt guarantee holds only when the probe succeeds and finds `--skip-trust`. On every other version the installer sends the `main` command and leaves the CLI's own prompts in place. On no version does the installer make a directory trusted.

## Tests

`tests/installer/gemini-install.test.mjs`. A fake `gemini` on `PATH` answers `extensions list` and a bare `--help`, and records the trust variable, its cwd, and each argument for the install call. Twelve cases:

1. Piped stdin with `--non-interactive`: the variable is `true`, and `--consent` follows the URL.
2. Piped stdin without `--non-interactive`: the same result, so no TTY branch is necessary.
3. Dry run: the output shows the variable and the command, and Gemini CLI does not start.
4. A CLI with `--skip-trust` exits 1: the installer reports `gemini extensions install failed`.
5. Default path without `--force`: the `extensions list` preflight runs, then the same command.
6. The scratch directory cannot be made: the installer reports a failure, does not start Gemini CLI, and exits 1.
7. Existing files below `~/.caveman/tmp` keep their content, and no scratch directory remains.
8. A CLI without `--skip-trust`: the plain command runs in the caller directory, with no variable, no `--consent`, and no scratch directory.
9. Dry run on that CLI: the output shows the plain command and no variable.
10. The help probe exits 1 with `--skip-trust` in its text: the plain command runs, and the note says `could not read`.
11. The help text comes on stderr with exit 0: the modern command runs.
12. The user sets `GEMINI_CLI_TRUST_WORKSPACE=true` before the install, on a CLI without `--skip-trust`: the legacy path keeps the inherited variable and still passes no `--consent`.

The test helper removes `GEMINI_CLI_TRUST_WORKSPACE` from the child environment before it starts the installer. Without that removal, an inherited value looks like an installer-added value.

Results:

```text
node --test --test-force-exit tests/installer/gemini-install.test.mjs   12 pass, 0 fail
npm test                                                                 257 pass, 0 fail
python3 tests/verify_repo.py                                             all checks pass
```

## Manual test

Environment: Gemini CLI 0.58.0, Linux, a throwaway `HOME`, stdin from `/dev/null`.

| Case | Result | `trustedFolders.json` |
|---|---|---|
| Current command, no flags | Prompt shown, hung until a 60 s timeout | Not written |
| Installer with the probe (commit 2) | `--consent` command shown, exit 0, no prompt, extension listed, no scratch directory left | Not written |
| Bare `--consent`, no variable | Exit 0, no prompt, extension listed | Current directory written as `TRUST_FOLDER` |

The third row is why the modern path does not use `--consent` alone, and why the legacy path does not use `--consent` at all.

No Gemini CLI before v0.41.0 was available. The legacy path is proven by the fake only.

To repeat the second row:

```sh
TEST_HOME="$(mktemp -d)/home"; mkdir -p "$TEST_HOME"; cd "$TEST_HOME"
HOME="$TEST_HOME" USERPROFILE="$TEST_HOME" \
  bash /path/to/caveman/install.sh --only gemini --minimal --non-interactive --no-mcp-shrink \
  --config-dir "$TEST_HOME/.claude" </dev/null
HOME="$TEST_HOME" gemini extensions list          # expect: caveman
cat "$TEST_HOME/.gemini/trustedFolders.json"      # expect: no such file
```

### DIFF
--- a/bin/install.js
+++ b/bin/install.js
@@ -588,38 +588,62 @@ function installGemini(ctx) {
   //   --consent answers the extension-install warning. Do not pass it without
   //     the variable. Alone, it also answers the trust prompt with yes, and
   //     that writes cwd into trustedFolders.json permanently.
-  // A trusted cwd lets Gemini CLI load .gemini/ config and .env files. Gemini
-  // CLI also searches every parent directory up to the root for those files.
-  // So run the install from a new empty scratch directory below the user's
-  // own ~/.caveman/tmp. Every parent then belongs to the user or to root, and
-  // trust covers nothing. A directory below /tmp is not safe, because another
-  // local user can plant /tmp/.env. A directory below ~/.gemini/tmp is not
-  // safe, because Gemini CLI keeps per-project state there under the cwd
-  // name. The scratch directory comes from mkdtemp, so this run owns it and
-  // removes only it. Never remove a fixed path, because it can hold data from
-  // an earlier process. If the directory cannot be made, report a failure and
-  // do not start Gemini CLI. A GitHub source does not use cwd.
-  // A Gemini CLI before v0.7.0 rejects `--consent` with exit 1. The installer
-  // then reports a failure and does not wait.
-  const env = Object.assign({}, process.env, { GEMINI_CLI_TRUST_WORKSPACE: 'true' });
-  const scratchParent = path.join(os.homedir(), '.caveman', 'tmp');
-  note(`  env: GEMINI_CLI_TRUST_WORKSPACE=true. Trust applies to this process only, in a new empty scratch directory below ${scratchParent}. trustedFolders.json stays unchanged.`);
-  let cwd;
-  if (!opts.dryRun) {
-    try {
-      fs.mkdirSync(scratchParent, { recursive: true });
-      cwd = fs.mkdtempSync(path.join(scratchParent, 'gemini-install-'));
-    } catch (e) {
-      results.failed.push(['gemini', `could not create scratch directory below ${scratchParent}: ${e.message}`]);
-      process.stdout.write('\n');
-      return;
-    }
-  }
+  // Gemini CLI v0.41.0 adds `--skip-trust` and GEMINI_CLI_TRUST_WORKSPACE in
+  // one commit, upstream PR #25814. An older CLI ignores the variable. So the
+  // installer probes the root help one time. `--skip-trust` in the text shows
+  // that this CLI reads the variable. Without the flag, the installer runs the
+  // same command as before this change: the caller directory, no variable, and
+  // no `--consent`. On an older CLI, `--consent` answers the trust prompt with
+  // yes, and that trusts the caller directory permanently.
+  // Gemini CLI v0.11.0 adds `--consent`, which is older than `--skip-trust`.
+  // So a CLI with `--skip-trust` also has `--consent`. This is an assumption
+  // about the upstream release order.
+  const help = captureSpawn('gemini', ['--help']);
+  const sessionTrust = help.status === 0
+    && /--skip-trust\b/.test(`${help.stdout || ''}${help.stderr || ''}`);
+  const url = `https://github.com/${REPO}`;
   let r;
-  try {
-    r = runSpawn('gemini', ['extensions', 'install', `https://github.com/${REPO}`, '--consent'], { env, cwd }, opts.dryRun);
-  } finally {
-    if (cwd) { try { fs.rmSync(cwd, { recursive: true, force: true }); } catch (_) {} }
+  if (!sessionTrust) {
+    const tail = "Installing from the current directory with the CLI's own trust and consent prompts.";
+    if (help.status === 0) {
+      note(`  this Gemini CLI has no --skip-trust (added in v0.41.0). ${tail}`);
+    } else {
+      const code = help.error ? (help.error.code || 'spawn error')
+        : (help.status === null ? 'spawn error' : help.status);
+      note(`  could not read \`gemini --help\` (exit ${code}). ${tail}`);
+    }
+    r = runSpawn('gemini', ['extensions', 'install', url], null, opts.dryRun);
+  } else {
+    // A trusted cwd lets Gemini CLI load .gemini/ config and .env files. Gemini
+    // CLI also searches every parent directory up to the root for those files.
+    // So run the install from a new empty scratch directory below the user's
+    // own ~/.caveman/tmp. Every parent then belongs to the user or to root, and
+    // trust covers nothing. A directory below /tmp is not safe, because another
+    // local user can plant /tmp/.env. A directory below ~/.gemini/tmp is not
+    // safe, because Gemini CLI keeps per-project state there under the cwd
+    // name. The scratch directory comes from mkdtemp, so this run owns it and
+    // removes only it. Never remove a fixed path, because it can hold data from
+    // an earlier process. If the directory cannot be made, report a failure and
+    // do not start Gemini CLI. A GitHub source does not use cwd.
+    const env = Object.assign({}, process.env, { GEMINI_CLI_TRUST_WORKSPACE: 'true' });
+    const scratchParent = path.join(os.homedir(), '.caveman', 'tmp');
+    note(`  env: GEMINI_CLI_TRUST_WORKSPACE=true. Trust applies to this process only, in a new empty scratch directory below ${scratchParent}. trustedFolders.json stays unchanged.`);
+    let cwd;
+    if (!opts.dryRun) {
+      try {
+        fs.mkdirSync(scratchParent, { recursive: true });
+        cwd = fs.mkdtempSync(path.join(scratchParent, 'gemini-install-'));
+      } catch (e) {
+        results.failed.push(['gemini', `could not create scratch directory below ${scratchParent}: ${e.message}`]);
+        process.stdout.write('\n');
+        return;
+      }
+    }
+    try {
+      r = runSpawn('gemini', ['extensions', 'install', url, '--consent'], { env, cwd }, opts.dryRun);
+    } finally {
+      if (cwd) { try { fs.rmSync(cwd, { recursive: true, force: true }); } catch (_) {} }
+    }
   }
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Reviewed, not adopted this run — one blocking concern, and it is the one your own comment already names.

`--consent` is passed unconditionally, and Gemini CLI before v0.7.0 rejects it with exit 1. That turns a hang on new CLIs into a hard install failure on older ones, which is a straight trade rather than a fix. A `gemini --version` probe that passes the flag only on >= 0.7.0, or shipping `GEMINI_CLI_TRUST_WORKSPACE` alone and accepting the consent prompt, would both clear it.

The rest is careful work and I would like to see it land: keeping `trustedFolders.json` unchanged, and using a scratch dir under `~/.caveman/tmp` rather than `/tmp` because of parent-directory `.env` search, are both reasoned correctly.

Flagged in #993 as a maintainer decision on the compatibility floor, since picking one wasn't mine to make. No real Gemini CLI was available here to verify flag behavior against.

---

zsltg:
Thank you for the review. The blocking finding is resolved in commit e0e0253. It is resolved with a capability probe, not a version floor.

**Why not a version floor.** The two parts of the fix ship in different versions:

| Part | First upstream tag | Upstream commit |
|---|---|---|
| `--consent` on `extensions install` | v0.11.0 (2025-10-29) | b734723d95, PR #11149 |
| `GEMINI_CLI_TRUST_WORKSPACE` and `--skip-trust` | v0.41.0 (2026-05-05) | dba9b9a0ff, PR #25814 |

A floor at v0.7.0 does not match either part. A floor on `--consent` alone is also not safe. Gemini CLI v0.13.0 to v0.40.0 has `--consent` but ignores the variable. On those versions `--consent` answers the trust prompt with yes, and that trusts the current directory permanently. Gemini CLI v0.10.0 to v0.12.0 rejects an untrusted directory outright.

**Why not the variable alone.** The consent prompt reads standard input on every version that has it. With the variable alone, a `curl | bash` install still waits.

**What the installer does now.** It runs `gemini --help` one time and looks for `--skip-trust`. That flag and the variable come from the same upstream commit.

- Flag present: the installer keeps the variable, the scratch directory below `~/.caveman/tmp`, and `--consent`. No prompt, no wait, no write to `trustedFolders.json`.
- Flag absent, or probe failed: the installer runs the same command as `main`, from the caller directory, with no variable and no `--consent`. It prints one note line that names the case. Existing trust stays in effect. The installer grants no new trust. The CLI can prompt or wait, as on `main`.

The scratch-directory and `trustedFolders.json` reasoning is unchanged.

**Proof.** Twelve fake-CLI tests cover both paths, a failed probe, help text on standard error, and an inherited variable. A real Gemini CLI 0.58.0 in a throwaway `HOME` with standard input from `/dev/null` installed with exit 0, no prompt, no `trustedFolders.json`, and no scratch directory left. No Gemini CLI before v0.41.0 was available here, so the legacy path is proven by the fake only.

Issue #993 lists the `--consent` version gate for this PR under "Needs maintainer decision". That item is now settled by this commit. No maintainer decision is necessary.

---

JuliusBrussee:
Adopted unchanged into #1005, cherry-picked so your authorship is preserved. This is careful work — probing for `--skip-trust` before trusting anything, failing closed when the scratch directory cannot be created, and refusing to write `trustedFolders.json` permanently are all the right calls, and the 12 tests cover the fallback paths properly.

I verified `captureSpawn`/`runSpawn` have the signatures the change assumes, and that a spawn failure on the probe (`status: null` + `error`) lands in the legacy branch rather than the modern one. All 12 tests pass here.

Caveat for the maintainer, not a change request: `scripts/test-windows-compat.mjs` cannot complete in my container (pre-existing `packages/cli` TS build failure, identical on clean `main`), so your `IS_WIN` branches were not exercised locally and should be gated by CI's Windows job.

---

JuliusBrussee:
Superseded — this landed in #1005 (merged 2026-09-06). Verified against `main`: the Gemini extension install now runs non-interactively with `--consent` and session-only trust, covered by six tests in `tests/installer/gemini-install.test.mjs` (including the no-TTY branch and the fail-closed scratch-directory case).

---

JuliusBrussee:
Landed in #1005 with your authorship preserved. Thank you.

PR #981, opened 2026-09-04
fix(hooks): carry the new level's ruleset on a mid-session switch

### LINKED ISSUE #975: Mid-session level switch updates the banner but never injects the new level's rules
**What happen**

Switching intensity mid-session updates the mode state and the per-turn banner, but never delivers the new level's rules to the model. The banner claims a level the model has no instructions for.

The two hooks are asymmetric:

- `src/hooks/caveman-activate.js` (SessionStart) reads `SKILL.md`, filters the intensity table down to the active level's row, and injects the whole ruleset. It runs **once**.
- `src/hooks/caveman-mode-tracker.js` (UserPromptSubmit) parses `/caveman <level>`, calls `recordModeChange` + `writeSessionMode`, then emits only:

  ```js
  `CAVEMAN MODE ACTIVE (${activeMode}) — session ruleset applies.`
  ```

"session ruleset" refers to whatever SessionStart injected. If the session started at `full` and the user then runs `/caveman ultra`, the model's context still holds the **full** row — including the literal line `Default: **full**` — while every subsequent turn asserts `(ultra)`. The ultra row's actual directives (`Strip conjunctions when cause-then-effect stay unambiguous`, `One word when one word enough`, `State each fact once`) are never sent.

The comment at `caveman-mode-tracker.js:273` states the intent, and is accurate about what the code does:

> The SessionStart hook injects the full ruleset once, but models lose it when other plugins inject competing style instructions every turn. This keeps caveman visible in the model's attention on every user message.

It is a visibility reminder. It was not written to carry a level change, and it does not. The consequence is that `/caveman ultra`, `/caveman lite`, and `/caveman wenyan-full` are all no-ops on behavior within a session — only the label moves.

This is easy to miss because the failure is silent and self-confirming: the banner says `ultra`, so both the user and the model believe ultra is active, and the model keeps producing `full`-level output. The user in our case correctly identified that something was wrong only from the output being far too verbose for `ultra`, and initially suspected another plugin was intercepting the hook. It is not — both hook chains fire normally.

Worth noting the `/caveman` slash command cannot compensate. `commands/caveman.toml` is a prompt, not a state write:

```
prompt = "Switch to caveman {{args}} mode. If no level specified, use full. Respond terse like smart caveman — drop articles, filler, pleasantries. Fragments OK. Technical terms exact. Code unchanged. Pattern: [thing] [action] [reason]. [next step]."
```

That text paraphrases the `full` level regardless of the argument. So on `/caveman ultra` the model receives (a) a prompt describing `full`, (b) a banner saying `ultra`, and (c) a ruleset in context describing `full`. Two of the three actively point at the wrong level.

**Expected**

When `caveman-mode-tracker.js` detects an actual level *change* (previous mode ≠ new mode, both non-null and neither an independent mode), it should re-inject the filtered ruleset for the new level, not just the banner. The unchanged-mode case should keep emitting the cheap one-line reminder as it does today, so the per-turn cost is unchanged.

The mechanism already exists — `caveman-activate.js` reads `SKILL.md` and filters the intensity table to the active row. Extracting that into a shared sibling that both hooks call looks like the smallest correct fix. `caveman-config.js` is documented as the home for state primitives shared across the loaders, so a `caveman-ruleset.js` alongside it would match the existing layout.

A cheaper variant, if re-injecting the whole ruleset per switch is unwanted: emit just the new level's intensity-table row plus the one-line banner. That is a few hundred bytes and would fix the behavioral gap, though it leaves the stale `Default: **full**` line and the stale row from SessionStart in context alongside it, which is why full re-injection seems safer.

Either way the fix should be gated on a change, so that a session that never switches pays nothing extra.

**Before/after example**

```
Input: /caveman ultra
       (session had started at level `full`)

Got:   CAVEMAN MODE ACTIVE (ultra) — session ruleset applies.
       ...and nothing else. Context still holds the SessionStart injection:
         "Default: **full**."
         "| **full** | Drop articles, fragments OK, short synonyms. Classic caveman. ... |"
       The `ultra` row was never sent. Model continues producing full-level
       output while every turn's banner asserts ultra.

Want:  CAVEMAN MODE ACTIVE — level: ultra

       <ruleset re-injected, intensity table filtered to:>
       | **ultra** | Strip conjunctions when cause-then-effect stay unambiguous.
       One word when one word enough. State each fact once. NO prose
       abbreviations (cfg/impl/req/res/fn/auth), NO arrows (X → Y) measured zero
       token saving under tokenizer, cost decode clarity. Code symbols, function
       names, API names, error strings: never touch |
```

Reproduction, no repo needed:

1. Start a session with no `defaultMode` configured anywhere, so it resolves to `full`.
2. Run `/caveman ultra`.
3. Observe the banner on the next turn reads `(ultra)` while output length and style stay at `full`.

Confirmed the two halves independently:

```console
$ cat ~/.claude/.caveman-sessions/<session-id>.mode
ultra                                  # state write is correct

$ echo '{"session_id":"probe","cwd":"/some/repo"}' \
    | node .../src/hooks/caveman-activate.js | head -1
CAVEMAN MODE ACTIVE — level: ultra      # activate.js emits the right thing
```

So `writeSessionMode` works and `caveman-activate.js` works. Only the path between them is missing: nothing re-runs the injection when the level changes.

Setting `defaultMode` in the user config is a workaround, but it only takes effect at the *next* SessionStart, so the in-session switch remains broken:

```jsonc
// %APPDATA%\caveman\config.json   (Windows)
{ "defaultMode": "ultra" }
```

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

The same asymmetry appears reachable from the opencode loader, since `src/plugins/opencode/plugin.js` evaluates the same hook sources, but this was only observed and verified on Claude Code.

**Version / install method**

| | |
| --- | --- |
| caveman plugin | 2.5.0 (`/plugin`, marketplace `JuliusBrussee/caveman`) |
| `@caveman-ai/cli` | 1.3.1 (npm, global) |
| native pack | 2.2.0 (`caveman-proxy.exe native-hook claude`) |
| Claude Code | 2.1.259 |
| OS | Windows 10 Pro 19045 |

Both integrations were installed and active: the plugin's own `SessionStart`/`UserPromptSubmit` hooks from `.claude-plugin/plugin.json`, plus the native pack's hooks written into `~/.claude/settings.json`. The native-pack hooks are not implicated — the gap is entirely within the plugin's two hook scripts.

akshath-raj:
Fixes #975.

## What was wrong

`/caveman ultra` mid-session moved the label and nothing else.

`caveman-activate.js` (SessionStart) reads `SKILL.md`, filters the intensity table to the active level's row, injects it — **once**. `caveman-mode-tracker.js` (UserPromptSubmit) wrote the new mode and emitted only its per-turn reminder. So the model kept whatever level SessionStart had given it — that level's intensity row and its literal `Default: **full**` line still sitting in context — while every following banner asserted the new level.

Reproduced on `main` before the fix:

```console
$ echo '{"session_id":"probe","source":"startup",...}' | node src/hooks/caveman-activate.js
... Default: **full**.
... | **full** | Drop articles, fragments OK, short synonyms. Classic caveman. ... |

$ echo '{"session_id":"probe","prompt":"<command-name>/caveman</command-name><command-args>ultra</command-args>"}' \
    | node src/hooks/caveman-mode-tracker.js
{"hookSpecificOutput":{...,"additionalContext":"CAVEMAN MODE ACTIVE (ultra). Enforce this reply: ..."}}
```

State write correct, banner correct, the `ultra` row never sent. #951's per-level reminder narrowed this but does not close it: it is a hand-maintained paraphrase, and it carries none of ultra's actual directives (`NO prose abbreviations`, `NO arrows`, `State each fact once`).

The failure is silent and self-confirming — the banner agrees with the user, so both sides believe the switch took, and only the output disagrees.

## The fix

**Shared loader.** `caveman-activate.js` already had the mechanism. Moved to `caveman-config.js` as `loadFilteredRuleset` / `canonicalModeLabel` / `rulesetBanner`, used by both loaders, so the two can no longer disagree about what a level means.

`caveman-config.js` rather than the `caveman-ruleset.js` sibling the issue suggests: the hook file set is pinned in `checksums.sha256`, `verify_repo.py` and the install lists, and `src/plugins/opencode/plugin.js` can only evaluate files it knows by name — a second shared sibling is standing drift risk, which is what the *"Keep mode-state logic in `caveman-config.js`"* rule in `CLAUDE.md` is about. SessionStart output is byte-identical after the move (diffed).

**Re-injection, gated on an actual change.** The tracker compares the new level against the mode read *before* the write:

- Compared through `canonicalModeLabel` — both spellings of wenyan-full reach storage (`parseModeChange` writes the `wenyan` alias, `getDefaultMode` accepts either), so a raw compare would read that no-op as a switch.
- A `null` previous mode counts as a change: caveman was off, so the model holds no ruleset at all — the strongest case for sending one.
- Independent modes (commit/review/compress) keep their own skills and get no prose ruleset.
- `reinforce` is the gate as well as the reminder, so the #634 repo opt-out suppresses both.

A session that never switches pays nothing extra — one reminder line, one state read, exactly as before:

```
/caveman ultra   → 5490 bytes (banner + ultra row + rules, full row absent)
next turn        →  278 bytes (unchanged)
```

**Degrade paths.** A `caveman-config.js` from before this change passes the tracker's shape check, so the new exports are resolved individually against stand-ins that reproduce the previous behavior — same pattern the per-session helpers already use. An unreadable `SKILL.md` degrades the same way (standalone install, no skills dir). Both are asserted.

## Second commit: `commands/caveman.toml`

The command prompt paraphrased the `full` row regardless of its argument, so `/caveman ultra` handed the model a prompt describing full *alongside* the ultra ruleset — conflicting instructions that would partly undo the fix. It also contradicted `lite`, which keeps articles and full sentences. Now states only what holds at every level and defers the rest to the injected rules.

## Tests

New `tests/test_mode_tracker_ruleset.js` (10 cases, picked up by the existing `for f in tests/test_*.js` CI step): switch injects the new row / drops the stale one, activation from off injects, same-level re-issue does **not**, the wenyan alias is not a switch, an ordinary turn still costs one line, independent modes get no prose ruleset, and the version-skew degrade — with a control case so that assertion cannot pass for the wrong reason.

Full suite green locally:

| | |
| --- | --- |
| `tests/verify_repo.py` | all checks passed |
| `tests/test_*.js` | 12 files, all passed |
| `python -m unittest discover -s tests` | 130 passed |
| `node --test tests/installer/*.test.mjs tests/hooks/*.test.mjs` | 245 passed |

`src/hooks/checksums.sha256` regenerated (same file set).

Note: the re-injected ruleset still carries SKILL.md's own `Default: **full**` line, since that describes the configured default rather than the active level and is what SessionStart already sends. Left as-is — changing it means editing SKILL.md and its synced copies.


JuliusBrussee:
Adopted in #993 — both commits cherry-picked with your authorship preserved. The strongest PR in this batch.

Specifically right: putting the shared loader in `caveman-config.js` rather than a new sibling (which would have had to be added to `checksums.sha256`, `verify_repo.py` and the install lists); gating re-injection on an actual change so an unchanged session pays nothing extra; handling the `wenyan`/`wenyan-full` storage alias, which a naive compare would have read as a switch; and testing the plugin-cache-drift degradation rather than assuming the newer exports exist.

Verified the injection resolves SKILL.md at all three installed depths. Thank you — nothing needed changing.

---

JuliusBrussee:
Landed in #993 with your authorship preserved. Thank you.

PR #583, opened 2026-06-30
fix(windows): Bun runtime, PowerShell robustness, path/encoding + 3 more fixes

palprateek:
Adds Bun as first-class runtime (>=1.3.14) and fixes 12 Windows-specific issues:
 
### Shell / Installer
- Rename `$Args` → `$InstallerArgs` in install.ps1 to avoid PowerShell automatic-variable collision (#366, #439, #440, #518)
- Guard `$MyInvocation.MyCommand.Path` against null in pipe mode so `irm ... | iex` doesn't crash (#546, #565)
- Add Bun runtime support — works standalone or alongside Node (#570)
- Fix Node at `C:\Program Files\nodejs\node.exe` not being quoted in spawn (#461)
 
### MSYS / Git Bash
- Convert MSYS paths (`/c/Users/...`) to Windows paths (`c:/Users/...`) in Claude Code hook commands so Node resolves them on Git Bash (#199)
 
### caveman-compress
- Pin all `read_text`/`write_text` calls to `encoding="utf-8"` to prevent charmap codec crash on Windows cp1252 (#533)
- Reconfigure stdout/stderr to UTF-8 in benchmark.py for ✅/❌ glyphs
 
### Cavecrew agents
- Remove `tools:` and `model:` from agent YAML frontmatter so Gemini CLI can load them; tool restrictions already expressed in prose (#473)
 
### caveman-shrink
- Loop sentinel restoration until stable so nested protected segments (enum values like `(STARTER/BUSINESS)`) aren't left as numeric placeholders (#444)
 
### CI
- Add Node.js test workflow for installer tests

PR #982, opened 2026-09-04
fix(hooks): convert MSYS-style CLAUDE_PLUGIN_ROOT to a Windows drive path

AmirF194:
On Windows with Git Bash as the shell, Claude Code sets `CLAUDE_PLUGIN_ROOT`
as an MSYS-style path (`/c/Users/...`). Both hook commands in
`.claude-plugin/plugin.json` pass that string straight to `node.exe`, which
treats a leading `/c/` as a relative path on the current drive rather than a
drive root, so module resolution fails with `MODULE_NOT_FOUND` and
`SessionStart`/`UserPromptSubmit` silently no-op with no user-visible error.

The fix converts the path to a Windows drive form (`c:/Users/...`) before
invoking `node`, using the same sed transform from the issue's own
suggested workaround. It only matches a single-letter drive segment
(`^/[a-zA-Z]/`), so a POSIX-style root on macOS, Linux, or native Windows
shells passes through unchanged.

This is the same transform as the MSYS/Git Bash portion of #583, isolated
into a small, focused change now that #583 no longer applies cleanly.

Verified:
- New regression test (`tests/hooks/plugin-json-windows-hook-path.test.mjs`)
  reads the actual command strings from `plugin.json` and runs them under
  `bash -c` with a stub `node` on PATH, asserting the MSYS root gets
  converted and an ordinary POSIX root passes through untouched. Fails on
  main (2/4), passes on this branch (4/4).
- `npm test`, the standalone `tests/test_*.js` suite, `python -m unittest
  discover -s tests`, and `python tests/verify_repo.py` all pass in a clean
  Docker container against this branch.
- I have not run this on an actual Windows host with Git Bash and node.exe.
  I did replicate the CI `windows-powershell` job's own bash-parse check
  (`bash -n -c $h.command`) locally, and it passes.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -11,7 +11,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-activate.js\"",
+            "command": "HOOK_ROOT=$(sed 's|^/\\([a-zA-Z]\\)/|\\1:/|' <<< \"${CLAUDE_PLUGIN_ROOT}\"); node \"$HOOK_ROOT/src/hooks/caveman-activate.js\"",
             "timeout": 30,
             "statusMessage": "Loading caveman mode..."
           }
@@ -23,7 +23,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node \"${CLAUDE_PLUGIN_ROOT}/src/hooks/caveman-mode-tracker.js\"",
+            "command": "HOOK_ROOT=$(sed 's|^/\\([a-zA-Z]\\)/|\\1:/|' <<< \"${CLAUDE_PLUGIN_ROOT}\"); node \"$HOOK_ROOT/src/hooks/caveman-mode-tracker.js\"",
             "timeout": 30,
             "statusMessage": "Tracking caveman mode..."
           }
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Adopted in #993, cherry-picked with your authorship, with one correction. The MSYS diagnosis was right and is the part that mattered.

The correction: the conversion used a `<<<` herestring, which is bash-only. On Debian and Ubuntu `/bin/sh` is dash, where it is a syntax error — `dash: 1: Syntax error: redirection unexpected` — so on any host running the hook through `sh` this would have taken down SessionStart and UserPromptSubmit together, for every user on that platform, to fix a path format only Windows produces. It is now `printf %s "$VAR" | sed`, which is POSIX and does the same conversion.

The harness is why this was easy to miss: it ran the command under `bash -c` explicitly, the one shell where the bug does not reproduce. Every case now runs under `sh` and `bash`, plus a path-with-spaces case.

---

AmirF194:
Good catch, that's a real bug I introduced: the herestring is bash-only and I only tested it under bash, so it would've broken SessionStart and UserPromptSubmit on any dash-based sh. printf piped to sed is the right fix.

AmirF194:
Closing this one since #993 already carries the fix (cherry-picked with the herestring correction). Thanks for folding it in.

PR #984, opened 2026-09-04
fix(opencode): inject the active level's filtered SKILL.md rules, not just the banner

### LINKED ISSUE #983: opencode plugin never sends level-specific ruleset content: /caveman ultra|lite|wenyan-* changes only the banner label
**What happen**

`src/plugins/opencode/plugin.js`'s `experimental.chat.system.transform` hook only ever pushes the generic label below into the system prompt:

```
CAVEMAN MODE ACTIVE (<mode>) — session ruleset applies.
```

It never reads `skills/caveman/SKILL.md`, so every intensity level (lite/full/ultra/wenyan-*) runs on identical, empty instructions. The SessionStart hook (`src/hooks/caveman-activate.js:325-404`) does the real work there: it reads SKILL.md, strips frontmatter, and keeps only the active level's intensity-table row and examples. `plugin.js` has no equivalent code path anywhere in its 215 lines, so an opencode session gets zero differentiated instructions even on its first turn, before any level switch.

Verified live on HEAD (367fdb7) by driving the plugin's real `chat.message` and `experimental.chat.system.transform` hooks against an `XDG_CONFIG_HOME` sandbox: `/caveman ultra` and `/caveman lite` produce system-prompt content that differs only in the banner's mode name, with no rules behind either.

This is the opencode side of the asymmetry #975 already reported for the other integration's mid-session switch, and it is worse here: opencode never had the rules in the first place, so there is no working baseline to drift from.

**Expected**

`/caveman ultra` and `/caveman lite` should inject that level's actual filtered SKILL.md rules, the way SessionStart already does for that other integration, not only a differently-labeled banner.

**Before/after example**
```
Input: /caveman ultra
Got:   CAVEMAN MODE ACTIVE (ultra) — session ruleset applies.
Want:  CAVEMAN MODE ACTIVE (ultra) — session ruleset applies.

       [ultra's filtered SKILL.md row, e.g. "Strip conjunctions when
       cause-then-effect stay unambiguous. One word when one word enough."]
```

**Platform**
- [ ] Claude Code
- [x] Other: opencode

main @ 367fdb7.

AmirF194:
Fixes #983.

`experimental.chat.system.transform` in the opencode plugin only ever pushed the generic label `CAVEMAN MODE ACTIVE (<mode>)`. It never read `skills/caveman/SKILL.md`, so every intensity level ran on identical, empty instructions. The SessionStart hook already does the real work for another integration (`src/hooks/caveman-activate.js:325-404`): read SKILL.md, strip frontmatter, keep only the active level's intensity-table row and examples.

This mirrors that filter locally in `plugin.js` (not bridged through `caveman-config.js`, since `loadConfig()`/`loadParse()` above already exist purely because the compiled Bun runtime can't `require()`/`import()` a CJS sibling, and a third bridge isn't worth it for a ~15-line filter) and appends the result after the banner. The stale-block replace in `experimental.chat.system.transform` now matches to end of string instead of just the banner line, so a mid-session level switch swaps in the new level's rules rather than stacking behind the old ones.

Also derived the stale-block regex from `reinforcementBanner()` itself (splitting on a sentinel and escaping) rather than duplicating the banner text as a second literal, so the two can't drift.

**Verification**
- New regression test (`tests/installer/opencode.test.mjs`) confirmed failing against unpatched `main` (ultra's row absent, only the generic banner present) and passing on this branch, run in Docker on Node 18/20/22 (the CI matrix).
- Full `npm test` + `node --test tests/*.js` + `python -m unittest discover -s tests`, in Docker: same 16 pre-existing failures on `main` and on this branch (unrelated `profile scope` tests), no new failures.
- `competing-prs.sh` re-run at submission against `src/plugins/opencode/plugin.js`: one candidate, #740 (adds `recordModeChange` calls to `applyModeChange`/`handleSessionCreated`), confirmed by patch to not touch `reinforcementBanner`, `loadFilteredRuleset`, or the `system.transform` handler.
- I have not run this against opencode's own runtime, only against the plugin's exported hooks driven directly, the same way the existing test in this file already does.

### DIFF
--- a/src/plugins/opencode/plugin.js
+++ b/src/plugins/opencode/plugin.js
@@ -115,8 +115,75 @@ function removeFlag() {
   }
 }
 
-function reinforcementLine(mode) {
+function reinforcementBanner(mode) {
   return 'CAVEMAN MODE ACTIVE (' + mode + ') — session ruleset applies.';
+}
+
+function escapeRegExp(str) {
+  return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+// Derived from reinforcementBanner() itself (split on a sentinel) rather than
+// re-spelling the banner text as a second regex literal: one source of truth,
+// and it stays in sync if the wording above ever changes.
+const [bannerPrefix, bannerSuffix] = reinforcementBanner('\0').split('\0');
+const staleBlock = new RegExp(
+  escapeRegExp(bannerPrefix) + '[a-z-]+' + escapeRegExp(bannerSuffix) + '[\\s\\S]*$'
+);
+
+// SKILL.md is the single source of truth for caveman behavior, filtered to
+// the active level the same way caveman-activate.js does (#792 priority-2;
+// #909 already shipped priority-1). Kept local, not bridged via loadConfig().
+function canonicalModeLabel(mode) {
+  return mode === 'wenyan' ? 'wenyan-full' : mode;
+}
+
+function loadFilteredRuleset(mode) {
+  const modeLabel = canonicalModeLabel(mode);
+  // Two layouts, no CLAUDE_PLUGIN_ROOT-style env var for opencode: installed
+  // (plugins/caveman/plugin.js, skills at ../../skills/caveman/) and dev tree
+  // (src/plugins/opencode/plugin.js, skills three levels up at repo root).
+  const candidates = [
+    join(here, '..', '..', 'skills', 'caveman', 'SKILL.md'),
+    join(here, '..', '..', '..', 'skills', 'caveman', 'SKILL.md'),
+  ];
+  let skillContent = '';
+  for (const candidate of candidates) {
+    try {
+      skillContent = readFileSync(candidate, 'utf8');
+      break;
+    } catch (e) { /* try next candidate */ }
+  }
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Adopted in #993, cherry-picked with your authorship, then one follow-up change.

The follow-up: `plugin.js` now calls the shared `loadFilteredRuleset` from `caveman-config.js` instead of keeping its own `canonicalModeLabel` and intensity-table parser. `loadConfig()` already evaluates that file, so the bridge costs no new mechanism, and CLAUDE.md names opencode specifically as the reason not to keep a second copy of shared logic. It is resolved per call so a stale installed `caveman-config.cjs` degrades to the banner rather than throwing inside a system-prompt hook.

Your `staleBlock` derivation from the banner itself was a nice touch and is kept as-is. Verified on both layouts.

---

AmirF194:
Agreed, loadFilteredRuleset in caveman-config.js is exactly where that belonged. I didn't realize opencode already had a bridge point for it. Thanks for the follow-up.

AmirF194:
Closing this since #993 already carries the fix (with the loadFilteredRuleset follow-up). Thanks for folding it in.

PR #985, opened 2026-09-04
fix(mcp-servers): forward SIGTERM/SIGINT from caveman-shrink to its upstream child

### LINKED ISSUE #742: caveman-shrink never forwards SIGTERM/SIGINT to its spawned upstream server, orphaning it

kavish-19:
## Summary

Fixes #742. `caveman-shrink` (the MCP proxy that wraps an upstream MCP server) only reacted to the spawned child's own `error`/`close` events — it never listened for signals sent to the wrapper itself. If something tears the wrapper down directly with SIGTERM/SIGINT (an MCP host ending a session, disabling a tool, reloading config) instead of killing a whole process group, the upstream child was left running, reparented to PID 1.

- Add `process.on('SIGTERM', ...)` / `process.on('SIGINT', ...)` handlers that forward the received signal to `upstream` via `upstream.kill(signal)`.
- The existing `upstream.on('close', ...)` handler already sets the correct exit code (`128 + signal number`) once the child actually stops, so no change needed there — the parent's event loop drains naturally after `close` removes the stdin listeners.

## Test plan

- [x] Reproduced the issue's own repro against unpatched `main`: after `kill -TERM` on the wrapper, the fake upstream process survived with `PPID 1`.
- [x] Added two regression tests to `tests/installer/mcp-shrink.runtime.test.mjs` (SIGTERM and SIGINT) that spawn the wrapper around a long-running child, record the child's real PID, signal the wrapper, and assert the child process no longer exists. Confirmed both fail against unpatched `index.js` and pass with the fix.
- [x] Full suite: `npm test` — 247/247 passing (245 pre-existing + 2 new).

### DIFF
--- a/src/mcp-servers/caveman-shrink/index.js
+++ b/src/mcp-servers/caveman-shrink/index.js
@@ -75,6 +75,19 @@ upstream.on('close', (code, signal) => {
   }
 });
 
+// Registering a handler here suppresses Node's default terminate-on-signal
+// behavior, so we must forward the signal to the child ourselves — otherwise
+// the wrapper would catch SIGTERM/SIGINT and never pass it on, leaving the
+// upstream process running, reparented to PID 1. Node keeps the process
+// alive only as long as something needs it to, so once `close` fires above
+// and removes the stdin listeners, the event loop drains and we exit with
+// the code/signal set there.
+function forwardSignalToUpstream(signal) {
+  if (!upstream.killed) upstream.kill(signal);
+}
+process.on('SIGTERM', () => forwardSignalToUpstream('SIGTERM'));
+process.on('SIGINT', () => forwardSignalToUpstream('SIGINT'));
+
 // JSON-RPC framing over stdio: messages are separated by newlines (the
 // MCP stdio transport uses LSP-like content but most servers emit one JSON
 // object per line). We line-buffer in both directions and parse opportunistically.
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Adopted in #993, cherry-picked with your authorship. Minimal and correct, and the process-level test that watches the real upstream PID is what made it easy to trust — confirmed it fails on main for both signals.

One extension on top: SIGHUP is forwarded too, folded into the same handler. The root cause is that the wrapper registered no termination handling at all, and SIGHUP shares it exactly while being the likeliest of the three to be triggered by hand. Your test's signal loop made that a one-line addition.

@Dhevenddra's #775 covers the same issue and is credited alongside you on the commit.

---

JuliusBrussee:
Superseded, and not through any fault in the diagnosis.

The SIGTERM/SIGINT forwarding you proposed is already on `main` via #993, which also added SIGHUP. #1015 now builds on that with the SIGKILL escalation for an upstream that traps the signal and declines to exit (from #995).

So the bug you identified is fixed — this branch was simply overtaken while open. Nothing left to change here; flagging it so you are not left waiting on a review.

---

JuliusBrussee:
Superseded: forwarding landed via #993, SIGKILL escalation is in #1015. Thanks for the diagnosis.

PR #986, opened 2026-09-04
fix(caveman-compress): reject a compressed body that is not smaller than the input

### LINKED ISSUE #776: caveman-compress: a "compression" that EXPANDS the file passes validation and is committed

kavish-19:
## Summary

Fixes #776. `compress_file()` only rejected a compressed body when it was *exactly identical* to the input, and `validate()` checks structural invariants only (headings, code blocks, URLs, paths, bullets, inline code) — nothing compares length. A rewrite that is structurally faithful but **longer** than the input (or the same length with different content) passed every check, got written over the original, and was reported as a successful compression, defeating the tool's entire purpose.

- Added a length gate right after the existing identical-output guard in `compress_file()`: aborts before any backup or write when `len(compressed_body.strip()) >= len(body.strip())`.
- Message matches the issue's expected format: `❌ Compression aborted: output is not smaller than input (69 >= 26 chars).`

## Test plan

- [x] Reproduced the issue's exact repro against unpatched `main`: a 26-char body regressed to 69 chars, `compress_file()` returned `True`, and the larger file was written with a backup created.
- [x] Added two regression tests to `tests/test_compress_safety.py`, mirroring `test_identical_compressed_output_does_not_touch_disk`'s style:
  - `test_expanded_compressed_output_does_not_touch_disk` — output longer than input
  - `test_same_length_compressed_output_does_not_touch_disk` — output same length, different content
  Confirmed both fail against unpatched `compress.py` and pass with the fix.
- [x] Full suite: `python -m unittest discover -s tests -v` — 132/132 passing. `npm test` (Node suite) — 245/245 passing, unaffected by this change.

### DIFF
--- a/skills/caveman-compress/scripts/compress.py
+++ b/skills/caveman-compress/scripts/compress.py
@@ -649,6 +649,22 @@ def _compress_file_locked(filepath: Path) -> bool:
         print("   already in caveman form. Original file is untouched (no backup created).")
         return False
 
+    # A rewrite that is structurally faithful but LONGER than the input passes
+    # every check below (validate() only checks structural invariants, not
+    # length) and would otherwise be written over the original and reported
+    # as a successful compression — the opposite of what this tool exists to
+    # do (issue #776). Same length is also a reject: a compression that saved
+    # nothing isn't a compression.
+    compressed_len = len(compressed_body.strip())
+    body_len = len(body.strip())
+    if compressed_len >= body_len:
+        print(
+            "❌ Compression aborted: output is not smaller than input "
+            f"({compressed_len} >= {body_len} chars)."
+        )
+        print("   Original file is untouched (no backup created).")
+        return False
+
     # Reassemble: frontmatter (verbatim) + compressed body
     compressed = frontmatter + compressed_body
 
--- a/tests/test_compress_safety.py
+++ b/tests/test_compress_safety.py
@@ -73,6 +73,29 @@ class CompressSafetyTests(unittest.TestCase):
             self.assertEqual(path.read_text(encoding="utf-8"), original)
             self.assertFalse((Path(tmp) / "task.original.md").exists())
 
+    def test_expanded_compressed_output_does_not_touch_disk(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            original = "# Heading\n\nFox jump dog.\n"
+            expanded = "# Heading\n\nThe quick brown fox jumps over the lazy dog, repeatedly.\n"
+            path = self._file_with(Path(tmp), original)
+            with mock.patch.object(compress_mod, "call_claude", return_value=expanded):
+                ok = compress_mod.compress_file(path)
+            self.assertFalse(ok)
+            self.assertEqual(path.read_text(encoding="utf-8"), original)
+            self.assertFalse((Path(tmp) / "task.original.md").exists())
+
+    def test_same_length_compressed_output_does_not_touch_disk(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            original = "# Heading\n\nFox jump dog now.\n"
+            same_length = "# Heading\n\nDog jump fox now.\n"
+            self.assertEqual(len(original.strip()), len(same_length.strip()))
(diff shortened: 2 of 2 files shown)


JuliusBrussee:
Adopted in #993, cherry-picked with your authorship. Confirmed both tests fail on main and pass with the guard.

Rejecting equal length as well as longer is the right call, and testing it separately rather than folding it into the expansion case is what made that visible.

@divya0795's #777 filed the same diagnosis first, but on a stale base whose diff against current main deletes workflow files; yours applied cleanly. Both are noted in the PR body.

---

AmirF194 (commented):
I ran this against `84ee683` (this PR's head) in a clean container.

The length gate at `skills/caveman-compress/scripts/compress.py:658-660` only runs on the first candidate, before the retry loop. A repaired candidate from the fix-attempt loop (`fixed`, assigned to `compressed` at line 731) is never re-checked, so a repair that restores structure by padding prose back out can still be written as a successful compression, longer than the input, the exact bug this PR closes on the first pass.

I mocked `call_claude` to return a short-but-structurally-invalid body first (drops the heading, fails `validate_headings`), then a fix-attempt body that restores the heading but is roughly 3x the original length:

```
$ python3 -m unittest repro_986 -v
Validation attempt 1
Heading count mismatch: 1 vs 0
Fixing with Claude...
Validation attempt 2
Validation passed
compress_file() returned: True
final file length: 99 original length: 34
final == long_fix (the longer output)? True
```

`compress_file()` returns `True` and the file on disk ends up ~3x longer, reached through `build_fix_prompt` instead of the first pass.

Moving the same `len(...) >= len(...)` check to also run on `fixed` (comparing against `body`, before line 731 replaces `compressed`) would close this path too, `continue`-ing to the next attempt on failure the same way the other guards in that loop already do.

kavish-19:
You're right, and thank you for the reproduction — I confirmed it independently before checking anything else.

Mocking `call_claude` with a short-but-invalid first candidate (drops the heading) followed by a longer repair:

```
Validation attempt 1 → Heading count mismatch: 1 vs 0
Fixing with Claude... → Validation attempt 2 → Validation passed
compress_file() returned: True
final len: 82   original len: 24
final == long_fix (the LONGER output)? True
```

The line references hold on `84ee683`: the gate is at 660, the retry loop opens at 688, and `compressed = fixed` is at 731 — so the guard runs once, before the loop, and a repaired candidate reaches `_write_target` unchecked. Exactly the bug this PR closes on the first pass, one branch over.

One wrinkle worth recording for anyone who finds this later: `fixed` is a whole file — `build_fix_prompt` is given one, and the anchor check above requires it to start with the original's first line — whereas `body` is body-only. So a literal `len(fixed) >= len(body)` over-rejects whenever frontmatter is non-empty; the comparison has to split the frontmatter off first.

**This is already fixed on `main`.** Commit 6ca113d (#993) introduced `_is_smaller_than_body()` and calls it at both sites — `compress.py:684` for the first candidate and `compress.py:759` for the repaired one, the latter after `_, fixed_body = split_frontmatter(fixed)`. Its docstring names the same gap you found ("Checking only the first candidate left the retry path able to write a longer file and report it as a successful compression — the original bug, one branch over"), and `tests/test_compress_safety.py` gained `test_expanded_retry_candidate_does_not_touch_disk` to cover it.

That commit also carries this PR's change, so both of the tests here are on `main` too. I've diffed this branch against `main` and there's nothing left in it that `main` doesn't already have — which is also why it now shows as conflicting. Closing as superseded rather than re-fixing merged code.

Good catch — the guard was genuinely incomplete as written here.

kavish-19:
Superseded by 6ca113d (#993), which carries this change plus the retry-path guard @AmirF194 identified above. Nothing here is missing from `main`.

AmirF194:
Thanks for confirming and tracing it to the same fix in #993, that matches what I found on #986's head.

Issue #187, opened 2026-04-16
mode-tracker: natural language regex false positives on activation and deactivation


shaun0927:
## Problem

The natural language activation/deactivation regexes in `hooks/caveman-mode-tracker.js` (added in #120) match too broadly, causing caveman to activate or deactivate on unrelated prompts.

### False activation

Any prompt containing "start" and "caveman" anywhere triggers activation, even when "caveman" is not the object of the action:

- `"How do I start a caveman fire?"` → activates caveman
- `"I want to start building a caveman-themed game"` → activates caveman
- `"Can you start explaining caveman lifestyle?"` → activates caveman

The root cause is `\b(activate|enable|turn on|start|talk like)\b.*\bcaveman\b` — the `.*` between "start" and "caveman" matches any intervening text.

### Negation ignored

`"don't activate caveman"` still activates. The negative guard checks for `stop|disable|turn off|deactivate` but not `don't`, `do not`, or `no`.

### False deactivation

`/\bnormal mode\b/i` matches without requiring "caveman" anywhere in the prompt:

- `"switch from normal mode to debug mode"` → deactivates caveman
- `"use normal mode for this task"` → deactivates caveman
- `"is this normal mode behavior?"` → deactivates caveman

Users discussing "normal mode" in any unrelated context lose caveman mode silently.

## Evidence

Tested the current regexes against 16 prompts — **9 failed (56% false positive rate)**. The failing cases are all prompts that a real user could plausibly type without intending to change caveman mode.

```
❌ "How do I start a caveman fire?"          → false activation
❌ "I want to start building a caveman game" → false activation
❌ "Can you start explaining caveman life?"  → false activation
❌ "don't activate caveman"                  → activation despite negation
❌ "switch from normal mode to debug mode"   → false deactivation
❌ "use normal mode for this task"           → false deactivation
❌ "is this normal mode behavior?"           → false deactivation
❌ "the normal mode of operation is..."      → false deactivation
```

## Suggested fix

1. Replace broad `start.*caveman` with exact `\bstart caveman\b` so "start a caveman fire" no longer matches
2. Require "caveman" alongside "normal mode" for deactivation: `(/\bnormal mode\b/i.test(prompt) && /\bcaveman\b/i.test(prompt))`
3. Add `don't`, `do not`, `no` to the negative guard

These three changes reduce the test failure rate from 9/16 to 0/19 (expanded test set) while preserving all intended activation phrases from the README.

## Platform
- [x] Claude Code
[...]

PR #992, opened 2026-09-05
test(hooks): add reproduction corpus and table test for mode activation (#187)

poorvith-mp:
### Summary

Adds a table-driven test corpus in `tests/fixtures/mode-activation/cases.json` and a runner in `tests/test_mode_activation_corpus.js` covering natural-language activation and deactivation behavior (#187, #672, #975).

### Why

Activation and deactivation bug reports often arrive as narrative anecdotes, making regressions hard to track without an explicit reproduction table.

This PR introduces a pure-data fixture set without modifying any existing production code or regexes:
1. **Intended triggers**: Explicit commands (`/caveman ultra`, `/caveman off`, `activate caveman`, `be brief`, `back to normal mode`).
2. **Negative prose controls**: General text that shouldn't trigger state changes (`"I was reading about caveman art"`, `"switch to a different branch in git"`, `"the ultra-wide monitor setup"`, `"switch from normal mode to debug mode"`).
3. **Documented known failures**: Exact edge cases reported in #187 (e.g. `"I want to start building a caveman game"`, `"don't activate caveman"`, `"fix(caveman): stop caveman being disabled by ordinary prose"`).

### Mechanics

- `tests/fixtures/mode-activation/cases.json` stores prompts, expected outcomes, categories, and references to open issues.
- `tests/test_mode_activation_corpus.js` exercises `parseModeChange` across all cases and validates parity against the hook (`caveman-mode-tracker.js`).
- Known-failing cases are explicitly flagged (`known_failing: true`) and assert current behavior so CI passes while maintaining an honest baseline for future regex improvements.
- Uses standard library only (`fs`, `path`, `assert`, `child_process`). Runs 28 assertions in ~3s.

### Verification

- `node tests/test_mode_activation_corpus.js` (28 passed, 0 failed)
- `node tests/test_caveman_parse.js` (49 passed, 0 failed)
- `node tests/test_mode_tracker_stdin.js` (18 passed, 0 failed)
- `python -m unittest tests/test_mode_tracker.py` (34 passed, 0 failed)

### DIFF
--- a/tests/fixtures/mode-activation/cases.json
+++ b/tests/fixtures/mode-activation/cases.json
@@ -1,0 +1,220 @@
+[
+  {
+    "id": 1,
+    "prompt": "activate caveman",
+    "expected": {
+      "action": "set",
+      "mode": "full"
+    },
+    "issue": null,
+    "reason": "Direct natural-language activation command",
+    "category": "intended_activation",
+    "known_failing": false
+  },
+  {
+    "id": 2,
+    "prompt": "turn on caveman mode",
+    "expected": {
+      "action": "set",
+      "mode": "full"
+    },
+    "issue": null,
+    "reason": "Natural-language turn-on phrase",
+    "category": "intended_activation",
+    "known_failing": false
+  },
+  {
+    "id": 3,
+    "prompt": "talk like caveman",
+    "expected": {
+      "action": "set",
+      "mode": "full"
+    },
+    "issue": null,
+    "reason": "Persona activation phrase",
+    "category": "intended_activation",
+    "known_failing": false
+  },
+  {
+    "id": 4,
+    "prompt": "stop caveman",
+    "expected": {
+      "action": "clear"
+    },
+    "issue": null,
+    "reason": "Direct natural-language deactivation command",
+    "category": "intended_deactivation",
+    "known_failing": false
+  },
+  {
+    "id": 5,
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Reviewed, not adopted this run. This adds a reproduction corpus and table test for #187 without a parser change, so it documents the false positives rather than fixing them — useful groundwork, but landing it alone would add a red suite.

Left for a run that can pair it with the regex change it is testing, so the corpus lands green in the same commit.

#993 today covered the mid-session ruleset, shrink signal forwarding, compress expansion and the MSYS hook path; #187 was not among them.

---

JuliusBrussee:
Adopted into #1005, cherry-picked so your authorship is preserved. Thanks — a table-driven corpus with `known_failing` recorded honestly is a genuinely useful regression surface, and it made the #187 fix straightforward to scope.

Two changes on top: cases 16 and 17 are now fixed, so I flipped them from `known_failing` to enforced (`expected: null`). Cases 18–20 stay as you wrote them — they are real, but the guard that catches them also swallows genuine polite deactivation requests like "can you turn off caveman?", and per the #838 note a dropped deactivation is the worse failure. That one needs a maintainer decision rather than a patch.

---

JuliusBrussee:
Landed in #1005 with your authorship preserved. Thank you.

PR #994, opened 2026-09-05
fix(stats): drop unsourced output-savings estimate, keep rule overhead

### LINKED ISSUE #991: fix(stats): retire unsourced 0.65 COMPRESSION constant to align with HONEST-NUMBERS.md
There is an internal contradiction between `src/hooks/caveman-stats.js` and `docs/HONEST-NUMBERS.md` regarding counterfactual savings.

### The contradiction

In `src/hooks/caveman-stats.js` (lines 77–81):

```javascript
// Mean per-task savings from benchmarks/results/*.json (avg_savings: 65 across
// 10 tasks, sonnet-4-20250514). Only 'full' has measured data; lite / ultra /
// wenyan modes show no estimate until benchmarked. Add an entry here when a new
// run is committed.
const COMPRESSION = { 'full': 0.65 };
```

Lines 351 (`deriveSavings`) and 528 use this constant to calculate and print `est. N saved` in user-facing stats output.

However:
1. `benchmarks/results/` currently contains only `.gitkeep`. There are no committed raw benchmark result runs in the tree.
2. `docs/HONEST-NUMBERS.md` explicitly promises:
   > "Output reduction vs default verbose replies | **Not published** | Harness exists, but repository has no committed reviewed raw result"
   
   and under *Measure it yourself*:
   > "1. `/caveman-stats` (Claude Code) reads session log and prints actual output/cache counts. **It publishes no counterfactual savings until a reviewed benchmark result is committed.**"

### Community context

Users running real-world workloads have repeatedly flagged this gap when checking their own bills:
- #727 tested 20 prompts across 300 runs on `claude-opus-4-8`: output fell 41–48%, but total cost shifted +0.06% because fresh input and cache reads dominated the invoice.
- #733 noted the divergence between the advertised 65% in terminal output and an actual measured 8.5%.

`docs/HONEST-NUMBERS.md` is transparent about the economics (input overhead vs output savings). But because `caveman-stats.js` still prints the speculative `0.65` calculation, users see a number in their terminal that the repo's own documentation disowns.

### Proposed fix

1. Remove the `COMPRESSION` object and the counterfactual `est. N saved` calculation from `src/hooks/caveman-stats.js`.
2. Keep all measured counters intact: actual output tokens, cache reads, fresh input, turn counts, and rule-overhead accounting.
3. Replace the estimate line with a pointer to `docs/HONEST-NUMBERS.md` explaining that provider-billed A/B testing outranks counterfactual estimates.
4. Update `tests/test_caveman_stats.js` to assert the measured output counters without the counterfactual estimate.

Happy to open a clean PR with this diff if you'd like it fixed.

AmirF194:
`caveman-stats.js` printed an "Est. tokens saved" figure computed from a hardcoded `COMPRESSION = { full: 0.65 }` ratio. That ratio traces to a benchmark run that was never committed: `benchmarks/results/` holds only `.gitkeep`. `docs/HONEST-NUMBERS.md` explicitly promises no counterfactual savings claim until a reviewed result lands, so the tool contradicted its own honesty doc on every invocation.

Removed `COMPRESSION` and everything downstream that derived a number from it: `deriveSavings`, `deriveNet`/`netLines`, `outputReductionPct`, the per-mode savings breakdown, the USD estimate, and the lifetime output-reduction percentage. The statusline suffix is now always written empty (so a fictional figure left over from before this fix does not linger on disk) rather than carrying a savings string.

Kept the rule-overhead estimate: `docs/HONEST-NUMBERS.md` separately sources and publishes that number (the per-turn input cost of the injected rules), and it does not depend on which benchmarked mode ran, so it now shows for any active mode rather than only `full`. Kept the `#601` per-mode token-attribution mechanism (`attributeByMode`, `readModeLog`) too: it reports real, already-happened token counts and has uses beyond the removed estimate.

Updated the docs and README passages that described the retired statusline savings example, and regenerated the `caveman-stats.js` entry in `src/hooks/checksums.sha256`.

- The updated `tests/test_caveman_stats.js` catches the bug: 8 of its cases fail against `main` as it stands today, and all 48 pass on this branch. Ran both ways in Docker under Node 18 and Node 22.
- `python tests/verify_repo.py` and `python -m unittest discover -s tests` both pass, which covers the regenerated `checksums.sha256` entry and the Python-side hook tests.
- I only ran this in Linux containers, so I have not tried it on Windows or macOS, and the Go engine suite is untouched by this change so I left it alone.

Fixes #991


JuliusBrussee:
Reviewed, no action taken in today's triage (#1005). The inconsistency you identify between the `0.65` constant and `HONEST-NUMBERS.md` looks real.

I am not taking it because this triage run operates under a standing rule not to publish or change any savings percentage or token-reduction number — removing an unsourced one is still a change to a published number, and which way to resolve it is a maintainer policy call rather than a bug fix. Flagged under "Needs maintainer decision" in #1005 so it is not lost.

---

JuliusBrussee:
Reviewed, deliberately not merged by this run.

Retiring an unsourced savings constant is a change to a published token-reduction number, and this daily triage is explicitly barred from touching benchmark, savings, or reduction figures in either direction. Even though the change *removes* a claim rather than adding one — which is the safe direction, and consistent with HONEST-NUMBERS.md — the call belongs to the maintainer.

No changes requested; the reasoning in #991 looks sound to me. Flagged under "Needs maintainer decision" in #1012 so it doesn't keep sitting.

---

AmirF194:
966a4911 already replaced this whole mechanism, removing the savings concept
entirely rather than just fixing the 0.65 constant. That supersedes this PR,
closing it.

PR #995, opened 2026-09-05
fix(shrink): escalate to SIGKILL on a stuck upstream

Dhevenddra:
Follow-up to #775, which you reviewed on #993. You kept the SIGHUP gap it found and asked for the two pieces that did not come across: the grace period for a child that ignores the signal, and moving teardown into a testable `shutdown.js`. Both are here, rebuilt on current `main` rather than rebased, so there is no months-old base and no merge commit this time.

## Problem

`forwardSignalToUpstream` on `main` sends the signal and stops there:

```js
function forwardSignalToUpstream(signal) {
  if (!upstream.killed) upstream.kill(signal);
}
```

An upstream that honours SIGTERM exits, and `close` fires. One that traps it and declines to exit keeps running, and since nothing else terminates the wrapper, both processes survive the teardown the host just asked for. That is the same orphan the forwarding was added to prevent, on the subset of servers that install their own handler.

## Fix

`shutdown.forward(signal)` arms a 2s timer when it sends the signal and sends SIGKILL if the child is still alive when it expires. The timer is cleared on `close`, so a server that exits promptly is not delayed by it.

Teardown moves into `src/mcp-servers/caveman-shrink/shutdown.js`: signal forwarding, the escalation, exit-code translation, and the stdin detach. `index.js` keeps the wiring. Timers are injected, so the escalation tests run synchronously and cannot flake on a loaded runner.

`shutdown.js` is added to `package.json` "files". The packaging test from #597 fails without it, which turned out to be a useful guard.

## Evidence

`node tests/test_mcp_shrink.js` gives 26 passed / 0 failed, up from 18 on main.

Each of the eight new tests was run against a variant with the behaviour deliberately removed, so none of them is decorative:

| Variant | Test that catches it |
|---|---|
| escalation removed | escalates to SIGKILL when the upstream ignores the signal |
| SIGKILL sent unconditionally | does not escalate when the upstream exits inside the grace period |
| liveness guard removed | a second signal is a no-op, not a duplicate kill |
| `close` leaves the timer armed | close clears a pending escalation |
| grace period set to 0 | the default grace period is a real duration |
| signal exit code returns 1 | close reports 128+signal |

End to end, wrapping a server that follows stdin EOF, before and after the change:

```
main  : exited code=0 in 27ms | description -> "This is tool that reads file"
branch: exited code=0 in 36ms | description -> "This is tool that reads file"
```

Same output, same exit code, and the grace timer stays out of the normal path.

`python tests/verify_repo.py` passes. `npm test` passes.

## One thing I left out

While building the end-to-end check I hit a separate case that is still open on `main`. It is not what you asked for, so it is not in this PR. An upstream that ignores **stdin EOF** is orphaned, because `endInput()` only ends the child's stdin and nothing follows up:

```
main  : HUNG - wrapper still alive 6s after stdin EOF (upstream orphaned)
branch: HUNG - wrapper still alive 6s after stdin EOF (upstream orphaned)
```

That matters most on Windows, where there are no real signals and the comment in `index.js` points teardown at this exact path. #775 carried a backstop for it inside `endInput()`. Happy to send it as its own PR if you want it.

## Notes

- MIT directory, so inbound = outbound.
- Signed off (DCO).
- The SIGKILL escalation cannot be exercised natively on Windows, where `kill()` terminates the target without running handlers. The logic is covered by the unit tests above; I do not have a Linux node install on this machine to demonstrate it in a real process.

### DIFF
--- a/src/mcp-servers/caveman-shrink/shutdown.js
+++ b/src/mcp-servers/caveman-shrink/shutdown.js
@@ -1,0 +1,67 @@
+// Teardown for the caveman-shrink proxy, kept out of index.js so it can be
+// driven by tests without spawning a real upstream process.
+//
+// Two jobs:
+//   1. Forward a termination signal to the upstream and, if the upstream
+//      ignores it, escalate to SIGKILL after a grace period. Without the
+//      escalation a server that traps SIGTERM and declines to exit keeps both
+//      itself and this wrapper alive after the host has asked them to stop.
+//   2. Translate the upstream's `close` into our own exit code, and detach the
+//      client-side stdin listeners so the event loop can drain.
+
+const { constants: osConstants } = require('os');
+
+// Long enough for a server to flush and close its own resources, short enough
+// that a host tearing down a session does not visibly hang on us.
+const DEFAULT_GRACE_MS = 2000;
+
+function createShutdown({
+  child,
+  detachInput = () => {},
+  spawnFailed = () => false,
+  graceMs = DEFAULT_GRACE_MS,
+  // Injectable so the escalation can be tested without waiting on a real clock.
+  timers = { setTimeout, clearTimeout },
+} = {}) {
+  let graceTimer = null;
+
+  function clearGrace() {
+    if (graceTimer === null) return;
+    timers.clearTimeout(graceTimer);
+    graceTimer = null;
+  }
+
+  return {
+    // Send `signal` to the upstream and arm the escalation. Safe to call more
+    // than once: `kill` marks the child, so an impatient second Ctrl-C is a
+    // no-op rather than a duplicate signal and a second timer. A child that has
+    // already exited on its own is left alone.
+    forward(signal) {
+      if (child.killed || child.exitCode !== null || child.signalCode !== null) return;
+      child.kill(signal);
+      graceTimer = timers.setTimeout(() => {
+        graceTimer = null;
+        // Still here after the grace period: the upstream is ignoring the
+        // signal, so take the one it cannot trap.
+        if (child.exitCode === null && child.signalCode === null) child.kill('SIGKILL');
+      }, graceMs);
+    },
+
+    // Returns the exit code this process should adopt. A child that died from
+    // a signal has no exit code of its own, so report it the way a shell does.
+    onClose(code, signal) {
+      clearGrace();
+      detachInput();
+      if (spawnFailed()) return 1;
+      if (signal) return 128 + (osConstants.signals[signal] || 1);
+      return code || 0;
+    },
+
+    // Exposed for tests; nothing in index.js needs to ask.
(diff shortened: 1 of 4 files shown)


JuliusBrussee:
Adopted into #1015, cherry-picked so your authorship is preserved. This is the follow-up that was asked for on #993 and it delivers both pieces.

I mutation-tested the escalation rather than taking the table on trust: deleting the `child.kill('SIGKILL')` line turns exactly one test red, so it is load-bearing. Your end-to-end check reproduces here too — the wrapper still compresses and exits 0, with the grace timer staying out of the normal path. Adding `shutdown.js` to the package `files` array unprompted was the right instinct.

I also checked the siblings: the `@caveman-ai/cli` launcher already escalates after a 10s grace and the Go proxy wraps no child, so this wrapper was the one spawner missing it. Worth recording why yours is correctly **not** `unref`'d where the launcher's is — an otherwise-idle wrapper would exit before escalating and leave the orphan.

On the stdin-EOF orphan you left out: please do send it as its own PR. Correct call to keep it separate.

---

src/mcp-servers/caveman-shrink/shutdown.js
```diff
@@ -0,0 +1,67 @@
...
+
+  return {
+    // Send `signal` to the upstream and arm the escalation. Safe to call more
+    // than once: `kill` marks the child, so an impatient second Ctrl-C is a
+    // no-op rather than a duplicate signal and a second timer. A child that has
+    // already exited on its own is left alone.
+    forward(signal) {
+      if (child.killed || child.exitCode !== null || child.signalCode !== null) return;
+      child.kill(signal);
+      graceTimer = timers.setTimeout(() => {
```
AmirF194, line 42: I hit this too and reproduced both sides in a clean node:20-slim container.

Unpatched `main` (15581d1400) only forwards the signal, so an upstream that traps SIGTERM never exits and the wrapper hangs forever:

```
$ docker run --rm --memory=256m -v <clone>:/repo:ro node:20-slim bash -c '
  node src/mcp-servers/caveman-shrink/index.js node -e "process.on(\"SIGTERM\",()=>{});process.stdin.resume();setInterval(()=>{},1000000)" &
  WP=$!; sleep 1; kill -TERM $WP
  for i in 1 2 3 4 5; do sleep 1; kill -0 $WP 2>/dev/null || { echo "exited after ${i}s"; exit; }; done
  echo "still alive after 5s"'
still alive after 5s
```

This branch (104019c5a3) escalates to SIGKILL after the grace period in `shutdown.js:39-48`, and the same repro against it exits cleanly:

```
exited after 3s
```

I only ran this on Linux; the code comments say Windows delivers no real signals and tears the child down directly, so I did not verify that path.

JuliusBrussee:

PR #998, opened 2026-09-05
fix(compress): compare code blocks by document position, not extraction type

### LINKED ISSUE #997: caveman-compress: validate_code_blocks cannot detect a fenced/indented block swapping relative position
**What happen**
`extract_code_blocks()` in `skills/caveman-compress/scripts/validate.py` (line 109) collects every fenced code block first, then appends every indented code block after them, regardless of where each block actually sits in the document: `return blocks + extract_indented_code_blocks(text)`.

`validate_code_blocks()` (line 284) compares this list between the original and the compressed file with a plain `c1 != c2`. Because the list is grouped by block TYPE rather than ordered by document position, a compression pass that swaps a fenced block and an indented block's relative position, leaving their content unchanged, produces an identical list on both sides. Validation reports the file unchanged even though the two documents put the blocks in a different order.

**Expected**
A validator that promises code blocks are "preserved exactly" should catch a change in block order, not only a change in block content.

**Before/after example**
Two documents, same fenced Python block and the same indented shell command, only their relative order in the document differs.

~~~
Original:            Intro. <fenced python block> Middle. <indented shell command> End.
"Compressed":         Intro. <indented shell command> Middle. <fenced python block> End.
~~~

Got:   validation passes (`is_valid=True`, no errors)
Want:  validation fails (the two blocks changed relative order)

Reproduced directly against `extract_code_blocks`/`validate_code_blocks` on current `main` (`5184b3d`), in a clean `python:3.11-slim` container:

~~~
extract_code_blocks(orig) == ["<fenced block>", "<indented block>"]
extract_code_blocks(comp) == ["<fenced block>", "<indented block>"]
c1 == c2 (extractor says unchanged): True
is_valid: True errors: []
~~~

**Platform**
- [x] Other: this is a validator-logic bug in the Python script itself, not platform-specific.

**Version / install method**
`main` @ `5184b3d`

AmirF194:
`extract_code_blocks()` (`skills/caveman-compress/scripts/validate.py:109`) concatenated all fenced blocks in document order, then appended all indented blocks in document order: `blocks + extract_indented_code_blocks(text)`. That preserves within-type order but throws away cross-type order. `validate_code_blocks()` diffs this list with a plain `c1 != c2`, so a compression pass that swaps a fenced block and an indented block's relative position, content unchanged, produced identical lists on both sides and reported the file unchanged.

Fix: `extract_code_blocks()` and `extract_indented_code_blocks()` now track each block's starting line, and the final list is merged by that position instead of by extraction type. Both extractors still decide block boundaries exactly as before; only the merge order changed.

Fixes #997

**Verification**
- New regression test (`TestCrossTypeBlockOrderIsValidated`): fails on `main`, passes on this branch, run in a clean `python:3.11-slim` container.
- Full `tests/test_validate_inline.py` (23 tests) and `python -m unittest discover -s tests` (136 tests) both pass in `python:3.11-slim` / `node:20-slim`, matching the repo's `python` CI job.
- One existing test's expected block order (`test_lone_indented_marker_does_not_capture_the_real_block`) encoded the old type-grouped order; updated to the correct document-position order.
- Did not check the JS/Node compression paths in `src/` and `mcp-servers/`; this fix is scoped to the Python validator only.

### DIFF
--- a/skills/caveman-compress/scripts/validate.py
+++ b/skills/caveman-compress/scripts/validate.py
@@ -72,6 +72,13 @@ def extract_code_blocks(text):
     fence must use same char and be at least as long as opening). Supports
     nested fences (e.g. an outer 4-backtick block wrapping inner 3-backtick
     content).
+
+    Merges fenced and indented blocks by their DOCUMENT POSITION, not by
+    type. validate_code_blocks compares this list positionally (c1 != c2),
+    so a type-grouped return (all fenced, then all indented) makes two
+    blocks that swapped RELATIVE order in the document produce the same
+    list on both sides: the comparison passes while the invariant it
+    exists to check (block order preserved) has actually been violated.
     """
     blocks = []
     lines = text.split("\n")
@@ -82,6 +89,7 @@ def extract_code_blocks(text):
         if not m:
             i += 1
             continue
+        start = i
         fence_char = m.group(2)[0]
         fence_len = len(m.group(2))
         open_line = lines[i]
@@ -103,10 +111,11 @@ def extract_code_blocks(text):
             block_lines.append(lines[i])
             i += 1
         if closed:
-            blocks.append("\n".join(block_lines))
+            blocks.append((start, "\n".join(block_lines)))
         # Unclosed fences are silently skipped — they indicate malformed markdown
         # and including them would cause false-positive validation failures.
-    return blocks + extract_indented_code_blocks(text)
+    ordered = sorted(blocks + extract_indented_code_blocks(text), key=lambda pair: pair[0])
+    return [block_text for _, block_text in ordered]
 
 
 def extract_indented_code_blocks(text):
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Adopted unchanged into #1005, cherry-picked so your authorship is preserved.

Keying the merge on start line is the right fix rather than reordering at the comparison site. I verified the two things that could have bitten it: the sort is stable, and fenced and indented starts cannot collide anyway because the indented scanner skips lines already inside a fence. `extract_indented_code_blocks` has no caller outside `extract_code_blocks`, so the return-type change is contained.

---

JuliusBrussee:
Superseded — this landed in #1005 (merged 2026-09-06). Verified against `main`: `validate.py` now merges fenced and indented blocks and sorts them by document offset before comparing —

```python
ordered = sorted(blocks + extract_indented_code_blocks(text), key=lambda pair: pair[0])
```

— which is the document-position comparison this PR proposed, with the reasoning preserved in the docstring above it.

---

AmirF194:
Confirmed on main, validate.py sorts fenced and indented blocks by document offset now. Closing this one, #1005 already carries it.

PR #1000, opened 2026-09-05
fix(shrink): require no space before "(" in function-call protection

### LINKED ISSUE #999: caveman-shrink: PROTECTED_PATTERNS function-call regex swallows ordinary parentheticals
**What happen**
`PROTECTED_PATTERNS` in `src/mcp-servers/caveman-shrink/compress.js` (line 61) includes `/[A-Za-z_][A-Za-z0-9_]*\s*\([^)]*\)/g` to shield real function-call syntax from prose compression. The `\s*` also matches an ordinary word followed by a space and a parenthetical, e.g. `useful (please read carefully)`, not only actual calls like `compress(text)`. The whole span gets sentinel-protected before `compressProse` runs, so filler/hedge/pleasantry stripping never reaches an English parenthetical.

**Expected**
Only real function-call syntax (no space before the paren) should be protected; a word followed by a space and a parenthetical aside should still compress normally.

**Before/after example**
```
Input: "This tool is useful (please read carefully) before continuing."
Got:   "This tool is useful (please read carefully) before continuing."
Want:  "This tool is useful (read carefully) before continuing."
```

Reproduced directly against `compress()` on current `main` (`5184b3d`), in a clean `node:20-slim` container.

**Platform**
- [x] Other: language-agnostic bug in the Node compressor itself, not platform-specific.

**Version / install method**
`main` @ `5184b3d`

AmirF194:
Fixes #999.

`PROTECTED_PATTERNS`' function-call entry had `\s*` between the identifier and the paren, so it matched "useful (please read carefully)" the same as "compress(text)". The whole span got sentinel-protected before `compressProse` ran, so a pleasantry or filler sitting inside an ordinary parenthetical never got stripped.

Dropped the `\s*`: real function-call syntax never has a space before `(`, so this stops matching English parentheticals without touching actual calls.

Added a test pinning both a pleasantry and a filler word getting stripped out of a parenthetical, plus that `compress(text, opts)` still protects; the new test breaks without the fix. Ran the full `test_mcp_shrink.js` suite on node 18/20/22 (the CI matrix), all green.


JuliusBrussee:
Adopted unchanged into #1005, cherry-picked so your authorship is preserved. Correct root cause and a minimal diff.

I checked the one interaction the PR does not mention: `plan type (STARTER/BUSINESS)` (#444) is the same "word, space, parenthetical" shape you are unprotecting, but the enum survives because the path rule sentinels `STARTER/BUSINESS` before the function-call rule runs. The #444 nesting test passes. Also confirmed this regex has no sibling copy anywhere else in the repo, so the one edit covers every host.

---

JuliusBrussee:
Superseded — this landed in #1005 (merged 2026-09-06). Verified against `main`: `PROTECTED_PATTERNS` in `src/mcp-servers/caveman-shrink/compress.js` now reads

```js
/[A-Za-z_][A-Za-z0-9_]*\([^)]*\)/g,   // function calls (no space before `(`,
                                       // else "word (aside)" reads as a call)
```

which is your change, comment and all.

---

AmirF194:
Checked compress.js on main, the function-call pattern and comment are exactly what this PR proposed. Closing in favor of #1005.

PR #1003, opened 2026-09-05
fix(redact): recognize ghs_ GitHub App installation tokens

### LINKED ISSUE #1002: Redaction never matches a GitHub App installation token (ghs_), and the new stateless JWT format widens the gap
## What

None of the redaction layers carry a `ghs_` pattern, so a GitHub App installation token passes through unredacted. GitHub is currently migrating those tokens to a much longer, more log-prone format, which turns a latent gap into a practical one.

## Where

Four independent sites, all missing the prefix entirely:

- `shared/platform/redact/redact.go` — `credentialPrefixRe` lists `xox`, `sk_live_`, `rk_live_`, `ghp_`, `github_pat_`, `AIza` and `sk-proj-`, but no `ghs_`.
- `shared/platform/redact/payload.go` — the `credential-prefix-token` literal-needle list has the same omission. This one gates whether the regex pass runs at all, so it has to be fixed alongside the regex or the regex is never reached.
- `proxy/internal/repointel/index.go` — the credential-prefix list used to reject secret-shaped values has `ghp-`, `ghp_`, `github_pat-`, `github_pat_`, but no `ghs_`.
- `packages/subagent-tax/lib/sink.mjs` — the key pattern covers `ghp_` and `gho_`, not `ghs_`.

## Why it matters more now

Per GitHub's changelog of 2026-05-15 (https://github.blog/changelog/2026-05-15-github-app-installation-tokens-per-request-override-header/), server-to-server installation tokens — which includes the Actions `GITHUB_TOKEN` — are moving from a short opaque string to a stateless JWT: `ghs_` followed by roughly 520 base64url characters containing two dots. A 520-character credential is far more likely to end up quoted in a log line, an error body or a CI transcript than the short opaque one was, which is exactly the material these redaction layers see.

The character class matters as much as the prefix. An alphanumeric-only class such as `ghs_[A-Za-z0-9]{36,}` does not match a stateless token at all — the JWT's header segment ends at a dot well before 36 alphanumeric characters have gone by — so a pattern written that way fails closed to "no match" rather than matching partially. GitHub's own recommended validation shape is `ghs_[A-Za-z0-9\.\-_]{36,}`.

## Suggested fix

Add a `ghs_[A-Za-z0-9._-]{36,}` branch at each of the four sites. The dash belongs last in the class so it is not read as a range, matching the convention the neighbouring `AIza[A-Za-z0-9_\-]{30,}` pattern already uses.

Over-matching is not a concern in practice. Go's RE2 has no backtracking, and in either engine the class run stops at the first character outside it, so a token followed by prose ends at the space. In `sink.mjs` the surrounding `\b` even trims a trailing sentence period for free, since `.` is a non-word character.

`ghp_`, `gho_` and `github_pat_` are a different token family and are unaffected by the format change; they need no widening.

## Not affected

`shared/platform/githubapp/` needs nothing. `InstallationToken.Token` is a plain string, minted just in time and never persisted, and no `len()` in the package constrains it, so the changelog's advice about sizing storage for 520 characters has no target there.

AmirF194:
Added the `ghs_` branch at all four sites, with the character class you specified (letters, digits, dots, hyphens, underscores, dash last in the class) so it matches the stateless JWT shape and not just the old opaque token.

`payload.go`'s needle list needed the same addition since it gates whether the regex pass even runs; without it the regex change alone would never fire.

Tests: a synthetic `ghs_`-prefixed JWT-shaped fixture at each of the four call sites, split mid-token in source so push protection doesn't flag the fixtures themselves. All four show the same failure on main and pass on this branch. Full `go test ./shared/... ./proxy/...` and the `subagent-tax` node suite still pass (50/50).

Left `ghp_`, `gho_` and `github_pat_` untouched per your note, they're a different token family.
Fixes #1002

### DIFF
--- a/packages/subagent-tax/lib/sink.mjs
+++ b/packages/subagent-tax/lib/sink.mjs
@@ -93,7 +93,7 @@ export function redactHeaders(headers) {
 // shift by a few chars; body_bytes always records the raw pre-redaction length.
 const BODY_PATTERNS = [
   { re: /[A-Za-z0-9._%+-]+@[A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)+/g, tag: "email" },
-  { re: /\b(?:sk-[A-Za-z0-9_-]{10,}|ghp_[A-Za-z0-9]{20,}|gho_[A-Za-z0-9]{20,}|AKIA[0-9A-Z]{16}|xox[baprs]-[A-Za-z0-9-]{10,}|AIza[0-9A-Za-z_-]{20,})\b/g, tag: "key" },
+  { re: /\b(?:sk-[A-Za-z0-9_-]{10,}|ghp_[A-Za-z0-9]{20,}|gho_[A-Za-z0-9]{20,}|ghs_[A-Za-z0-9._-]{36,}|AKIA[0-9A-Z]{16}|xox[baprs]-[A-Za-z0-9-]{10,}|AIza[0-9A-Za-z_-]{20,})\b/g, tag: "key" },
 ];
 
 export function redactBodyString(text) {
--- a/packages/subagent-tax/tests/sink.test.mjs
+++ b/packages/subagent-tax/tests/sink.test.mjs
@@ -21,6 +21,15 @@ test("redactBody scrubs emails and credential-shaped strings, keeps JSON valid",
   assert.ok(out.system.includes("/Users/x stays"), "non-secret content untouched");
 });
 
+test("redactBody scrubs a GitHub App installation token (stateless JWT format)", () => {
+  const token = "ghs_" + "1eyJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJnaXRodWIifQ.c2lnbmF0dXJlLXBhcnQ";
+  const body = { model: "m", system: `installation token: ${token} in transcript`, messages: [] };
+  const out = redactBody(body);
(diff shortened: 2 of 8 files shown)


JuliusBrussee:
Adopted into #1005, cherry-picked so your authorship is preserved. All four sites correct, dash last in the class, dot inside it, and splitting the fixtures mid-token to avoid push protection is a nice touch.

One addition on top: the package doc comment in `redact.go` enumerates the covered vendor prefixes and still listed the pre-`ghs_` set, so it claimed narrower coverage than the code has. Also verified there is no fifth site — `residual.go` reuses `credentialPrefixRe` rather than keeping its own list.

---

JuliusBrussee:
Superseded — the `ghs_` redaction landed in #1005 (merged 2026-09-06), which adopted this change across all four sites. Verified against `main`: `credentialPrefixRe`, the `credential-prefix-token` needle list, `safeTerm`, and `sink.mjs` all carry `ghs_` today.

Your follow-up #1009/#1010 for `gho_`/`ghu_`/`ghr_` is adopted in #1012.

---

AmirF194:
Verified redact.go on main carries the ghs_ pattern across all four sites #1005 mentions. Closing here since #1010 already covers the gho_/ghu_/ghr_ follow-up.

PR #1007, opened 2026-09-06
fix(proxy): forward OpenCode session headers on the opencode-go mount

### LINKED ISSUE #1006: proxy: opencode-go compat mount drops the x-opencode-session header that Pi sends
**What happen**

OpenCode sent this notice on 2026-09-06:

> Some of your requests to OpenCode Go are missing an x-opencode-session header. If we don't have this we cannot properly optimize our service. Starting 09/06 requests missing this header may error.

Pi sends the header. The Caveman proxy removes it.

Pi 0.85.1 adds `x-opencode-session` and `x-opencode-client: pi` to each request that has a session ID. The condition is `model.provider` equal to `opencode` or `opencode-go`, or a base URL host equal to `opencode.ai`. The check uses the provider name, not the base URL. Thus Pi sends the headers after the caveman extension points the base URL at the gateway. Source in `@earendil-works/pi-coding-agent` 0.85.1, `src/core/provider-attribution.ts`:

```ts
function getSessionHeaders(model: Model<Api>, sessionId: string | undefined): Record<string, string> | undefined {
	if (!sessionId) return undefined;
	if (
		model.provider !== "opencode" &&
		model.provider !== "opencode-go" &&
		!matchesHost(model.baseUrl, OPENCODE_HOST)
	) {
		return undefined;
	}
	return { "x-opencode-session": sessionId, "x-opencode-client": "pi" };
}
```

The proxy builds the upstream headers from a fixed allowlist. `Base.SanitizeAndMapHeaders` makes a new header map and copies only the listed names (`proxy/providers/adapter.go:433`). The list has `content-type`, `accept`, `accept-encoding`, `idempotency-key`, the OpenAI headers, the Anthropic headers, and `api-version`. It does not have `x-opencode-*`. The opencode-go compat mount overrides this function (`namedAdapter.SanitizeAndMapHeaders`, `proxy/providers/openaicompat/openaicompat.go:184`). The override returns early for each path that is not `/v1/messages` (line 189). On `/v1/messages` it moves the key into `x-api-key`. It does not restore the session headers on any path. The gateway assigns the map to the upstream request (`proxy/internal/gateway/proxy.go:348`).

The pi extension routes three OpenCode Go APIs through this mount (`packages/pi-extension/src/protocol.ts:96`): `anthropic-messages` to `/v1/messages`, `openai-completions` to `/v1/chat/completions`, and `openai-responses` to `/v1/responses`. The proxy drops the headers on all three paths.

PR #948 (merged 2026-09-02) added the built-in `/compat/opencode-go` mount and the automatic Pi route. Before #948, the extension sent OpenCode Go models to `/w/pi/openai/v1`, and the proxy sent them to `api.openai.com` (#946). PR #969 fixed the auth scheme on `/v1/messages`, and PR #972 fixed the routing rules. Thus #948 exposed the existing allowlist limit on the new route.

The local proxy database (`~/.caveman/caveman.db`, table `requests`, column `ts` in UTC) shows the traffic from Pi. All rows have `agent_slug=pi`:

| Day (UTC) | Endpoint | Status | Requests |
|---|---|---|---|
| 2026-09-01 | /compat/opencode-go/v1/responses | 200 | 7 |
| 2026-09-01 | /compat/opencode-go/v1/chat/completions | 200 | 1 |
| 2026-09-04 | /compat/opencode-go/v1/chat/completions | 200 | 23 |
| 2026-09-05 | /compat/opencode-go/v1/chat/completions | 200 | 18 |
| 2026-09-05 | /compat/opencode-go/v1/chat/completions | 502 | 2 |

The rows from 2026-09-01 are before the merge of #948. They most likely came from a local build of the #948 branch. The local reflog shows the first commit of that branch six minutes before the first row. The rows have no build hash, so this is an inference. The two 502 responses have no explanation in the proxy log. No live rejection by OpenCode is claimed from them.

The database stores no headers, and `CAVE_CAPTURE_DIR` stores bodies and metadata only. Thus no local record shows the headers that Pi sent. The evidence for the header loss is the source above.

Not affected:

- Pi in direct mode. The request goes to `opencode.ai` with the headers.
- The OpenCode CLI without `caveman enable opencode`. Its requests do not pass through the proxy.

Not verified: the OpenCode CLI with `caveman enable opencode` and an OpenCode Go model. The `/w/opencode` route uses the same allowlist. Thus it can drop the same headers. The issue text attributes `x-opencode-project` and `x-opencode-request` to the OpenCode CLI. This attribution is not checked against the OpenCode CLI source.

**Expected**

The opencode-go compat mount forwards `x-opencode-session`, `x-opencode-client`, `x-opencode-project`, and `x-opencode-request` to the upstream. The adapter copies the four names before the early return at `openaicompat.go:189`. The copy applies only to the `opencode-go` mount. The shared allowlist does not change, and no `x-opencode-*` wildcard is added.

Two tests pin the behavior:

- An adapter test in `openaicompat_test.go` with a matrix over `/v1/messages`, `/v1/chat/completions`, and `/v1/responses`. It asserts the exact values when the headers are present. It asserts that the headers stay absent when the inbound request has none. It asserts that a second named mount forwards none of the four.
- An extension of `TestStandaloneOpenCodeGoAuthEndToEnd` in `standalone_test.go`. It puts the headers in the existing `inbound` field and asserts the values at the fake upstream, next to the current auth and body assertions.

The fix lives in the Go proxy binary. Users need a new `bin-v*` release, not only a CLI bump.

Related: #946, #948, #967, #969, #970, #972.

**Before/after example**

```
Input: caveman pi --provider opencode-go --model kimi-k2.6 -p 'Reply with OK'
Got:   the upstream request has no x-opencode-session header; OpenCode may reject it from 2026-09-06
Want:  the upstream request carries x-opencode-session and x-opencode-client, the same as direct mode
```

**Platform**
- [ ] Claude Code
- [ ] Codex
- [x] Other: Pi (OpenCode Go provider through the proxy)

**Version / install method**

CLI 1.3.3 on bin-v1.1.6, `caveman enable pi`. Repository at `e0e0253` on `main`, after PR #948, #969, and #972. Pi 0.85.1.

zsltg:
## Summary

- The opencode-go compat mount now forwards `x-opencode-session`, `x-opencode-client`, `x-opencode-project`, and `x-opencode-request` to the upstream (fixes #1006). Pi 0.85.1 sends the first two on each request that has a session ID. OpenCode announced that from 2026-09-06 a request without `x-opencode-session` can get an error.
- `Base.SanitizeAndMapHeaders` builds a new header map from a fixed allowlist, thus the proxy dropped every `x-opencode-*` header. The named adapter in `proxy/providers/openaicompat/openaicompat.go` now copies the four names after the base call and before the early return for non-Messages paths. Thus the copy runs on `/v1/messages`, `/v1/chat/completions`, and `/v1/responses`.
- The copy is gated on the mount prefix `/compat/opencode-go`. The shared allowlist, the credential mapping, and the `x-api-key` remap on `/v1/messages` are unchanged. No `x-opencode-*` wildcard is added, so another named mount does not learn the session identity of the caller.
- The four header names live in one package-level slice with a comment that gives the reason.

## Proof

- New adapter test `TestNewNamed_ForwardsOpenCodeSessionHeaders` in `proxy/providers/openaicompat/openaicompat_test.go`. It covers the three paths with the headers present and absent. It asserts that the messages path still sets `x-api-key` and `anthropic-version`. It asserts that a second named mount at `/compat/other` forwards none of the four headers.
- `TestStandaloneOpenCodeGoAuthEndToEnd` in `proxy/internal/standalone/standalone_test.go` gets a `wantHeaders` field and three new cases, one per path. Each case sends `x-opencode-session` and `x-opencode-client` through `/w/pi/compat/opencode-go/...` and asserts the exact values at the fake upstream, next to the existing auth and body assertions. The four existing cases now assert that no `x-opencode-session` gets to the upstream.
- Regression proof: with the tests from this branch and `openaicompat.go` from `main`, both tests fail. The adapter test reports four empty headers on each of the three paths. The standalone test reports `upstream x-opencode-session = ""` on all three new cases.
- `go build ./...`, `gofmt -l`, and `go vet` on the changed packages are clean. `go test ./providers/... ./internal/standalone/... ./internal/gateway/...` passes.
- Local end-to-end run on 2026-09-06: a proxy built from this branch, a `compat: opencode-go` override to a loopback fake upstream, and `CAVE_SSRF_ALLOWLIST=127.0.0.1:9999`. The fake upstream received `x-opencode-session` and `x-opencode-client` on `/v1/chat/completions` and `/v1/responses`. The binary from bin-v1.1.6 sent none.
- Not run: `proxy/cmd/caveman-proxy`, `proxy/internal/nativeruntime`, and `proxy/internal/store` fail on this machine because the temp directory is group-writable. They fail the same way on a clean `main` and do not touch the changed code.

## Notes

- The fix lives in the Go proxy binary. Users need a new `bin-v*` release, not only a CLI bump.
- The OpenCode CLI with `caveman enable opencode` and an OpenCode Go model goes through the `/w/opencode` route. That route uses the same allowlist and is not changed here. The attribution of `x-opencode-project` and `x-opencode-request` to the OpenCode CLI comes from #1006 and is not checked against the OpenCode CLI source.
- No local record shows the headers that the installed Pi client sent to the proxy. The database and `CAVE_CAPTURE_DIR` store no headers. The claim about Pi rests on the quoted `getSessionHeaders` source in #1006.

### DIFF
--- a/proxy/internal/standalone/standalone_test.go
+++ b/proxy/internal/standalone/standalone_test.go
@@ -1115,16 +1115,19 @@ func TestStandaloneOpenCodeGoAuthEndToEnd(t *testing.T) {
 	const anthropicResponse = `{"id":"msg","type":"message","role":"assistant","model":"minimax-m3","content":[{"type":"text","text":"OK"}],"usage":{"input_tokens":1,"output_tokens":1}}`
 	const openAIBody = `{"model":"glm-5.2","messages":[{"role":"user","content":"Reply with OK"}]}`
 	const openAIResponse = `{"id":"chat","model":"glm-5.2","choices":[{"message":{"role":"assistant","content":"OK"}}],"usage":{"prompt_tokens":1,"completion_tokens":1}}`
+	const responsesBody = `{"model":"glm-5.2","input":"Reply with OK"}`
+	const responsesResponse = `{"id":"resp","model":"glm-5.2","output":[{"type":"message","content":[{"type":"output_text","text":"OK"}]}],"usage":{"input_tokens":1,"output_tokens":1}}`
 
 	cases := []struct {
-		name       string
-		path       string
-		body       string
-		response   string
-		inbound    map[string]string
-		wantURL    string
-		wantAPIKey string
-		wantAuth   string
+		name        string
+		path        string
+		body        string
+		response    string
+		inbound     map[string]string
+		wantURL     string
+		wantAPIKey  string
+		wantAuth    string
+		wantHeaders map[string]string
 	}{
 		{
 			name:       "messages with inbound x-api-key",
@@ -1161,6 +1164,57 @@ func TestStandaloneOpenCodeGoAuthEndToEnd(t *testing.T) {
 			wantURL:  "https://opencode.ai/zen/go/v1/chat/completions",
 			wantAuth: "Bearer sk-inbound",
 		},
+		{
+			name:     "messages forwards the OpenCode session headers",
+			path:     "/w/pi/compat/opencode-go/v1/messages",
+			body:     anthropicBody,
+			response: anthropicResponse,
+			inbound: map[string]string{
+				"x-api-key":          "sk-inbound",
+				"x-opencode-session": "ses_messages",
+				"x-opencode-client":  "pi",
+			},
+			wantURL:    "https://opencode.ai/zen/go/v1/messages",
+			wantAPIKey: "sk-inbound",
+			wantHeaders: map[string]string{
+				"x-opencode-session": "ses_messages",
+				"x-opencode-client":  "pi",
+			},
+		},
+		{
+			name:     "chat completions forwards the OpenCode session headers",
+			path:     "/w/pi/compat/opencode-go/v1/chat/completions",
+			body:     openAIBody,
+			response: openAIResponse,
+			inbound: map[string]string{
+				"authorization":      "Bearer sk-inbound",
+				"x-opencode-session": "ses_chat",
+				"x-opencode-client":  "pi",
+			},
+			wantURL:  "https://opencode.ai/zen/go/v1/chat/completions",
+			wantAuth: "Bearer sk-inbound",
+			wantHeaders: map[string]string{
+				"x-opencode-session": "ses_chat",
+				"x-opencode-client":  "pi",
+			},
+		},
+		{
+			name:     "responses forwards the OpenCode session headers",
+			path:     "/w/pi/compat/opencode-go/v1/responses",
+			body:     responsesBody,
+			response: responsesResponse,
+			inbound: map[string]string{
+				"authorization":      "Bearer sk-inbound",
+				"x-opencode-session": "ses_responses",
+				"x-opencode-client":  "pi",
+			},
+			wantURL:  "https://opencode.ai/zen/go/v1/responses",
+			wantAuth: "Bearer sk-inbound",
+			wantHeaders: map[string]string{
+				"x-opencode-session": "ses_responses",
+				"x-opencode-client":  "pi",
+			},
+		},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


JuliusBrussee:
Adopted into #1012, cherry-picked so your authorship is preserved.

Verified before adopting: your adapter test fails on unmodified source across all three paths, then passes after the change.

Two things I checked and agree with. Gating `forwardOpenCodeHeaders` on `a.prefix` rather than on `req.URL.Path` is the more robust choice — the `/w/<agent>` workspace prefix can't confuse adapter identity the way a path match could. And the `/w/opencode` route you flagged as unverified in the issue turns out not to exist as a provider mount: `opencode-go` is the only entry in `builtinCompat`, so there's no second mount dropping these headers.

Nothing to change.

---

PR #1010, opened 2026-09-07
fix(redact): recognize gho_/ghu_/ghr_ GitHub OAuth and App tokens

### LINKED ISSUE #1009: Redaction covers ghs_ but never gho_ (OAuth), ghu_ or ghr_ (App tokens)
## What

Redaction covers GitHub App installation tokens (`ghs_`, fixed in #1003/#1005) but never covered OAuth access tokens (`gho_`), App user-to-server tokens (`ghu_`) or App refresh tokens (`ghr_`). All three are current, non-migrating GitHub token formats, so this is a live gap today, not a forward-looking one.

## Where

Four independent sites, same shape as the `ghs_` gap:

- `shared/platform/redact/redact.go`, `credentialPrefixRe`: has `ghp_` and now `ghs_`, but no `gho_`, `ghu_` or `ghr_`.
- `shared/platform/redact/payload.go`, the `credential-prefix-token` needle list: same omission, and it gates whether the regex pass runs at all.
- `proxy/internal/repointel/index.go`, `safeTerm`'s prefix denylist: same three missing.
- `packages/subagent-tax/lib/sink.mjs`: already has `gho_` (confirmed by testing the live regex), still missing `ghu_` and `ghr_`.

## Character class

Unlike `ghs_`, which is migrating to a stateless JWT, `gho_`, `ghu_` and `ghr_` are plain alphanumeric tokens with no dots or dashes, the same shape `ghp_` already uses. A `gho_[A-Za-z0-9]{20,}`-style class (matching the neighbouring `ghp_[A-Za-z0-9]{20,}` convention) covers all three without the JWT-specific character class `ghs_` needed.

## Not affected

`shared/platform/redact/residual.go` reuses `credentialPrefixRe` rather than keeping its own list, so it needs no separate change.

AmirF194:
Added `gho_`, `ghu_` and `ghr_` branches at the same four sites #1003 touched for `ghs_`. Unlike `ghs_`, these three are not migrating to a JWT shape, so they use the plain alphanumeric class the neighbouring `ghp_` pattern already uses, not the wider dotted class `ghs_` needed.

`sink.mjs` already had `gho_`, it was only missing `ghu_` and `ghr_`.

Added a fixture per token per site (split mid-token in source so push protection does not flag the fixtures), following the same pattern as the existing `ghs_` tests. The new test cases fail without the change and pass with it. Full `go test ./...` passes; the `subagent-tax` node suite passes in full (51/51). Did not run the full `pnpm -r build`/`pnpm -r test` across the other workspace packages, since this change touches only `subagent-tax` and its own suite already covers it end to end.

Fixes #1009

### DIFF
--- a/packages/subagent-tax/lib/sink.mjs
+++ b/packages/subagent-tax/lib/sink.mjs
@@ -93,7 +93,7 @@ export function redactHeaders(headers) {
 // shift by a few chars; body_bytes always records the raw pre-redaction length.
 const BODY_PATTERNS = [
   { re: /[A-Za-z0-9._%+-]+@[A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)+/g, tag: "email" },
-  { re: /\b(?:sk-[A-Za-z0-9_-]{10,}|ghp_[A-Za-z0-9]{20,}|gho_[A-Za-z0-9]{20,}|ghs_[A-Za-z0-9._-]{36,}|AKIA[0-9A-Z]{16}|xox[baprs]-[A-Za-z0-9-]{10,}|AIza[0-9A-Za-z_-]{20,})\b/g, tag: "key" },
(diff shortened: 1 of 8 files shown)


JuliusBrussee:
Adopted into #1012, cherry-picked so your authorship is preserved.

Verified before adopting: applied your test files against unmodified source and watched all four sites fail for the right reason, then confirmed green after the source change. I also grepped `ghp_` repo-wide independently and found no fifth site, so the coverage is complete — and you were right that `residual.go` needs nothing, since it reuses `credentialPrefixRe`.

Nothing to change. Splitting the fixtures mid-token so push protection never sees a contiguous literal was a good call.

---

AmirF194:
gho_, ghu_ and ghr_ are all in redact.go on main now, matches this PR. Closing, #1012 has it.

PR #1013, opened 2026-09-07
fix(init): honor force when refreshing rules

ghdtjdwn:
Running `caveman-init --force` on an already initialized Cursor, Windsurf, or Cline repository reports `skipped-already-installed` and leaves stale rules in place. The sentinel check returns before the force-overwrite branch, despite the documented refresh behavior.

Honor `--force` when checking the sentinel. Regression tests exercise all three targets through the CLI, verifying that default runs preserve existing content, `--force --dry-run` reports an overwrite without writing, and `--force` restores the current rule and frontmatter.

Validation on macOS (Node v24.18.0):
- `node tests/test_caveman_init.js`: 15 passed. The three new cases failed before the fix.
- `npm test`: 279 passed.
- `node --check src/tools/caveman-init.js` and `git diff --check`: passed.

Changes are limited to the MIT initializer and its tests.

### DIFF
--- a/src/tools/caveman-init.js
+++ b/src/tools/caveman-init.js
@@ -187,7 +187,7 @@ function processAgent(agent, targetDir, ruleBody, opts) {
   // "our older ruleset" from "a user file that happens to quote the rule".
   // Refreshing on a guess would clobber their content, so skip and let the
   // existing --force flag be the deliberate refresh path.
-  if (existing.includes(SENTINEL)) {
+  if (!opts.force && existing.includes(SENTINEL)) {
     return { status: 'skipped-already-installed', label: '=' };
   }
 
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Adopted into #1015, cherry-picked so your authorship is preserved. Confirmed red first: all three of your cases fail on unmodified `caveman-init.js`, reporting `skipped-already-installed` where `1 overwritten` was expected.

Good catch on a one-character bug with real reach — once a repo was initialized, an edit to `src/rules/caveman-activate.md` could never reach `.cursor/rules/caveman.mdc`, `.windsurf/rules/caveman.md` or `.clinerules/caveman.md` by any route. It is also better supported than your description claims: `INSTALL.md` documents `--force` as "Re-run even if already installed" and the tool's own `--help` says "overwrite existing rule files", so this brings the code into line with two independent docs rather than changing a contract.

I checked the append-mode sibling at line 173 and deliberately left it alone — those files are shared with the user and marker-fenced, so forcing there could clobber content we cannot prove is ours. Leaving it out was the right call.

---

JuliusBrussee:

PR #577, opened 2026-06-26
fix(tracker): reduce per-turn injection when competing compression plugin detected

lg320531124:
## Problem

When caveman and another token-compression plugin (e.g. ponytail) are both active, they both inject `additionalContext` on `UserPromptSubmit`. The model receives two contradictory style instruction sets, causing unpredictable output.

## Fix

In `caveman-mode-tracker.js`, before emitting the per-turn `additionalContext`, check `settings.json` for known compression plugins (ponytail, grill-me, grilling). If a conflict is detected:

- **Reduce** the per-turn injection to a one-liner: `"CAVEMAN MODE ACTIVE (full) — reduced injection (another compression plugin detected). Caveman rules still apply per SessionStart injection."`
- **Skip** the full style reminder to avoid competing with the other plugin's per-turn injection
- The **SessionStart full ruleset is unchanged** — caveman rules still apply, just the per-turn reinforcement is reduced

Detection checks `settings.json` `installed_plugins` and `permissions` keys for known compression plugin IDs.

## Trade-off

This is a **conservative approach** — caveman still activates and its SessionStart ruleset is still injected, but the per-turn reinforcement is reduced when a conflict is detected. The alternative (disabling caveman entirely) would lose the SessionStart ruleset, which is too aggressive.

## Testing

With ponytail installed in settings.json:
- Before: model receives both `"CAVEMAN MODE ACTIVE (full). Drop articles/filler..."` and `"PONYTAIL MODE ACTIVE (full). The ladder enforced..."` every turn
- After: model receives `"CAVEMAN MODE ACTIVE (full) — reduced injection"` + ponytail's full injection = no contradiction

Fixes #574

PR #1014, opened 2026-09-07
fix(hooks): reduce per-turn reinforcement when a compression plugin conflicts

### LINKED ISSUE #574: [Bug] Multiple token-compression plugins conflict — caveman + ponytail/other compression skills inject contradictory style instructions
## Problem

When caveman and another token-compression skill (e.g. ponytail) are both active, they both inject `additionalContext` on the `UserPromptSubmit` hook event. The model receives **two contradictory style instruction sets** in the same turn, causing inconsistent output — sometimes terse, sometimes verbose, sometimes a confusing mix.

## Reproduction

1. Install caveman (full mode) and ponytail (full mode) — both are popular Claude Code plugins
2. Start a session — both hooks fire on `UserPromptSubmit`
3. Observe the model receives:
   - From caveman: `"CAVEMAN MODE ACTIVE (full). Drop articles/filler/pleasantries/hedging. Fragments OK. Code/commits/security: write normal."`
   - From ponytail: `"PONYTAIL MODE ACTIVE — level: full. The ladder enforced. Stdlib and native first. Shortest diff, shortest explanation."`
4. Model behavior becomes unpredictable — sometimes follows caveman rules, sometimes ponytail rules, sometimes both partially

## Root Cause

`caveman-mode-tracker.js` (line 122-132) unconditionally emits `additionalContext` when caveman is active, without checking whether another compression plugin is already injecting competing style rules. Same for `caveman-activate.js` (line 91) which emits the full ruleset on `SessionStart`.

There is no conflict detection or de-escalation mechanism.

## Related but Different

- #184 addresses the `.caveman-active` **flag file** being global (per-session file lock conflict)
- This issue addresses the **output semantics** — two compression instruction sets competing for the model's attention in the same turn

## Suggested Fix

In `caveman-activate.js` and `caveman-mode-tracker.js`, before injecting compression rules:

1. **Detect** known conflicting plugins by checking `settings.json` for registered plugins (ponytail, grill-me, etc.) or scanning `CLAUDE.md` for their characteristic markers
2. **De-escalate**: if a conflicting compression plugin is detected, either:
   - Skip caveman's injection entirely (let the other plugin handle compression), or
   - Reduce to a minimal one-liner instead of the full ruleset, or
   - Emit a warning that both are active and suggest the user pick one
3. **Document** the interaction in README — list known conflicting plugins and recommended coexistence strategy

## Environment

- caveman v18e45320 (latest)
- Claude Code with both caveman + ponytail plugins installed
- macOS, Node.js v22

AmirF194:
Re-ports the conflict-detection behavior from #577 against current HEAD. That branch predated the per-session (#599/#601) and level-aware (#975) refactors and can't be cherry-picked as-is, but the settings.json read and plugin-name detection still apply unchanged.

When settings.json's `installed_plugins` or `permissions` names a known compression plugin (ponytail, grill-me, grilling), the per-turn reminder in `caveman-mode-tracker.js` now reduces to a one-liner instead of restating the full ruleset. SessionStart's injection and the mid-session level-switch ruleset (#975) are untouched, same as #577 proposed.

Plugin list is taken from #577's body; I haven't independently verified those are the only compression plugins caveman users pair it with.

Added tests in `test_mode_tracker.py` for the reduced/full/unrelated-plugin/malformed-settings paths. The ponytail and permissions-key cases fail on main.

Fixes #574


JuliusBrussee:
Reviewed, not adopted — this needs a maintainer decision rather than changes from you.

The implementation is careful. You regenerated `checksums.sha256` (which the other PR I adopted today forgot), covered the JSONC fallback, and tested both the malformed and the unreadable-settings paths including an `EISDIR` case for the outer catch. No technical objection.

Two things put it above my line. #577 proposed this and was closed once already, and I cannot tell from here whether that was staleness or a decision. And the detection turns on three hardcoded plugin names you were straightforward about not having verified: a user pairing a fourth compression plugin gets no back-off, while anyone with "grilling" in a settings key gets a silently reduced ruleset. Whether caveman defers to its neighbours at all is a product call.

Flagged for the maintainer in #1015 with that framing. Not a rejection — thanks for re-porting it cleanly.

---

AmirF194:
#577 was staleness, not a decision: closing comment there says "the conflict-detection goal remains useful, but this PR became a whole-repo stale merge... please re-port the narrow #574 behavior against the current parser/runtime." That's exactly what this PR does.

The three-plugin hardcode is a real product call, agreed that's yours to make.

JuliusBrussee:
**Needs changes**, and the re-port against current HEAD was the right call — the #599/#975 refactors did make #577 uncherry-pickable.

Two things before this can land:

1. **The `permissions` half likely never fires.** Your fixture writes `permissions={"grill-me@grill-me": True}`, i.e. plugin names as top-level keys. Claude Code's `permissions` object is keyed `allow`/`deny`/`ask`, so `Object.keys(settings.permissions)` yields those three and never a plugin name — the test passes because the fixture uses a shape the host doesn't emit. I couldn't confirm the real key from this container (no populated `settings.json` here), so please check against your own before changing anything; `installed_plugins` has no corroboration anywhere in this repo either.

2. **The JSONC fallback and the parsed path disagree.** On a valid-JSON `settings.json`, `permissions: {"allow": ["Bash(ponytail:*)"]}` detects nothing; add one comment to the same file and the raw-substring fallback matches it. Detection shouldn't depend on whether the user comments their settings. The `JSON.parse`-with-fallback pattern itself is fine — `caveman-activate.js:395` does exactly that, for the same reason — but it's testing a boolean there, not substring-matching names across the whole file.

The behavior question underneath (#574) is a maintainer call and is in #1015 for him: backing off to a one-liner assumes the other plugin's rules are the better ones to keep, which isn't obviously true.

---

AmirF194:
Confirmed the real shape from this repo's own agents/agents.json and agents/profiles/qwen.json: permissions is allow/deny/ask arrays of rule strings, never plugin names as keys. Fixed both to read those arrays, so the parsed path now catches the same rule strings the JSONC fallback matches on. Added a test that runs one input through both paths.

#574 behavior question stays with you in #1015.

AmirF194:
Both fixes from your last review are in (7cec223): permissions now reads the allow/deny/ask arrays instead of object keys, and the JSONC fallback agrees with the parsed path on the same rule string. Anything else you want changed before this lands?

JuliusBrussee:
**Reviewed, not adopted — this needs a maintainer decision rather than a triage call.** The re-port against current HEAD is clean and the tests are the right ones, so this is not a quality objection.

The blocker is the shape: it hardcodes three third-party plugin names into a shipped hook, and you say plainly you couldn't verify that list is complete. An allowlist that quietly reduces reinforcement for the plugins on it and not for the ones off it is a policy choice about how caveman behaves alongside other plugins, and it's one the maintainer should make — including whether detection should key on something more durable than a name list that goes stale the moment a fourth compression plugin ships. It also lands in `src/hooks/`, so it carries a `checksums.sha256` regeneration.

Leaving it open and flagged on #1103 under "needs maintainer decision". Thanks for carrying #577's work forward — the settings.json read and the four test paths are reusable whatever shape the detection ends up taking.

---

PR #1018, opened 2026-09-08
fix(browse): close Chrome and remove temp profile dir on SIGTERM/SIGINT/SIGHUP

### LINKED ISSUE #1016: browse: caveman-browse installs no signal handler — SIGTERM orphans its Chrome and leaks the temp profile dir (same defect as #742)
**What happen**

`caveman-browse` releases its Chrome process and its temp profile directory only from the deferred `Close` calls in `main()`. `main()` registers no signal handler, so the SIGTERM an MCP host sends at session teardown kills the process outright: Chrome survives, reparented to PID 1, and the `caveman-browse-*` directory under `$TMPDIR` is never removed.

`browse/cmd/caveman-browse/main.go`:

```go
driver, err := browse.NewCDPDriver(context.Background(), browse.CDPOptions{...})
...
defer driver.Close()   // kills Chrome and does os.RemoveAll(tmpProfile)
...
srv := mcp.NewServer("caveman-browse", browse.BrowserTools(session), logger)
if err := srv.Serve(os.Stdin, os.Stdout); err != nil { ... }
```

`signal.Notify` appears nowhere in `browse/`. The stdin-EOF path is correct: `mcp.Server.Serve` returns `nil` on `io.EOF`, the defers run, and cleanup happens. Only the signal path leaks.

This is the same defect #742 fixed for `caveman-shrink`, in the browse driver.

**Expected**

SIGTERM/SIGINT/SIGHUP should run the same cleanup as stdin EOF: close Chrome, remove the temp profile directory.

**Repro**

```sh
BIN=~/.caveman/bin/caveman-browse

# leaks: killed by signal
CAVEMAN_BROWSE_EPHEMERAL=1 "$BIN" < <(sleep 60) >/dev/null 2>&1 &
SRV=$!
# wait for Chrome to appear, then:
CH=$(pgrep -P $SRV -f "Google Chrome" | head -1)
kill -TERM $SRV
ps -o ppid= -p $CH               # 1  -> orphaned, still running
ls -d "$TMPDIR"caveman-browse-*  # its profile directory is still there

# clean, for contrast: stdin EOF
CAVEMAN_BROWSE_EPHEMERAL=1 "$BIN" < /dev/null >/dev/null 2>&1 &
# server exits 0, Chrome exits, its profile directory is removed
```

Measured impact on one machine: 24 orphaned Chrome instances holding 8.9 GB of RAM, and 32 stale profile directories holding 3.3 GB of disk, accumulated at roughly 5 per day.

**Platform**
- [x] Claude Code

**Version / install method**

Plugin `caveman@caveman` at `df2ccd85c94e`, binary release `bin-v1.1.4`, installed from the marketplace. macOS 26.6.2 (Darwin 25.6.0), Chrome 152.0.7977.77.

**Suggested fix**

Mirror the shape #742 landed in `src/mcp-servers/caveman-shrink/index.js`: in `main()`, `signal.Notify` on SIGTERM/SIGINT/SIGHUP and run the same cleanup before exiting.

As a host-side workaround until then, a wrapper that owns the child's stdin and closes it on those signals is enough, since the EOF path already cleans up correctly.

AmirF194:
Fixes #1016.

`main()` never called `signal.Notify`, so an MCP host's SIGTERM at session
teardown killed `caveman-browse` before its deferred `driver.Close()` and
`store.Close()` ran. Only the stdin-EOF path (`srv.Serve` returning `nil` on
`io.EOF`) reached those defers, so a signal-driven shutdown orphaned Chrome
and left its temp profile directory on disk. Same shape as #742/#985, ported
to the browse driver's Go binary.

`srv.Serve(os.Stdin, os.Stdout)` blocks reading stdin until EOF, so it can't
be interrupted directly. The fix runs it on its own goroutine behind a new
`runUntilSignal(ctx, srv, in, out)`, with `ctx` from
`signal.NotifyContext(..., os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)`.
On a signal, `ctx.Done()` wins the `select` and `runUntilSignal` returns
`nil` without waiting for `Serve`, so `main` returns normally and its
deferred `Close` calls run.

Verification (Docker, `golang:1.26`, matching `go.mod`):
- Added `TestRunUntilSignalReturnsOnContextDoneWithoutWaitingForServe`, which
  drives an unclosed `io.Pipe` reader (blocks forever, like stdin with no
  EOF) through an already-canceled `ctx` and asserts a prompt return instead
  of an indefinite block; it does not compile before this change, since
  `runUntilSignal` is new. `TestRunUntilSignalReturnsServeError` pins the
  unchanged EOF path.
- `go build ./...`, `go vet ./...`, and `go test ./...` (the repo's own `go`
  CI job) all pass.
- I couldn't run the issue's own live repro (checking the orphaned Chrome
  process's PPID and the leftover profile directory), since this environment
  has no real Chrome binary; the tests above exercise the same control-flow
  defect without one.

### DIFF
--- a/browse/cmd/caveman-browse/main.go
+++ b/browse/cmd/caveman-browse/main.go
@@ -9,10 +9,12 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/JuliusBrussee/caveman/browse"
@@ -52,9 +54,26 @@ func main() {
 	defer session.Close()
 
 	srv := mcp.NewServer("caveman-browse", browse.BrowserTools(session), logger)
-	if err := srv.Serve(os.Stdin, os.Stdout); err != nil {
+
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+	defer cancel()
+	if err := runUntilSignal(ctx, srv, os.Stdin, os.Stdout); err != nil {
 		logger.Error("serve", "err", err)
 		os.Exit(1)
+	}
+}
+
+// runUntilSignal returns as soon as srv.Serve finishes or ctx is done, instead of
+// blocking on it directly: an unwrapped call cannot be interrupted by a signal, so
+// main's deferred Close calls never run and the browser is orphaned.
+func runUntilSignal(ctx context.Context, srv *mcp.Server, in io.Reader, out io.Writer) error {
+	serveErr := make(chan error, 1)
+	go func() { serveErr <- srv.Serve(in, out) }()
+	select {
+	case err := <-serveErr:
+		return err
+	case <-ctx.Done():
+		return nil
 	}
 }
 
--- a/browse/cmd/caveman-browse/main_test.go
+++ b/browse/cmd/caveman-browse/main_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"log/slog"
@@ -9,8 +10,10 @@ import (
 	"runtime"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/JuliusBrussee/caveman/browse"
+	"github.com/JuliusBrussee/caveman/mcp"
 )
 
 func TestOpenRecoveryStoreCreatesFreshCavemanHome(t *testing.T) {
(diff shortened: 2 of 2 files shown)


JuliusBrussee:
Reviewed this closely — the patch is right and the diagnosis is exact. `signal.NotifyContext` alone would not have helped, because `srv.Serve` blocks reading stdin and never observes the context; wrapping it so `ctx.Done()` can win is the minimal shape. Both tests earn their place, especially the unclosed `io.Pipe` case, which is the only one that would catch a regression to the bare call.

What stops me adopting it: `browse/` in this repo is a consumer copy. Per CLAUDE.md's repository routing, browse driver work belongs to `JuliusBrussee/caveman-browse`, so a merge here would be overwritten on the next sync. Re-targeting it there should be a clean port — the diff doesn't depend on anything in this repo.

Not a defect in your work; a routing question for the maintainer, raised in #1015. Left open.

---

JuliusBrussee:
**Right fix, wrong repository.** `browse/` in this repo is a consumer copy; the source of truth is [`JuliusBrussee/caveman-browse`](https://github.com/JuliusBrussee/caveman-browse), and changes here are limited to pinned integration or an explicitly requested cross-repo sync. A signal-handler fix landing only in this copy would be overwritten by the next sync.

The defect itself (#1016) reads as real and the parallel to #742 looks right. Re-opening this against `caveman-browse` should get it moving; the diff should port as-is.

Not closing anything — that is the maintainer's call, and they may prefer to keep this open as a tracking reference.

---

JuliusBrussee:
Not actioned here — repo routing, not a judgement on the change.

This PR touches only `browse/**`, which `CLAUDE.md` routes to `JuliusBrussee/caveman-browse`. The `browse/` tree in this repo is a consumer copy that maintainers edit only for a pinned integration bump or an explicitly requested cross-repo sync, so a daily-triage run can't land it. Same for issue #1016.

The change itself reads well: `srv.Serve` blocking on stdin until EOF really does make the bare call uninterruptible, so `main`'s deferred `Close` calls never run on SIGTERM, and both tests pin the two paths (ctx-done and ordinary EOF) rather than just the new one.

@JuliusBrussee would need to say whether this lands upstream in `caveman-browse` and syncs back, or lands here directly.

---

PR #1022, opened 2026-09-09
docs: clarify caveman skill routing

vladimirtsk:
## Summary
- shorten the Caveman skill description to fit compact agent catalogs
- keep Caveman opt-in by making explicit mode selection the routing trigger
- leave the behavioral prompt unchanged

## Validation
- YAML frontmatter parses
- description is 57 characters and ends with a period
- `python3 tests/verify_repo.py` also fails on upstream `main` at the existing synchronized-copy check; transiently syncing generated copies allowed the changed description to pass those consistency checks before the unrelated package-manifest check

### DIFF
--- a/skills/caveman/SKILL.md
+++ b/skills/caveman/SKILL.md
@@ -1,9 +1,6 @@
 ---
 name: caveman
-description: >
-  Ultra-compressed communication mode that cuts output tokens while keeping
-  technical accuracy. Levels: lite, full, ultra and the wenyan variants. Use for
-  /caveman, "caveman mode", "talk like caveman", "be brief" or "less tokens".
+description: Use for terse technical replies or explicit caveman mode.
 ---
 
 Respond terse like smart caveman. All technical substance stay. Only fluff die.


JuliusBrussee:
Reviewed, not merged — this is an activation-policy change rather than a docs change, so it needs the maintainer.

The `description` frontmatter is what drives skill activation, and the short form drops the trigger phrases the README promises and the hook mirrors in its natural-language matching (`/caveman`, "caveman mode", "talk like caveman", "be brief", "less tokens").

There's a genuine argument on your side, which is why I'm not calling this wrong: CLAUDE.md notes every `skills/*/` description "competes for activation on ordinary work", and a narrower description is one of the few levers against over-triggering. But that's a tradeoff between under- and over-activation that only an eval run can price, and this triage doesn't ship SKILL.md behavior changes it can't measure.

Left open for @JuliusBrussee. If you have a case where the current description mis-fires on ordinary work, adding it here would help make the call.

---

PR #1025, opened 2026-09-09
fix(ccr): dedupe RepositoryMap payloads across sessions

AmirF194:
`PutObject` computes a typed object's ID from `Type + SessionID + Source + RepositoryState + ContentHash`, so `ON CONFLICT(object_id) DO NOTHING` only dedups within one session. A `RepositoryMap` object gets a fresh `SessionID` every session (it's rebuilt from scratch each time), but its `ContentHash` repeats whenever the underlying repository state hasn't changed since the last session. The result: an unchanged repo stores a full copy of the same ~MB-scale map on every session, growing `ccr.db` without bound.

I confirmed this is specific to `RepositoryMap`, not a missing lookup that every type shares: forcing `TaskDecision` through the same two-session scenario with matching content shows the same non-dedup, because there is no per-type content lookup at all, only the ID-collision check above. Other types legitimately produce different content per session even when it happens to match, so applying the same shortcut to all of them would be wrong (the second test below pins that boundary).

Fix: when a `RepositoryMap` put finds an existing row with the same `(source, repository_state, content_hash)` from any session, store a small row that references that row's data (`data_ref`) instead of writing another full copy. `GetObject`/`ListSessionObjects`/`FindTaskDecision` resolve the reference transparently, so every caller still sees the exact original bytes and each session still gets its own `object_id` (session-scoped listing is unaffected). Nothing in the table is ever deleted or the `data`/`data_ref` columns ever updated after insert, so a referenced row can't later be pointed at a row that becomes a pointer itself.

The new schema column needed a migration (`ALTER TABLE typed_objects ADD COLUMN data_ref ...`), and adding it exposed a real bug in that pattern under concurrent processes: `mem/cmd/cavemem`'s own `TestConcurrentRememberSubprocesses` (24 subprocesses opening the same fresh db) failed with `duplicate column name: data_ref` two out of five runs, because the PRAGMA-check-then-ALTER sequence isn't atomic across processes sharing one file. Fixed by treating that specific error as "another process already added it" rather than a real failure.

Verification:
- New regression test: two sessions putting the same repository map grow storage by roughly 2x the payload size before this change and roughly 0x after, and both sessions still read back the exact bytes.
- A second new test locks the scope: two sessions putting the same `TaskDecision` content still store two full copies, so that type must keep its own per-session data.
- `go build ./... && go vet ./... && go test ./...` (whole module, matches the `go` CI job) is green, including `mem/cmd/cavemem`'s concurrent-subprocess test run 5 times in a row after the migration fix.
- Opened a database written by the previous schema directly and confirmed it opens cleanly under the new code with its existing rows read back unchanged.

I only ran this in Linux containers, so I can't speak to the `macos` and `windows` CI legs specifically, though nothing here is filesystem- or OS-dependent.

### DIFF
--- a/engine/ccr/store_repositorymap_dedup_test.go
+++ b/engine/ccr/store_repositorymap_dedup_test.go
@@ -1,0 +1,123 @@
+package ccr_test
+
+import (
+	"testing"
+
+	"github.com/JuliusBrussee/caveman/engine/ccr"
+)
+
+func TestRepositoryMapDedupedAcrossSessions(t *testing.T) {
+	s, err := ccr.OpenMemory()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer s.Close()
+
+	payload := make([]byte, 200000)
+	for i := range payload {
+		payload[i] = byte(i)
+	}
+
+	before, err := s.Summary()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	id1, err := s.PutObject(ccr.Object{
+		Type: ccr.ObjectRepositoryMap, Source: "native:repository-map",
+		SessionID: "session-a", RepositoryState: "git:abc123",
+		TransformVersion: "repository-map-v1", Currentness: ccr.Current,
+		Lifecycle: ccr.Warm, Data: payload,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	id2, err := s.PutObject(ccr.Object{
+		Type: ccr.ObjectRepositoryMap, Source: "native:repository-map",
+		SessionID: "session-b", RepositoryState: "git:abc123",
+		TransformVersion: "repository-map-v1", Currentness: ccr.Current,
+		Lifecycle: ccr.Warm, Data: payload,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if id1 == id2 {
+		t.Fatal("expected distinct object ids per session (ListSessionObjects must still find each session's own row)")
+	}
+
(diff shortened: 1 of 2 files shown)


JuliusBrussee:
Adopted into #1015, cherry-picked with authorship preserved. Good diagnosis and real measurement, and scoping the dedup to `RepositoryMap` because `FindTaskDecision` json_extracts from `data` was the right call.

Two things I extended, both invisible to your tests rather than mistakes in them:

1. The key included `repository_state`, which is itself part of the object ID and advances on every commit — so one long session over a large repo still re-stored the map per commit. `content_hash` is `sha256(data)` verified in `prepareObject`, so it's sufficient alone. Cross-commit case went 600 KB → 200 KB.
2. `resolveObjectData` ran inside the open `rows` cursor in `ListSessionObjects`. The store is `SetMaxOpenConns(1)`, so the nested query waits forever on the connection the cursor holds — a hang, on the path `transitionAfterCompaction` and the receipt builder use. Now drained and closed first, as `Summary()` already does. Your `GetObject` tests couldn't see it; `QueryRow.Scan` closes before resolving.

Thanks — this was a large real win.

---

AmirF194:
Both extensions are real. The cursor deadlock especially, that's a nastier failure mode than what I found and my tests wouldn't have caught it either since GetObject never holds the cursor open. Good catch.

JuliusBrussee:

AmirF194:
Landed via #1015 with authorship intact, good outcome.

PR #1027, opened 2026-09-09
fix(proxy): parse compat-mount usage with a configurable provider dialect

### LINKED ISSUE #1026: provider_malformed usage accounting on a Z.AI compat mount (Claude Code traffic): zero requests compression-eligible
**Environment**

- caveman-proxy `bin-v1.1.6` (per `.bin-manifest.json`); `caveman-proxy version` prints `dev`; Linux x86_64
- Self-configured compat mount: `/compat/zai` -> `https://api.z.ai/api/anthropic` (GLM coding plan, Anthropic-compatible endpoint); mode `compress`
- Client: Claude Code, all model aliases mapped to `glm-5.3` / `glm-5.3-flash`

**Symptom**

`caveman-proxy stats` after ~2,500 requests in one day:

```json
"token_accounting": {"provider_complete": 74, "provider_malformed": 2361, "unavailable": 122},
"compression_tokens_before": 0,
"compression_tokens_saved": 0,
"requests_eligible_for_compression": 0,
"compression_token_count_basis": "unavailable"
```

92% of responses classify `provider_malformed`, so no provider token arithmetic runs and nothing becomes compression-eligible on this mount.

**What the upstream returns (captured through the proxy)**

Non-streaming, final usage block:

```json
"usage": {"input_tokens": 19, "output_tokens": 3, "cache_read_input_tokens": 0,
          "server_tool_use": {"web_search_requests": 0}, "service_tier": "standard"}
```

Streaming `message_start` (input count zero, no cache fields):

```json
data: {"type": "message_start", "message": {"id": "...", "usage": {"input_tokens": 0, "output_tokens": 0}}}
```

Streaming `message_delta` carries the real usage (`input_tokens: 19`), again without `cache_creation_input_tokens`.

Deviations from the Anthropic usage contract:

1. `message_start.usage.input_tokens` is always 0.
2. `cache_creation_input_tokens` never appears in any usage block.
3. Extra nonstandard fields are present (`server_tool_use`, `service_tier`).
4. The shape is inconsistent between calls (see reproduction).

**Correlation in the `requests` table**

- Every `provider_complete` row has `cache_status` `hit` or `miss` (72/72 streamed, same pattern non-streamed).
- Every `provider_malformed` row has `cache_status` `unknown` (1,795/1,795 streamed).

So the completeness verdict appears to hang on cache-field parsing, which this provider emits only sometimes.

**Reproduction**

- Hand-rolled curl requests through the mount (small ones, and a 340 KB streamed response with thinking, tools and `anthropic-beta` headers) usually parse `provider_complete`.
- A real one-shot `claude -p "say ok"` through the mount: 337 KB request, valid 922-byte response, classified `provider_malformed`.
- Two near-identical back-to-back curl probes (same body, seconds apart): one `provider_complete` / `miss`, the next `provider_malformed` / `unknown`. The provider's payload shape itself flip-flops between calls.

**Ask**

Lenient usage accounting for compat mounts, or a per-mount flag for it: treat a missing `cache_creation_input_tokens` as 0, take input tokens from the final usage block when `message_start` reports zeros, and ignore unknown extra fields — so a mount that is Anthropic-shaped-but-not-exact can still account tokens. Happy to test a build against real traffic and report the malformed rate before/after.

sebastianrueckerai:
Fixes #1026.

## Root cause

Named compat mounts are built with `Provider: "openai_compatible"`, so usage
accounting runs through the shared OpenAI-shape parser. An upstream that
answers the Anthropic Messages wire shape on such a mount reports
`input_tokens` **exclusive** of cache reads/writes. The OpenAI-shape parser
treats `input_tokens` as cache-inclusive, so every cache-warm response trips
the cache-total-over-input contradiction check in `mergeSnakeUsage` and
classifies `provider_malformed` with `cache_status: unknown`. The verdict
flips with cache temperature because only warm rows carry a nonzero cache
read — which is why the issue's back-to-back identical probes alternated
complete/malformed.

Captured on the reporting mount (Z.AI GLM, before this change):

- cache-cold: `{"input_tokens":1316,"output_tokens":16,"cache_read_input_tokens":0,...}` → `provider_complete` / `miss`
- cache-warm, same body seconds later: `{"input_tokens":36,"output_tokens":4,"cache_read_input_tokens":1280,"server_tool_use":{"web_search_requests":0},"service_tier":"standard"}` → `provider_malformed` / `unknown`

The other three deviations from the issue were already handled and are now
pinned by tests: `message_start` zeros merge away against the final usage
block, a missing `cache_creation_input_tokens` counts as 0, and the extra
`server_tool_use` / `service_tier` fields are ignored.

## Change

New per-mount option `compat.<name>.usage_dialect` (default empty = shared
OpenAI-shape parser, behavior unchanged for every existing mount and
first-party adapter). `"anthropic"` keys the mount's usage parser to Anthropic
semantics. The dialect changes usage parsing only — routing, header mapping,
telemetry provider, and pricing keep the mount's `openai_compatible`
identity. Unknown dialect values fail config load instead of silently
reverting to the misreading parser.

Plumbed as `Base.UsageProvider` (empty keeps `Provider`), consumed by
`ParseUsage` and `NewUsageScanner`. Documented in
`docs/technical/configuration.md`.

## Tests

Captured-shape tests at the scanner level (warm cache hit, cold cache miss,
non-stream), the default-off verdict (unchanged malformed classification
without the opt-in), config parsing and validation, and the standalone
wiring. Full Go suite: 68/68 packages pass. A mutation check (dialect mapping
neutered) fails the warm tests with exactly the reported symptom, so the
tests catch the regression they guard against.

## Verification on the reporting mount

Before (2,705 rows, official bin-v1.1.6): 2,478 `provider_malformed` (91.6%),
`cache_status unknown` on every malformed row; 100 `provider_complete`
(75 `hit` / 25 `miss`); 122 `unavailable`.

After this build + `usage_dialect: anthropic` (33-minute window of live
Claude Code traffic, 45 requests, stream and non-stream):

- 45/45 `provider_complete` (100%), all `cache_status hit`, 0 malformed,
  0 unavailable
- token arithmetic now matches the wire: `input_tokens` 1316 = 36 fresh +
  1280 cached on the probe bodies
- probe matrix (2× stream, 2× non-stream, cache-cold then warm): 4/4
  `provider_complete` / `hit`
- cumulative `provider_malformed` counter stopped growing at the restart;
  the last malformed row was written by the old process 0.1 s before the new
  one started listening

`requests_eligible_for_compression` stays 0 on this mount and is NOT caused
by the usage misclassification: eligibility is gated by the compression
recovery path (`mcpRecoveryAvailable` — the caveman MCP retrieve tool or
wrap-verified recovery), which this client does not provide. That is a
separate mechanism this PR does not change.

### DIFF
--- a/docs/technical/configuration.md
+++ b/docs/technical/configuration.md
@@ -128,7 +128,30 @@ compat:
   local-model:
     base_url: http://127.0.0.1:11434/v1
     api_key_env: LOCAL_MODEL_API_KEY
+  zai:
+    base_url: https://api.z.ai/api/anthropic
+    wire_dialect: anthropic
 ```
+
+`wire_dialect` declares the wire grammar the mount's upstream actually speaks.
+Empty (the default) keeps the shared OpenAI shape everywhere. Set `anthropic`
+only when the upstream answers the Anthropic Messages protocol on the mount.
+It changes two behaviors, both scoped to Anthropic Messages-path requests:
+
+- Usage accounting parses with the Anthropic dialect. Such an endpoint reports
+  `input_tokens` exclusive of cache reads/writes, which the OpenAI-shape
+  contradiction check misreads as malformed usage on every cache-warm
+  response, so token accounting (and with it cache status and compression
+  eligibility) silently drops those rows.
+- Compression zones follow the Anthropic Messages grammar, which keys the
+  live/frozen boundary on the request's own `cache_control` breakpoints. The
+  OpenAI grammar would mark only the latest messages live, leaving the
+  uncached post-breakpoint tail frozen or rewriting a block the provider
+  already cached.
+
+Routing, headers, and pricing keep the mount's OpenAI-compatible identity in
+every dialect, and OpenAI-protocol paths on the same mount keep the OpenAI
+grammar.
 
 Self-hosted private or loopback upstreams require an explicit
 `CAVE_SSRF_ALLOWLIST` entry. See [Security and privacy](security-and-privacy.md).
--- a/docs/technical/proxy-and-providers.md
+++ b/docs/technical/proxy-and-providers.md
@@ -128,6 +128,13 @@ the key in `Authorization: Bearer`. A real inbound Bearer token keeps its header
 on every path. OpenCode Go rejects a Bearer header on its Anthropic path, so
 this rule is necessary for the `anthropic-messages` models.
 
+The header is the only thing the path decides on its own. If the upstream
+answers the Anthropic Messages protocol, also declare `wire_dialect: anthropic`
+on the mount, or its usage blocks are parsed with OpenAI cache semantics and
+every cache-warm response is recorded as malformed usage — which drops the
+request from token accounting and from compression eligibility. See
+[Configuration](configuration.md).
+
 ## Modes
 
 | Mode | Request behavior |
--- a/proxy/internal/config/config.go
+++ b/proxy/internal/config/config.go
@@ -78,6 +78,14 @@ type ProviderConfig struct {
 type CompatConfig struct {
 	BaseURL   string `yaml:"base_url"`
 	APIKeyEnv string `yaml:"api_key_env"`
+	// WireDialect selects which provider's usage-accounting dialect the mount
+	// parses responses with. Empty keeps the shared OpenAI-shape parser.
+	// "anthropic" is for a mount whose upstream answers the Anthropic Messages
+	// wire shape: its usage block reports input_tokens exclusive of cache
+	// reads/writes, which the OpenAI-shape contradiction check misreads as
+	// malformed on every cache-warm row (issue #1026). Applies to the whole
+	// mount, not per path. Unknown values fail config load.
+	WireDialect string `yaml:"wire_dialect"`
 }
 
 // knownModes is the set of accepted runtime modes; anything else fails closed to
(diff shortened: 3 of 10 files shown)


JuliusBrussee:
Adopted into #1015, cherry-picked with authorship preserved. Fixtures captured from the real upstream, a validator that fails closed on an unknown dialect, and a default-path test proving first-party mounts are untouched — that's the shape that makes a config-flag change safe to take.

I verified it's load-bearing rather than trusting the tests: deleting the `Base.UsageProvider` override reddens the warm-cache and non-stream cases with `Malformed:true CacheStatus:unknown`, the reporter's exact signature.

One change: `anthropicWireZones` carried its own copy of the Messages path set, which `SanitizeAndMapHeaders` already answers via `anthropicMessagesPath` to choose `x-api-key` over Bearer. Routed through that one rule plus a table test asserting the two agree — a mount sending Anthropic auth headers while compressing under OpenAI zone rules is a bug waiting for the first Messages sub-path. Also added a pointer from `docs/technical/proxy-and-providers.md`, since its two-wire-protocols paragraph is where someone configuring this mount would look.

Your call to keep the dialect explicit rather than inferred is right, and I said so in the PR body: guessing wrong the other way inflates token counts instead of declining to report them.

---

sebastianrueckerai:
Thanks for the careful take-up in #1015. The anthropicMessagesPath routing is the right call — I cherry-picked 7903c195 back onto this branch (f362076) so the two stay identical, and the local instance that produced the verification numbers now runs exactly that code: probe after reinstall classifies provider_complete / hit / eligible on the mount, cache_bust still 0. Happy to close this in favor of #1015 whenever it lands.

JuliusBrussee:

PR #1029, opened 2026-09-10
fix(stats): price rule overhead at the cache rate before net verdict

pedrohlmelo:
`/caveman-stats` tells users on long agentic sessions to turn caveman off when the bill says caveman is saving them money.

`deriveNet()` subtracts the rules' input tokens (1,250 × turns) from the output tokens saved, as if both cost the same per token. They don't:

- every model in `MODEL_OUTPUT_PRICE_PER_M` bills input at 1/5 of its output rate;
- the rules ride the cached prefix, so after the first turn a Claude Code session reads them from cache at 0.1× the input rate.

Real session (Opus 5, 35 turns, every response a cache hit):

```
Est. tokens saved:     25,242 (~65% of output)
Est. rule overhead:    43,750 (input, ~1,250/turn over 35 turns)
Est. net:              -18,508 (caveman cost more than it saved for this workload — consider turning it off)
```

At billed rates that session saved about $0.63 of output and the rules cost about $0.03 of cached input. The lifetime view has the same problem: my `--all` shows `Est. net: -47,003 … consider turning it off` across sessions that were all net savings.

This goes in the same direction as docs/HONEST-NUMBERS.md ("Provider-billed A/B totals outrank output-only estimates"). The stats line was the one place that compared the two buckets 1:1.

## Approach

- `inputCostMultiplier(usage)` prices one input token of each response at that response's own logged mix: uncached 1×, cache write 1.25× (5m) / 2× (1h), cache read 0.1×. `parseSession` stores it per message.
- `ruleOverheadUsd()` sums 1,250 × multiplier × input price over the turns. Input price = output price × 0.2.
- `netLines()` adds `Est. net (USD)` when the model has a known price. The turn-it-off advice follows the USD net when there is one. When the token net is negative but the USD net is positive, the token line keeps its number and is relabelled `raw token count — the rules are mostly cached input, billed far below output; see USD net`.
- History rows gain `est_rule_overhead_usd` (omitted when the model is unpriced). The lifetime view shows a USD net only when every netted row carries the field, so the token and USD nets always describe the same sessions.

Every uncertain input is resolved so the rules look more expensive, never cheaper:

- a cache write with no TTL split is priced at the 1h rate (2×);
- Fable 5.1's 0.025× read rate is not used (0.1× everywhere);
- a response with no input breakdown is priced as fully uncached;
- the rules sit at the front of the prompt, the part most likely to be cached, yet each turn is priced at its average input token.

Unchanged: the token overhead and net lines (same numbers), `CAVEMAN_RULE_OVERHEAD_TOKENS`, and all output for unpriced or unknown models. The #145 regime (short, uncached replies) still reports a loss on both lines.

After (same 35-turn shape, test fixture):

```
Est. rule overhead:    43,750 (input, ~1,250/turn over 35 turns)
Est. net:              -18,530 (raw token count — the rules are mostly cached input, billed far below output; see USD net)
Est. net (USD):        +$0.603 (rules priced at this session's input mix, ~0.13× input rate)
```

## Testing

Six tests added to tests/test_caveman_stats.js. The first three below fail against unpatched main:

- cached 35-turn Opus 5 session: token net negative, USD net positive, no turn-it-off advice
- uncached single turn (#145): still negative on both lines, advice kept
- `inputCostMultiplier`: uncached / read / 5m / 1h / missing-TTL / empty usage
- history row carries `est_rule_overhead_usd` and the lifetime view nets it
- regression guards: no USD line for an unpriced model; no lifetime USD line when a netted row predates the field

Local runs: `tests/test_caveman_stats.js` 61/61, every `tests/*.js` file passes, `node --test tests/installer/*.test.mjs tests/hooks/*.test.mjs` 279/279, `python -m unittest discover -s tests` 137 OK, `tests/verify_repo.py` passes. `src/hooks/checksums.sha256` is regenerated for `caveman-stats.js`.

The stats SKILL.md and README each gain one sentence about the USD line. The hand-committed `plugins/caveman/skills/caveman-stats/SKILL.md` copy is left untouched.


JuliusBrussee:
**Reproduced, not adopted — maintainer decision.** The complaint is real and well evidenced: `deriveNet()` does subtract input tokens from output tokens 1:1, and a cached-prefix session can therefore be told to turn caveman off while the bill says the opposite. Your worked example is the clearest statement of it I've seen.

Two reasons it isn't on the triage branch:

1. The 1:1 sum is a *documented deliberate* choice, not an oversight — `caveman-stats.js:362` says so in as many words ("different buckets, but summing them is the only honest whole-budget delta"). Changing which number drives the turn-it-off advice is a policy call on a savings figure, and this routine is explicitly barred from changing savings numbers. It needs the maintainer.
2. It collides with #991/#994, which propose retiring the unsourced `COMPRESSION = 0.65` constant that `estSavedTokens` is derived from. If that lands, the basis of the net line changes underneath this and the USD net inherits the same unsourced estimate, now dressed in dollars. These two want deciding together.

One thing to fix regardless of the outcome: the diff edits `plugins/caveman/skills/caveman-stats/SKILL.md`. That's a mirror — CLAUDE.md says never edit synced copies, and that particular directory isn't touched by the sync workflow either, so an edit there propagates nowhere.

The conservative-rounding discipline in your approach (1h cache-write rate, no Fable read rate, unbreakdown-able responses priced as uncached) is the right instinct and worth keeping whichever way it goes.

---

pedrohlmelo:
Thanks for the triage. I've dropped the edit to `plugins/caveman/skills/caveman-stats/SKILL.md` (0838fbc), so the USD sentence now lives only in `skills/caveman-stats/SKILL.md` and its README.

Two notes for whoever makes the call:

- **On "the same estimate, now dressed in dollars":** `main` already prints `Est. saved (USD)` from that same `COMPRESSION` ratio (`caveman-stats.js:537` for the session view, `:461` for lifetime). This PR doesn't put a new dollar figure on the 0.65 estimate. It prices the *overhead* side with the same rate card, so the verdict stops weighing a cached input token as if it cost as much as an output token.
- **On #991 / #994:** agreed, these should be decided together. If #994 lands and the savings estimate goes away, `deriveNet`/`netLines` go with it and I'll close this. If the estimate stays, the turn-it-off advice is still wrong for any long cached session, and this is the fix. The overhead pricing (`inputCostMultiplier`, `ruleOverheadUsd`) doesn't depend on the 0.65 constant, so it would carry over to any sourced savings figure that replaces it.

pedrohlmelo:
Closing — superseded by 966a491 (merged in #1015), which removed the savings estimate, rule-overhead and net lines from `/caveman-stats` entirely. That's outcome A from my note on #991, so there's no net verdict left to fix. The cache-mix pricing idea lives on in `docs/technical/stats-accounting.md` for the proxy report. Thanks for the triage.

PR #1031, opened 2026-09-10
test(agent): gate SIGKILL escalation on readiness

### LINKED ISSUE #1030: test(agent): SIGKILL escalation test races worker startup
**What happen**

The test `sandbox timeout waits for stubborn worker close after SIGKILL escalation` fails on loaded CI runners. It failed on PR #1015 in the `engine-ci` job `typescript`. The PR changes no file under `packages/agent/`. The same test failed earlier on PR #974, which changed only markdown.

The test asserts an elapsed-time window:

```js
assert.equal(elapsed >= 1_200 && elapsed < 3_000, true);
```

The failed job reported 1050 ms, below the 1200 ms floor. The assertions for `result.text === "terminated"` and `cave_sandbox_timeout` passed.

- Failed job: https://github.com/JuliusBrussee/caveman/actions/runs/34450460971/job/102784780302
- Commit: `18ec31ccc1cf0d220b162312a1a2eebf05d7ca23`, run attempt 1, no rerun

**Expected**

The test proves that the runtime sends SIGTERM, escalates to SIGKILL after 250 ms, and waits for child close. The proof must not depend on worker startup time.

**Before/after example**

Got (CI, Node 22):

```text
not ok 290 - sandbox timeout waits for stubborn worker close after SIGKILL escalation
  duration_ms: 1050.442215
  Expected values to be strictly equal:

  false !== true

  packages/agent/tests/framework.runtime.mjs:5850:10
```

Want: the test passes on every runner, or fails only when the runtime does not escalate to SIGKILL or does not wait for close.

**Likely mechanism**

- The fixture tool `ignore_timeout` sets `timeoutMs: 1_000` and installs `process.on("SIGTERM", ...)` only inside `execute()` ([fixture](https://github.com/JuliusBrussee/caveman/blob/18ec31ccc1cf0d220b162312a1a2eebf05d7ca23/packages/agent/tests/fixtures/sandbox-agent.mjs#L157)).
- The runtime starts the deadline before `spawn` ([runtime.ts#L4755](https://github.com/JuliusBrussee/caveman/blob/18ec31ccc1cf0d220b162312a1a2eebf05d7ca23/packages/agent/src/runtime.ts#L4755)). On timeout it sends SIGTERM and schedules SIGKILL 250 ms later ([runtime.ts#L4791](https://github.com/JuliusBrussee/caveman/blob/18ec31ccc1cf0d220b162312a1a2eebf05d7ca23/packages/agent/src/runtime.ts#L4791)).
- If worker startup consumes the deadline, SIGTERM arrives before the handler exists. The worker exits at once. The SIGKILL timer never fires. `elapsed` lands near 1000 ms.

This is the likely cause, not the proven one. `terminalError ??= error` keeps the first error, so it masks every exit cause after the abort. The runtime settles on `close`, not `exit`, so a process that exited just before the abort but closed after it is masked too. The passing assertions prove only that the close handler did not settle before the abort and that close came inside the 250 ms window. No runtime channel exposes worker readiness or the terminating signal.

An isolated probe with the same 1000 ms deadline and 250 ms escalation reproduces the mechanism (Linux, Node 24.8.0):

| Handler installed | Exit signal | Elapsed |
| --- | --- | --- |
| Before deadline | SIGKILL | 1320 ms |
| After deadline | SIGTERM | 1059 ms |

**Proposed fix**

1. Add a package-internal sandbox lifecycle observer to `executeSandboxedTool`: stderr chunks as they arrive, each signal sent, and close with code and signal. Do not re-export it from `src/index.ts`. Do not change the deadline or the `terminalError` semantics.
2. Make the fixture signal readiness. After it installs the SIGTERM handler, it writes one marker line to stderr. The worker runs under `unshare --net`, so loopback is not available; stderr is an inherited descriptor and the parent already receives it live.
3. Gate the escalation test on that marker. Give the tool a long `timeoutMs`, abort through `RunOptions.signal` when the marker arrives, then assert the ordered sequence SIGTERM → SIGKILL → close(SIGKILL) → run rejection. Prove that the rejection settled after close with a sequence counter, not wall-clock.
4. Add a separate startup-timeout test with a tiny `timeoutMs`. There, a prompt SIGTERM exit is correct.
5. Remove the elapsed-time assertion. Keep a bounded test timeout.

Lowering the 1200 ms floor lets the test pass without SIGKILL. Raising `timeoutMs` only moves the race threshold.

**Ownership**

`CLAUDE.md` routes Agent SDK work to `caveman-agent-sdk`. `engine-ci` runs the copy in this repository, so the fix must land here too. A PR with the readiness-gated test is in preparation.

**Related**

- #1015 — PR where the failure was observed; a comment proposes fake-clock or event assertions
- #974 — earlier failure of the same test on a markdown-only PR
- #995 — SIGKILL escalation for MCP shrink; does not change this test
- #942 — MCP subprocess timeouts; closed, does not address this test

**Platform**
- [x] Other: `packages/agent` test suite, `engine-ci` job `typescript`

**Version / install method**
- Commit `18ec31ccc1cf0d220b162312a1a2eebf05d7ca23`
- CI: Node 22, ubuntu runner

zsltg:
Fixes #1030.

The test `sandbox timeout waits for stubborn worker close after SIGKILL escalation` fails on loaded CI runners. It failed on #1015 and #974, which changed no file under `packages/agent/`. This PR removes the startup race from the test and replaces its wall-clock assertion with an event-order proof.

## Fixes

| Issue(s) | Root cause | Hosts/OS affected | Test added | Supersedes |
| --- | --- | --- | --- | --- |
| #1030 (seen on #1015, #974) | The fixture installed its SIGTERM handler inside `execute()`. The runtime starts the deadline before `spawn`. When startup consumed the deadline, SIGTERM arrived before the handler existed. The worker exited at once, the SIGKILL timer never fired, and `elapsed` fell below the 1200 ms floor. | `packages/agent` test suite, engine-ci job `typescript`, all OS | `sandbox abort waits for stubborn worker close after SIGKILL escalation` and `sandbox timeout during worker startup settles the run` in `tests/framework.runtime.mjs` | — |

## Change

- `src/runtime.ts`: adds a package-internal sandbox lifecycle observer. `executeSandboxedTool` reports each stderr chunk as it arrives, each signal it sends, and the child close event before the result settles. `src/index.ts` does not re-export the setter, so the public API does not change. An observer exception cannot change the tool outcome. The deadline, `terminate`, and `terminalError` semantics do not change.
- `tests/fixtures/sandbox-agent.mjs`: `ignore_timeout` writes one stderr marker after it installs the handler and uses a 10 s timeout. A new `startup_timeout` tool has a 1 ms deadline and no handler.
- `tests/framework.runtime.mjs`: the escalation test waits for the marker, aborts the run through `RunOptions.signal`, and asserts the order SIGTERM → SIGKILL → close(SIGKILL) → run rejection with a sequence counter. The startup-timeout test proves that a deadline during startup reaches the model context and settles the run.

The old floor of 1200 ms cannot be lowered without losing SIGKILL coverage. A longer `timeoutMs` alone only moves the race threshold.

## Verification

- `pnpm test` in `packages/agent`, in a privileged `node:22` container where user namespaces work: `tests 319 / pass 319 / fail 0`.
- The two new tests, 10 runs in a loop: `pass 2 / fail 0` on each run.
- Race proof: with `await new Promise((r) => setTimeout(r, 1_500))` inserted before the handler install, the escalation test still passes. The old test could not pass with that delay. The delay is not in the commit.
- `python3 tests/verify_repo.py`: all checks pass.
- Hosts with `kernel.apparmor_restrict_unprivileged_userns=1` cannot run the `unshare`-based sandbox tests. Those tests fail there before and after this change with `Operation not permitted`.

## Ownership

`CLAUDE.md` routes Agent SDK work to `caveman-agent-sdk`. engine-ci runs the copy in this repository, so the fix lands here. The change is small so that the SDK repository can mirror it.

### DIFF
--- a/packages/agent/tests/framework.runtime.mjs
+++ b/packages/agent/tests/framework.runtime.mjs
@@ -5900,16 +5900,41 @@ test("sandbox timeout during worker startup settles the run", { timeout: 15_000 
       return fauxAssistantMessage("terminated");
     },
   ]);
-  // The deadline expires during startup. A prompt exit on SIGTERM is correct
-  // here, thus this test asserts no SIGKILL escalation.
-  const result = await run(sandboxAgent, "timeout", {
-    ensureRuntime: false,
-    entryPath: "tests/fixtures/sandbox-agent.mjs",
-    model: faux.getModel(),
-    streamFn: faux.provider.streamSimple.bind(faux.provider),
+  // The deadline expires during startup. No SIGTERM handler exists, thus the
+  // worker must die on SIGTERM. The exit signal proves that. This test does
+  // not assert the absence of a SIGKILL event: the SIGKILL timer runs on time,
+  // not on child state, and under load it can fire after the child is already
+  // dead of SIGTERM but before the parent processes the close event.
+  let step = 0;
+  const events = [];
+  setSandboxLifecycleObserver((event) => {
+    if (event.kind !== "stderr") events.push({ at: ++step, ...event });
   });
+  let settledAt = 0;
+  let result;
+  try {
+    result = await run(sandboxAgent, "timeout", {
+      ensureRuntime: false,
+      entryPath: "tests/fixtures/sandbox-agent.mjs",
+      model: faux.getModel(),
+      streamFn: faux.provider.streamSimple.bind(faux.provider),
+    });
+  } finally {
+    settledAt = ++step;
+    setSandboxLifecycleObserver(undefined);
+  }
   assert.equal(result.text, "terminated");
   assert.match(observed, /cave_sandbox_timeout/);
+  const sigterm = events.find((event) => event.kind === "signal" && event.signal === "SIGTERM");
+  const close = events.find((event) => event.kind === "close");
+  assert.notEqual(sigterm, undefined);
+  assert.notEqual(close, undefined);
+  assert.equal(sigterm.at < close.at, true);
+  // The worker exited on SIGTERM: no handler was installed, thus no escalation
+  // was needed to end it.
+  assert.equal(close.signal, "SIGTERM");
+  // The run must not settle before the child closes.
+  assert.equal(close.at < settledAt, true);
 });
 
 test("sandbox rejects parent and worker definition drift before tool execution", async () => {
--- a/packages/agent/tests/framework.runtime.mjs
+++ b/packages/agent/tests/framework.runtime.mjs
@@ -50,6 +50,7 @@ import {
   resolveCaveRoute,
   runAgentInternal,
   sandboxSourceReadFlags,
+  setSandboxLifecycleObserver,
 } from "../dist/runtime.js";
 import { memoryFilePath, mutateMemories, readMemories } from "../dist/memory-store.js";
 import { sourceGraphSHA256 } from "../dist/source-graph.js";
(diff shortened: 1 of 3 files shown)


packages/agent/tests/framework.runtime.mjs
```diff
@@ -5827,27 +5828,88 @@ test("sandbox fails closed on child-process permission without OS tree containme
...
+  faux.setResponses([
+    fauxAssistantMessage(fauxToolCall("startup_timeout", {}, { id: "startup-timeout" })),
     (context) => {
       observed = JSON.stringify(context.messages);
       return fauxAssistantMessage("terminated");
     },
   ]);
-  const started = Date.now();
+  // The deadline expires during startup. A prompt exit on SIGTERM is correct
+  // here, thus this test asserts no SIGKILL escalation.
```
AmirF194: The comment on line 5903-5904 says this test "asserts no SIGKILL escalation," but nothing in the body checks that. `result.text` comes from the faux model's second scripted response, which is `"terminated"` unconditionally, not something derived from which signal path ran. The `observed` match on `cave_sandbox_timeout` only proves the initial timeout fired (`terminalError ??= new Error("cave_sandbox_timeout")` at runtime.ts:4854), and that error gets set the same way whether the runtime later exits promptly on SIGTERM or escalates all the way to SIGKILL.

So if a future change accidentally made this path go through the SIGKILL branch too, this test would still pass. The first test already wires up `setSandboxLifecycleObserver` for exactly this kind of proof; this one could reuse it and assert no `signal: SIGKILL` event was ever emitted, the way `ready.at < sigterm.at` etc. prove ordering above.

zsltg: Correct. The comment claimed an assertion that the body did not make. Fixed in f3990d0.

The test now uses the lifecycle observer. It asserts that the runtime sent SIGTERM, that the child closed with `signal === "SIGTERM"`, and that the run settled after the close event. The exit signal is fixed at death, thus it proves the prompt SIGTERM exit directly.

The test does not assert the absence of a `SIGKILL` event, on purpose. `killProcessTree` sends `process.kill(-pid, signal)` and swallows `ESRCH`. The SIGKILL timer fires 250 ms after SIGTERM on time, not on child state. Under load, the timers phase can run before the parent processes the close event, so the observer can record a SIGKILL sent to a group that is already dead of SIGTERM. That assertion would be timing-dependent, which is the class of flake this PR removes. `close.signal === "SIGTERM"` holds in both orders.

Verification in a privileged `node:22` container: both new tests 10/10 loop runs `pass 2 / fail 0`; full suite 319/319.

---

zsltg:
**Ownership of `packages/agent/src/runtime.ts` — needs a maintainer decision**

The [Devin review](https://app.devin.ai/review/juliusbrussee/caveman/pull/1031) flags `runtime.ts:4722-4729`: `packages/agent/` is a historical copy of `caveman-agent-sdk`, and repository rules restrict edits there to pinned integration or an explicit sync. The flag is correct. This PR adds a 57-line package-internal seam to that copy, and the PR body describes the SDK mirror as a later step. That is the reverse of the documented direction.

Why the PR is in this shape:

- `engine-ci` runs this copy. The flake blocks PRs in this repository (#974, #1015) whatever the SDK repository contains.
- A contributor cannot land the fix in `caveman-agent-sdk` first. That repository is private.
- A test-only fix is not possible. The test needs the observer seam in `runtime.ts` to see readiness and signals. Without it, only a wall-clock window remains, which is the flake.

The seam is small by design so that a mirror stays simple: one setter, one event type, three call sites; not exported from `src/index.ts`, so the public API does not change. The deadline, `terminate`, and `terminalError` semantics do not change.

Options:

1. **Accept this PR as an explicit reverse sync.** Then mirror the three files into `caveman-agent-sdk`. I can prepare the patch on request.
2. **Port to `caveman-agent-sdk` first, then sync here.** This PR closes in favor of that sync. The issue (#1030) and the tests are ready to copy.
3. **Reject the runtime change.** Then the test stays timing-based and the flake remains.

A maintainer needs to pick one. I will follow whichever applies.

---

packages/agent/tests/framework.runtime.mjs
AmirF194: Thanks for the fix and the explanation on skipping the SIGKILL assertion, the timer-vs-close race you described makes sense as a reason to keep that one timing independent.

JuliusBrussee:
**Maintainer decision — routing, not quality.** The approach is the right one: replacing a wall-clock window with an ordered SIGTERM → SIGKILL → close(SIGKILL) → rejection assertion removes the race instead of hiding it, and the separate startup-timeout test keeps the case the old assertion was accidentally covering. Proving it by inserting a 1.5 s delay before the handler install is the check I'd have asked for.

Two notes:

- I could not run it. No `packages/agent/node_modules` in this container and the sandbox tests need user namespaces, so this is a read, not a verification. Your own `319/319` stands unchallenged.
- This routine can't merge or adopt it: `packages/agent/**` is routed to `JuliusBrussee/caveman-agent-sdk`, and the fix landing here while engine-ci runs this copy is exactly the tension you flagged in #1030. That needs the maintainer, and it's in #1015's body.

Worth saying plainly: this is the check currently red on #1015, so it isn't a cosmetic flake — thanks for chasing it to the mechanism.

---

JuliusBrussee:
**Confirmed in the wild — this flake just failed CI on an unrelated PR.** Worth prioritising.

`sandbox timeout waits for stubborn worker close after SIGKILL escalation` failed on #1055 ([job](https://github.com/JuliusBrussee/caveman/actions/runs/34943910082/job/104298722018)) at `duration_ms: 1073`, i.e. under the `elapsed >= 1_200` floor — exactly the startup race you describe in #1030. That PR touches none of `packages/agent`, so it cost a reviewer a red check for nothing.

Measured the margin while diagnosing it: the test passes locally five out of five at **1289–1297 ms** against a 1200 ms floor. About 7 % of headroom on a wall-clock assertion, which is why it holds on most runners and not all.

Your diagnosis and approach both look right to me. Two things I liked on reading the diff:

- Replacing the wall-clock window with a sequence counter over lifecycle events is the correct shape — a loaded runner changes every duration but never the event order, and your comment says exactly that.
- Driving the abort off the `cave-sandbox-ready` stderr marker removes the race at its source rather than widening the window, which would only have moved the flake.

The `setSandboxLifecycleObserver` seam being package-internal and not re-exported from `src/index.ts`, with the observer's exceptions swallowed so observation can never change a tool's outcome, is the right call too.

I could not port this into #1055 myself: `packages/agent/` is out of scope for the daily triage run (source of truth is `caveman-agent-sdk`), and I lack permission to re-run the failed job (403). So this needs the maintainer.

Thanks for doing it properly instead of bumping the timeout.

---

JuliusBrussee:
Not actioned here — repo routing, not a judgement on the change.

This PR touches only `packages/agent/**`, which `CLAUDE.md` routes to `caveman-agent-sdk`; the copy in this repo is historical and edited only for a pinned integration bump or an explicitly requested cross-repo sync. A daily-triage run can't land it. Same for issue #1030.

The approach looks right: replacing the `elapsed >= 1_200 && elapsed < 3_000` wall-clock window with an ordered lifecycle-event assertion removes the startup race rather than papering over it, and the split into a separate "timeout during worker startup" case keeps the no-handler path covered. Deliberately not asserting the absence of SIGKILL in that second test is the correct call — that timer is time-driven, not child-state-driven.

@JuliusBrussee would need to route this to the SDK repo.

---

JuliusBrussee:
Evidence this is worth landing, if it helps prioritise: the test you're fixing just failed a real CI run.

`typescript` on #1089 (head `1b25285`):

```
packages/agent test: not ok 290 - sandbox timeout waits for stubborn worker close after SIGKILL escalation
```

`packages/agent/**` is byte-identical between that head, its base, and the previous head where the same job was green — so nothing in that PR touched it. It's the startup race #1030 describes, landing on an unrelated PR and costing it a red check.

Still out of scope for me to port (routed to `caveman-agent-sdk`, and this run can't edit `packages/agent/**`), so this is just a data point for @JuliusBrussee, not a request to you. Your approach — ordered lifecycle events instead of a wall-clock window — is the thing that actually removes the race.

---

PR #1033, opened 2026-09-11
fix(proxy): preserve provider-defined headers

aguanitallc:
## Summary

- replace the shared provider-header allowlist with transparent end-to-end forwarding
- strip proxy credentials, Caveman-private headers, and RFC hop-by-hop headers
- preserve the existing trace opt-in and provider-specific authentication/billing rules
- remove the now-redundant OpenCode-only forwarding path

## Why

Provider APIs can introduce required headers independently of Caveman releases. The fixed shared allowlist silently dropped `x-opencode-session`, causing OpenCode Go requests to fail with `MissingSessionID`. Fixing each provider header after an outage repeats the same bug class.

## Validation

- `go test ./proxy/providers/... ./proxy/internal/gateway/...`
- `go test ./proxy/...`
- live Hermes → Caveman → OpenCode Go request with an isolated config containing `fallback_providers: []`

### DIFF
--- a/proxy/providers/adapter.go
+++ b/proxy/providers/adapter.go
@@ -431,19 +431,25 @@ func (b Base) ResolveUpstreamURL(ctx context.Context, req *http.Request, route R
 }
 
 func (b Base) SanitizeAndMapHeaders(ctx context.Context, req *http.Request, credential Credential, _ *url.URL) (http.Header, error) {
-	out := http.Header{}
-	copyIfPresent(out, req.Header, "content-type")
-	copyIfPresent(out, req.Header, "accept")
-	copyIfPresent(out, req.Header, "accept-encoding")
-	copyIfPresent(out, req.Header, "idempotency-key")
-	copyIfPresent(out, req.Header, "openai-organization")
-	copyIfPresent(out, req.Header, "openai-project")
-	copyIfPresent(out, req.Header, "anthropic-version")
-	copyIfPresent(out, req.Header, "anthropic-beta")
-	copyIfPresent(out, req.Header, "api-version")
-	if env.Bool("CAVE_PROPAGATE_TRACE_HEADERS_UPSTREAM", false) {
-		copyIfPresent(out, req.Header, "traceparent")
-		copyIfPresent(out, req.Header, "tracestate")
+	out := req.Header.Clone()
+	for _, name := range []string{
+		"authorization", "proxy-authorization", "proxy-authenticate",
+		"connection", "proxy-connection", "keep-alive", "te", "trailer",
+		"transfer-encoding", "upgrade",
+	} {
+		out.Del(name)
+	}
+	for _, name := range strings.Split(req.Header.Get("connection"), ",") {
+		out.Del(strings.TrimSpace(name))
+	}
+	for name := range out {
+		if strings.HasPrefix(strings.ToLower(name), "x-cave-") {
+			out.Del(name)
+		}
+	}
+	if !env.Bool("CAVE_PROPAGATE_TRACE_HEADERS_UPSTREAM", false) {
+		out.Del("traceparent")
+		out.Del("tracestate")
 	}
 	out.Set("user-agent", appendUserAgent(req.UserAgent()))
 	switch b.Provider {
@@ -482,6 +488,7 @@ func (b Base) SanitizeAndMapHeaders(ctx context.Context, req *http.Request, cred
 				out.Set("authorization", "Bearer "+credential.Key)
 				out.Set("x-goog-user-project", quotaProject)
 			} else {
+				out.Del("x-goog-user-project")
 				out.Set("x-goog-api-key", credential.Key)
 			}
 		}
--- a/proxy/providers/adapter_contract_test.go
+++ b/proxy/providers/adapter_contract_test.go
@@ -147,12 +147,15 @@ func TestSanitizeAndMapHeaders_NoKeyLeakAndCorrectMapping(t *testing.T) {
 		t.Run(c.provider, func(t *testing.T) {
 			b := Base{Provider: c.provider}
 			req, _ := http.NewRequest(http.MethodPost, "/x", nil)
-			// Inbound client headers that MUST NOT leak upstream:
+			// Inbound proxy credentials and hop-by-hop headers must not leak upstream.
 			req.Header.Set("authorization", "Bearer CLIENT-CAVE-KEY")
 			req.Header.Set("x-cave-api-key", "cave_live_clientclient_secret")
 			req.Header.Set("x-cave-upstream-key", "should-not-be-copied-verbatim")
 			req.Header.Set("content-type", "application/json")
-			req.Header.Set("x-internal-secret", "nope")
+			req.Header.Set("x-opencode-session", "session-123")
+			req.Header.Set("x-provider-feature", "enabled")
+			req.Header.Set("connection", "x-hop-private")
+			req.Header.Set("x-hop-private", "nope")
 
 			out, err := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "UPSTREAM-PROVIDER-KEY"}, nil)
 			if err != nil {
@@ -180,19 +183,22 @@ func TestSanitizeAndMapHeaders_NoKeyLeakAndCorrectMapping(t *testing.T) {
 					}
 				}
 			}
-			// x-cave-* and arbitrary inbound secrets are not forwarded.
+			// Caveman-private and hop-by-hop headers are not forwarded.
 			if out.Get("x-cave-api-key") != "" {
 				t.Error("x-cave-api-key forwarded upstream")
 			}
 			if out.Get("x-cave-upstream-key") != "" {
 				t.Error("x-cave-upstream-key forwarded upstream")
 			}
-			if out.Get("x-internal-secret") != "" {
-				t.Error("non-allowlisted x-internal-secret forwarded upstream")
+			if out.Get("connection") != "" || out.Get("x-hop-private") != "" {
+				t.Error("hop-by-hop header forwarded upstream")
 			}
-			// Allowlisted passthrough still works.
+			// Provider-defined end-to-end headers pass through without an allowlist.
 			if out.Get("content-type") != "application/json" {
 				t.Error("content-type should pass through")
+			}
+			if out.Get("x-opencode-session") != "session-123" || out.Get("x-provider-feature") != "enabled" {
+				t.Error("provider-defined header was dropped")
 			}
 		})
 	}
--- a/proxy/providers/openaicompat/openaicompat.go
+++ b/proxy/providers/openaicompat/openaicompat.go
@@ -180,9 +180,7 @@ func (a namedAdapter) InspectRequest(ctx context.Context, body providers.BodyRea
 // inbound Bearer token keeps its scheme on every path, as the repository
 // preserves the auth scheme of an inbound credential. The placeholder token is
 // not a real credential. The gateway replaces it from the environment after
-// this mapping, so it must land in the header of the wire protocol. The method
-// also forwards the OpenCode session headers on every path of the opencode-go
-// mount. See openCodeSessionHeaders.
+// this mapping, so it must land in the header of the wire protocol.
 func (a namedAdapter) SanitizeAndMapHeaders(ctx context.Context, req *http.Request, credential providers.Credential, upstream *url.URL) (http.Header, error) {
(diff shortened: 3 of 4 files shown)


proxy/providers/adapter.go
```diff
@@ -431,19 +431,25 @@ func (b Base) ResolveUpstreamURL(ctx context.Context, req *http.Request, route R
...
-	copyIfPresent(out, req.Header, "idempotency-key")
-	copyIfPresent(out, req.Header, "openai-organization")
-	copyIfPresent(out, req.Header, "openai-project")
-	copyIfPresent(out, req.Header, "anthropic-version")
-	copyIfPresent(out, req.Header, "anthropic-beta")
-	copyIfPresent(out, req.Header, "api-version")
-	if env.Bool("CAVE_PROPAGATE_TRACE_HEADERS_UPSTREAM", false) {
-		copyIfPresent(out, req.Header, "traceparent")
-		copyIfPresent(out, req.Header, "tracestate")
+	out := req.Header.Clone()
```
AmirF194, line 434: I ran this against `de876ff` in a clean `golang:1.26` container.

`proxy/providers/adapter.go:434` switches header forwarding from an allowlist (`copyIfPresent`, 9 named headers) to `req.Header.Clone()` plus a denylist of hop-by-hop and `x-cave-*` names. `Cookie`, `Referer`, and `X-Forwarded-For` match none of the denylist entries, so they now forward to the upstream LLM provider unchanged. On `main` (`15581d1`) they were stripped, since the old allowlist never copied them.

```go
b := Base{Provider: "openai"}
req.Header.Set("cookie", "session=super-secret-proxy-session")
out, _ := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "UPSTREAM-KEY"}, nil)
// de876ff:  out.Get("cookie") == "session=super-secret-proxy-session"
// 15581d1:  out.Get("cookie") == ""
```

The PR's own new test only checks `x-cave-*`, `connection`, and two named hop-by-hop headers; it doesn't cover `Cookie` or other client-context headers the old allowlist implicitly dropped. I haven't checked whether any real caller of this proxy actually sends a `Cookie` header, only that the sanitizer no longer strips one if present. Adding those names to the denylist would close this without losing the provider-header fix.

zsltg (changes requested):
**Verdict: do not merge.**

The PR replaces a fix that `main` already contains. It also changes the header sanitizer from deny-by-default to allow-by-default. Three findings block. Two are risks. Each finding comes from a probe run on the PR diff applied to `15581d1`. The probe test is at the end.

**Relation to earlier work**

Issue #1006 reported that the fixed allowlist drops `x-opencode-session`. PR #1007 fixed this on the `/compat/opencode-go` mount only. That fix landed on `main` in #1012 (commit `733c262`). The base of this PR already contains it. Thus the `MissingSessionID` outage is not a reason to merge. The description gives a second reason: providers can add headers between releases. That is a change to the trust boundary, and it needs its own justification.

**`proxy/providers/adapter.go:434`: blocker. Inbound credentials and client context reach the upstream.**

`req.Header.Clone()` forwards each inbound header that the denylist does not name. The denylist does not name the credential headers that the proxy itself reads and writes. It also does not name client-context headers. Probe result, with `main` as the comparison:

| Inbound header | openai | anthropic | gemini | openai_compatible | `main` |
|---|---|---|---|---|---|
| `cookie: sid=secret` | forwarded | forwarded | forwarded | forwarded | dropped |
| `x-forwarded-for: 10.0.0.1` | forwarded | forwarded | forwarded | forwarded | dropped |
| `x-api-key: CLIENT-KEY` | forwarded | overwritten | forwarded | forwarded | dropped |
| `x-goog-api-key: GOOG-KEY` | forwarded | forwarded | overwritten | forwarded | dropped |

The same gap gives two credentials on the Anthropic Bearer path. With `Credential{Key: "OAUTH", Scheme: "bearer"}` and inbound `x-api-key: STALE`, the output has `Authorization: Bearer OAUTH` and `X-Api-Key: STALE`. The standalone resolver takes an inbound `x-api-key` before a Bearer token, so the standalone path does not get to this state. The `Resolve` interface is open, so another resolver can. The sanitizer must not depend on the resolver for this guarantee.

A longer denylist does not close this finding. A name that is not on the list passes, including headers that do not exist yet. Only explicit forwarding rules give deny-by-default. If you keep wide forwarding, state it as a boundary change. Then strip each supported inbound credential header before you map the resolved credential: `x-api-key`, `api-key`, `x-goog-api-key`, `x-goog-user-project`. Also strip `cookie`, `x-forwarded-*`, `x-real-ip`, and `referer`.

Test coverage is also missing. The PR deletes the `x-internal-secret` assertion in `adapter_contract_test.go:186`. That assertion was the only test for an arbitrary unlisted header. Add negative tests for an arbitrary secret, each provider credential header, client context, and the Bearer-plus-`x-api-key` case.

**`proxy/providers/adapter.go:442`: blocker. A second `Connection` line bypasses the strip.**

`req.Header.Get("connection")` returns the first value only. A request with `Connection: keep-alive` and a second line `Connection: x-hop-private` forwards `x-hop-private: secret`. Go's server keeps hop-by-hop names in the inbound header map, so the proxy must strip each value itself. Iterate `req.Header.Values("connection")` and split each value. Add a test with two field lines. See RFC 9110 §7.6.1.

**`proxy/providers/adapter.go:477`: blocker. The Gemini placeholder keeps a caller-chosen quota project.**

With the `no-key-required` Bearer placeholder and inbound `x-goog-user-project: caller-project`, the early `break` keeps the cloned quota header and sets no key. Then `applyUpstreamAuthFallback` sets `x-goog-api-key` from the environment and does not remove the quota header (`auth_fallback.go:24`). The upstream gets the operator's API key with a quota project that the caller chose. The deletion at line 491 covers only the direct API-key branch. On `main` this cannot occur, because the allowlist never copies the header for `gemini`. Strip the quota header before the branch. Set it only for validated Gemini OAuth and the existing Vertex policy. Add a test for the placeholder with an inbound quota header.

**`proxy/providers/adapter.go:499`: risk. Vertex headers are sent twice.**

The `vertex` branch still calls `copyIfPresent` for `x-goog-user-project` and `x-vertex-ai-llm-request-type`. After the clone, each call adds the value a second time. The probe shows `[qp qp]` and `[shared shared]`. These are billing and routing headers. An upstream can reject or combine repeated fields. Remove the two calls. Add a test that reads `Header.Values`, not only `Header.Get`. I did not do a live Vertex test.

**`proxy/providers/openaicompat/openaicompat_test.go:360`: risk. Session identity goes to unrelated mounts.**

The PR inverts the other-mount assertion in `TestNewNamed_ForwardsOpenCodeSessionHeaders`. The test now requires that a compat mount other than `opencode-go` receives the four `x-opencode-*` headers. Issue #1006 asked for the opposite: another mount has no use for these headers and must not learn the session identity of the caller. Keep the mount gate and the other-mount negative assertion from PR #1007.

**PR description: nit.**

The "Why" section says that the allowlist "silently dropped `x-opencode-session`". That was true before `733c262`. It is not true on the base of this PR. State that the PR replaces an existing fix with a general policy, and give the reason for the policy on its own.

**What I did and did not do**

- I applied the PR diff to a scratch copy of `15581d1` and ran the probe test that follows. I did not run the full suite, and I made no live provider calls.
- The Gemini direct API-key branch now deletes `x-goog-user-project` before it sets `x-goog-api-key`. That part is correct.
- `Host` and `Content-Length` are not a concern. Go stores `Host` on `Request.Host`, and the client transport writes `Content-Length` from `Request.ContentLength`.
- The gateway assigns the returned map directly to the upstream request (`proxy/internal/gateway/proxy.go:348`). There is no later sanitizer.
- @AmirF194 reported the `cookie` case. The other findings are new.

**Recommendation**

Close this PR in favor of the merged #1012 fix. If the maintainers want transparent forwarding as a policy, open a new PR that does all of the following:

- Keeps explicit forwarding rules for credential and client-context headers.
- Iterates each `Connection` value.
- Removes the duplicated Vertex copies.
- Strips the Gemini quota header before the placeholder branch.
- Restores the other-mount negative assertion.
- Restores a negative test for an arbitrary unlisted header.

<details>
<summary>Probe test</summary>

Apply `gh pr diff 1033` to a checkout of `15581d1`. Put this file in `proxy/providers/`. Then run `go test ./providers/ -run TestProbe -v` from `proxy/`.

```go
package providers

import (
	"net/http"
	"testing"
)

func TestProbe(t *testing.T) {
	// Cross-provider credential and client-context forwarding.
	for _, p := range []string{"openai", "anthropic", "gemini", "openai_compatible"} {
		b := Base{Provider: p}
		req, _ := http.NewRequest(http.MethodPost, "/x", nil)
		req.Header.Set("x-api-key", "CLIENT-KEY")
		req.Header.Set("cookie", "sid=secret")
		req.Header.Set("x-forwarded-for", "10.0.0.1")
		req.Header.Set("x-goog-api-key", "GOOG-KEY")
		out, _ := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "UPSTREAM-KEY"}, nil)
		t.Logf("%s x-api-key=%q cookie=%q xff=%q x-goog-api-key=%q", p,
			out.Get("x-api-key"), out.Get("cookie"), out.Get("x-forwarded-for"), out.Get("x-goog-api-key"))
	}
	// Second Connection line.
	{
		b := Base{Provider: "openai"}
		req, _ := http.NewRequest(http.MethodPost, "/x", nil)
		req.Header.Add("connection", "keep-alive")
		req.Header.Add("connection", "x-hop-private")
		req.Header.Set("x-hop-private", "secret")
		out, _ := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "K"}, nil)
		t.Logf("connection: x-hop-private=%q", out.Get("x-hop-private")) // "secret"
	}
	// Anthropic Bearer plus inbound x-api-key.
	{
		b := Base{Provider: "anthropic"}
		req, _ := http.NewRequest(http.MethodPost, "/x", nil)
		req.Header.Set("x-api-key", "STALE")
		out, _ := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "OAUTH", Scheme: "bearer"}, nil)
		t.Logf("anthropic: authorization=%q x-api-key=%q", out.Get("authorization"), out.Get("x-api-key")) // both set
	}
	// Vertex duplicates.
	{
		b := Base{Provider: "vertex"}
		req, _ := http.NewRequest(http.MethodPost, "/x", nil)
		req.Header.Set("x-goog-user-project", "qp")
		req.Header.Set("x-vertex-ai-llm-request-type", "shared")
		out, _ := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "K"}, nil)
		t.Logf("vertex: %v %v", out.Values("x-goog-user-project"), out.Values("x-vertex-ai-llm-request-type")) // [qp qp] [shared shared]
	}
	// Gemini placeholder plus inbound quota project.
	{
		b := Base{Provider: "gemini"}
		req, _ := http.NewRequest(http.MethodPost, "/x", nil)
		req.Header.Set("x-goog-user-project", "caller-project")
		out, err := b.SanitizeAndMapHeaders(req.Context(), req, Credential{Key: "no-key-required", Scheme: "bearer"}, nil)
		t.Logf("gemini: err=%v x-goog-user-project=%q x-goog-api-key=%q", err, out.Get("x-goog-user-project"), out.Get("x-goog-api-key")) // header kept, no key
	}
}
```

</details>

JuliusBrussee:
**Needs changes** — the problem is real, but swapping the allowlist for a denylist forwards client credentials upstream. Probe on your branch vs. `main`:

| header | `main` | #1033 |
|---|---|---|
| `cookie: session=…` | dropped | **forwarded** |
| `x-api-key` (client's own key, openai / openai_compatible) | dropped | **forwarded** |
| `x-forwarded-for`, `cf-connecting-ip` | dropped | **forwarded** |
| `x-goog-user-project` on non-gemini providers | dropped | **forwarded** |

`main` leaks none of these; every one is new here. You caught the gemini case with `out.Del("x-goog-user-project")`, which is the same class — the denylist just can't enumerate it.

Note the outage you cite is already fixed: `forwardOpenCodeHeaders` landed on `main` via #1007, so this removes a working fix rather than repairing a broken path. The design question — allowlist that needs a release per provider header, vs. transparent forwarding that needs an exhaustive denylist — is the maintainer's call, and it's in the PR body for him. If it goes transparent, it needs `cookie`, client auth headers and forwarding/IP headers stripped, plus a test that a *new* sensitive header fails closed.

---

proxy/providers/adapter.go
AmirF194: Thanks for confirming the cookie case and catching the other forwarding gaps.

Issue #1008, opened 2026-09-06
caveman-proxy pins deleted ccr.db WAL inode → `caveman retrieve` intermittently returns "recovery handle not found" mid-session


Yashpal-0:
## Summary

The long-lived `caveman-proxy` daemon opens `~/.caveman/ccr.db` (WAL mode) and holds it for the entire session. When the DB file is replaced on disk (new inode — e.g. after a config change, update, or a checkpoint that swaps files), the proxy keeps its file descriptors pointing at the **deleted** inode. Because WAL checkpoints only fully publish when the last reader/writer closes, and the always-on proxy never closes its orphaned handle, `caveman retrieve` (via `caveman-mcp` or `caveman-engine retrieve`) intermittently reads a snapshot that is missing recently-written recovery handles → returns:

```
retrieve: ccr: recovery handle not found
```

A full opencode restart kills the proxy; the next launch opens the current file and retrieve "works again for a while" — matching the user-visible symptom exactly.

## Environment

- OS: Linux (Ubuntu)
- CLI: `@caveman-ai/cli` 1.3.3, binary release `bin-v1.1.6`
- Host: opencode (native integration via `caveman enable opencode`)
- Proxy: `/home/USER/.caveman/bin/caveman-proxy`, listening 127.0.0.1:8787

## Evidence

Proxy holds deleted WAL/SHM inodes while the on-disk `ccr.db` is a different inode:

```
$ ls -l /proc/<proxy-pid>/fd | grep ccr
lrwx------ 10 -> /home/USER/.caveman/ccr.db-wal (deleted)
lrwx------ 11 -> /home/USER/.caveman/ccr.db-shm (deleted)
lrwx------  9 -> /home/USER/.caveman/ccr.db        (old inode 89173)

$ ls -li ~/.caveman/ccr.db
1220765 -rw------- ... ccr.db     ← current inode, different from proxy's
```

Store schema confirms handles are durable in WAL-mode SQLite:

```
$ sqlite3 ~/.caveman/ccr.db ".tables"
recoveries  typed_objects
```

Retrieve works after proxy restart (fresh proxy opens current inode, no `(deleted)` markers):

```
$ caveman tools retrieve ccr_<handle> "query"   # succeeds
```

## Root cause

[...]

PR #1034, opened 2026-09-11
fix: preserve recovery-store locks and verify elisions

andrebrait:
## Problem

Native tool output could advertise a newly stored `ccr://` handle that a separate MCP reader immediately rejected as unknown. A long-lived producer still held an unlinked WAL/SHM pair while readers used newly created sidecars for the same database.

The cause is `secureSQLiteFile`: opening and closing an ordinary descriptor for the database or SHM file drops **all POSIX record locks held by that process on the inode**, including SQLite's locks. Both post-open permission checks and preparing a path already open elsewhere in the same process could therefore let a short-lived reader unlink a live writer's sidecars.

## Changes

Two separately reviewable, DCO-signed commits:

1. **Preserve SQLite locks while securing files.** Linux uses an inode-validated `O_PATH` descriptor and `fchmodat`, with a pinned procfs fallback for older kernels. Other Unix platforms use native no-follow chmod; Windows retains handle-based validation. Newly created files are closed before atomic, no-overwrite publication. Existing permission, symlink and inode checks remain in place. The shared helper also covers memory-store callers.
2. **Verify recovery before publishing Pi elisions.** The existing session recovery client verifies the advertised handle, exact byte length and SHA-256. A capability-gated `verify_only` mode on `caveman_retrieve` returns metadata without consuming the recovery delivery ledger. Missing, malformed, mismatched, old or unavailable recovery retains the original output, including non-text blocks.

No provider/OAuth routing changes, dependency additions, live-database migration code, or generated bundles are included.

## Reproduction and regression coverage

The storage regression keeps a writer open, opens and closes subprocess readers, then writes new objects and verifies both old and new originals from fresh readers. It also covers repeated same-process opens and direct path preparation. An independent reproduction failed with an unknown handle before this fix and recovered byte-identical content afterward.

Publication regressions cover missing/conflicting/malformed references, wrong proof hashes and lengths, older companions, failed/timed-out verification, preserved non-text output, and verification followed by the model's first full retrieval.

## Verification

On this PR branch, rebased onto current upstream `main`:

- `go build ./...` — passed
- `go vet ./...` — passed
- `go test ./...` — 68 packages passed; 4 have no tests
- `pnpm --filter @caveman-ai/pi test` — build/typecheck and all 38 tests passed

Additional verification of the same fixes before rebasing:

- Race-detector run of the cross-process locking regression passed.
- Forced `fchmodat2` failures using `strace -f -e inject=fchmodat2:error=ENOSYS`; the real older-Linux fallback passed both the independent reproduction and focused CCR permission/locking tests.
- CCR tests cross-compiled for Darwin/arm64 and Windows/amd64 (not runtime-tested locally).
- Deployed and exercised through installed OMP tools against a shared live producer: exact recovery at startup, after session replacement, and after consumer restart. An unavailable recovery store retained the original output. The formerly stranded handle became recoverable, and live SQLite integrity checks passed.

## Compatibility notes

Older MCP companions do not support non-consuming verification, so the updated Pi extension conservatively retains original output until the companion is upgraded. Availability verification is a publication-time check, not a promise against later deletion. This prevents new lock loss; it does not automatically repair already-diverged or corrupted databases.

### DIFF
--- a/engine/ccr/sqlite_file_security_linux.go
+++ b/engine/ccr/sqlite_file_security_linux.go
@@ -1,0 +1,39 @@
+//go:build linux && !js
+
+package ccr
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strconv"
+
+	"golang.org/x/sys/unix"
+)
+
+func chmodSQLiteFile(path string, info os.FileInfo) error {
+	// Closing an ordinary descriptor for the database or -shm drops ALL POSIX
+	// locks held by SQLite in this process. O_PATH pins the inode without opening
+	// it for I/O; closing this metadata-only descriptor does not release locks.
+	fd, err := unix.Open(path, unix.O_PATH|unix.O_NOFOLLOW|unix.O_CLOEXEC, 0)
+	if err != nil {
+		return err
+	}
+	file := os.NewFile(uintptr(fd), path)
+	defer file.Close()
+	opened, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	if !opened.Mode().IsRegular() || !os.SameFile(info, opened) {
+		return fmt.Errorf("file changed while opening")
+	}
+	if err := unix.Fchmodat(fd, "", 0o600, unix.AT_EMPTY_PATH); !errors.Is(err, unix.EOPNOTSUPP) && !errors.Is(err, unix.EINVAL) {
+		return err
+	}
+	// Kernels before fchmodat2/AT_EMPTY_PATH require procfs. This is the pinned
+	// descriptor's kernel-controlled link, NOT the swappable database pathname.
+	// chmod performs no open/close, so SQLite's locks remain intact. Fail closed
+	// if procfs is unavailable; never fall back to opening the database for I/O.
+	return unix.Chmod("/proc/self/fd/"+strconv.Itoa(fd), 0o600)
+}
--- a/engine/ccr/sqlite_file_security_unix.go
+++ b/engine/ccr/sqlite_file_security_unix.go
@@ -1,0 +1,25 @@
+//go:build !linux && !windows && !js
+
+package ccr
+
+import (
+	"fmt"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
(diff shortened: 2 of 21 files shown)


andrebrait:
Local validation is green: full Go build/vet/tests and all 38 Pi-extension tests, plus live cross-process recovery verification. GitHub has gated both CI and engine-ci as action_required for this first-time contribution. Could a maintainer please approve the workflow runs so the cross-platform checks can execute? Runs: https://github.com/JuliusBrussee/caveman/actions/runs/34613710421 and https://github.com/JuliusBrussee/caveman/actions/runs/34613710743

JuliusBrussee:
**First commit adopted** into #1015, cherry-picked with authorship preserved. I verified the diagnosis independently before reading your fix: securing `ccr.db` + `-wal` + `-shm` drops the whole POSIX lock set — each file alone is survivable, it is the combination that is fatal — and reverting `chmodSQLiteFile` to an ordinary descriptor reproduces #1008's `ccr: recovery handle not found` in all three of your scenarios. Thank you, this was a precise piece of work.

Added on top: your procfs fallback was never executed by any test, because every CI runner has `fchmodat2` — while being the branch that ships to RHEL 9, Debian 12 and Amazon Linux 2023. It now sits behind a seam with the full open/close-peer regression running through it.

**Second commit (Pi elision verification) not adopted** — capability negotiation in `packages/pi-extension` is a wider blast radius than a triage run should take, so it stays here for the maintainer.

One reviewer note I could not verify here: `createSQLiteFile` now depends on `os.Link`, and any error other than `EEXIST` propagates as a hard failure — so a `~/.caveman` on exFAT or a CIFS mount without hardlink support would fail store creation where the old `O_CREAT|O_EXCL` open succeeded. A fallback on non-`EEXIST` link errors would close that.

---

JuliusBrussee:
Closing as landed: the first commit (ccr.db POSIX lock set) was adopted into #1015 with your authorship preserved and merged to main on 2026-09-15. The second commit's elision verification was not taken — see the #1015 discussion for why; reopen a focused PR if you want to pursue it. Thank you.

Issue #743, opened 2026-07-24
Feature request: Full native support for Oh My Pi (OMP)


mikeethanh:
Please add full native OMP support, rather than only allowing users to manually copy the Caveman SKILL.md files.

OMP repository:

https://github.com/can1357/oh-my-pi

Ideally, Caveman's installer would detect OMP and install the integration using OMP's native plugin system.

Caveman already works conceptually in OMP because its core behavior is implemented through skills. Users can manually copy the skill directories into OMP's skill location.

However, this only provides part of the experience. Claude Code-specific hooks, session activation, status tracking, stats, commands, installer detection, and lifecycle integration do not work as a complete native OMP plugin.

Nongkers:
great

mikeethanh:
yes

JuliusBrussee:
Acceptance detail carried over from duplicate #16: native Oh My Pi support must keep Caveman active beyond the first prompt. A one-shot copied skill invocation is not sufficient; persistence across subsequent prompts is part of the integration’s acceptance criteria.

TomLucidor:
Would love to see this being pushed forward

PR #782, opened 2026-08-01
feat(install): add native OMP support

ousamabenyounes:
Fixes #743

## Summary
- Add `omp` provider detection and native installer routing (Oh My Pi).
- Prepare a managed OMP plugin package at `~/.omp/caveman-plugin` with `package.json#omp`, an extension entry, skills, commands, agents, and an activation rule, then delegate lifecycle to `omp plugin install` / `omp plugin uninstall`.
- Restore any previous prepared package if `omp plugin install` fails.
- Update installer/security docs (`INSTALL.md`, `README.md`, `SECURITY.md`) for the new user-facing OMP support.

## Rebased onto latest `main`
The branch was 26 commits behind `main` and `CONFLICTING`. Rebased onto `origin/main` and resolved the conflicts:
- The unified installer moved from `cli/install.js` to `bin/install.js` (now the source of truth; `cli/install.js` is a thin shim). The OMP code was **re-applied to `bin/install.js`** accordingly, and the OMP test now targets `bin/install.js`.
- One implementation change from the original: the local `copyDirRecursive` helper no longer exists in `bin/install.js`, so skill-dir copying uses stdlib `fs.cpSync(src, dest, { recursive: true })` (Node 18+).
- `INSTALL.md` / `README.md` / `SECURITY.md` conflicts resolved onto main's current structure, with the OMP entries re-added (install-matrix row, supported-agent list, install network-access note).

## Test Verification (RED → GREEN)
New OMP tests fail on the upstream base (feature absent):

```text
node --test tests/installer/omp.test.mjs   # against origin/main bin/install.js
1..3
# pass 0
# fail 3
```

Pass with the fix applied:

```text
node --test tests/installer/omp.test.mjs
1..3
# pass 3
# fail 0
```

Full installer suite — baseline preserved, +3 new OMP tests:

```text
# origin/main baseline
npm test → # tests 156 # pass 156 # fail 0

# fix/issue-743 (this branch)
npm test → # tests 159 # pass 159 # fail 0
```

Local CI replay (`npm test` + standalone `tests/test_*.js` + `python -m unittest discover -s tests`) is green on both the upstream baseline and this branch.

Note: the `typescript` and `windows` checks are red on `main` itself (a pre-existing `pnpm-lock.yaml` drift — `@caveman-ai/cli@^1.1.0`), unrelated to this change, which touches no `package.json` / lockfile / engine / packages files.

PR #1036, opened 2026-09-11
feat(omp): add native extension lifecycle and package entry

andrebrait:
## Summary
Add a separately typed native OMP entry to @caveman-ai/pi while retaining the Pi entry and shared gateway/native-hook protocol. Keep session-owned recovery, routing and prompt context in one runtime with serialized, generation-safe lifecycle transitions.

- OMP prompt arrays preserve existing blocks; session switch/tree/branch invalidates old asynchronous work and renews per-session recovery.
- Provider overrides are removed and selected models restored to direct endpoints on closure. OMP has no model_select extension event, so routing is revalidated at before_agent_start; Pi retains its existing model_select handling.
- Compress only successful supported read/bash results; preserve error, recovery, mutation/custom-tool and non-text output behavior.
- Build and publish separate Pi/OMP entries, explicit manifest metadata, optional host peers and a package license notice.

## Related work / scope
Related installer proposal:#782. This is runtime-package lifecycle support, not another installer change. Separate from my recovery-store PR#1034: no recovery-store publication or locking changes here, and no running Caveman deployment was changed.

## Verification
- npm test: TypeScript/build and39 protocol/provider/recovery/runtime/integration tests pass.
- npm run test:omp: built manifest-selected native entry lifecycle regression passes (disposable MCP processes and HTTP fixture).
- Real OMP loader/ExtensionRunner/AgentSession: native Core policy delivered on ordinary prompt, fresh steer and promoted follow-up using isolated disposable gateway/native-hook/MCP protocol fixtures and a scripted model response. No live daemon or external model service accessed.
- npm package preview includes both built entries, README and LICENSE.

Tested OMP18.1.17 integration SHA6aef0e8ad51b3bc5ea7a5f2a255c3d48e4c5af72. The version string alone does not establish that published builds include external-input/queued-startup fixes. OMP's current project-trust compatibility API is not a sandbox.

AI-assisted implementation, independently reviewed for contract conformance, lifecycle correctness/hostile input, test honesty and unnecessary complexity. Commits carry DCO sign-off.

### DIFF
--- a/LICENSING.md
+++ b/LICENSING.md
@@ -26,6 +26,7 @@ it as MIT adoption surface.
 | `packages/agent/` | MIT | Agent runtime, build compiler, Claude lane, framework adapters, and coding-agent API. |
 | `packages/create-caveman-agent/` | MIT | Zero-runtime-dependency Agent SDK initializer. |
 | `packages/cli/` | MIT | Funnel/on-ramp. Launches BSL binaries but does not contain engine code. |
+| `packages/pi-extension/` | MIT | Pi and OMP adoption surface; invokes the CLI and MCP binaries without embedding Engine code. |
 | `packages/sdk/typescript/` | MIT | Thin client and structural SDK surface. |
 | `packages/sdk/python/` | MIT | Thin client; distribution name is `caveman-sdk`. |
 | `packages/subagent-tax/` | MIT | Local zero-provider-call harness-prefix measurement tool. |
--- a/packages/pi-extension/LICENSE
+++ b/packages/pi-extension/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Julius Brussee
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/pi-extension/README.md
+++ b/packages/pi-extension/README.md
@@ -1,12 +1,13 @@
 # @caveman-ai/pi
 
-Native [Caveman](https://getcaveman.dev) extension for the [Pi coding agent](https://pi.dev).
+Native [Caveman](https://getcaveman.dev) extension for the [Pi coding agent](https://pi.dev)
+and [Oh My Pi (OMP)](https://omp.sh).
 
 One extension, four jobs:
 
 - **Proxy routing** — points the selected model's provider at your local Caveman
-  proxy (`/w/pi/...`) with a baseUrl-only override. Pi keeps owning auth, model
-  names, pricing, and `models.json`; nothing is copied or rewritten. The
+  proxy (`/w/pi/...`) with a baseUrl-only override. Your host keeps owning auth,
+  model names, pricing, and `models.json`; nothing is copied or rewritten. The
   extension routes a provider only when its base URL points at the upstream
   host of the proxy route (`api.anthropic.com`, `api.openai.com`,
   `generativelanguage.googleapis.com`, or `opencode.ai`), or at a named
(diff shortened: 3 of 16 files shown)


JuliusBrussee:
**Maintainer decision, not a code-quality verdict.** This adds a second published entry point to `@caveman-ai/pi` and a new officially-supported host. That is a distribution call — what a `pi-v*` tag ships — rather than a bug fix, and this routine triages feature work instead of adopting it.

Not reviewed for correctness: `packages/pi-extension` has no `node_modules` in this container, so I could not run `npm test` or `npm run test:omp` to check your results rather than take them on trust.

Two things worth the maintainer's eye: the `package-lock.json` churn is a dependency change inside a published package, and this overlaps #743 and #782, which propose the installer half of the same OMP support — worth deciding as one thing.

Separately: your #1034 has its first commit adopted into #1015 today, with details on that PR.

---

andrebrait:
## Re: #1067 (merged) — `installOmp` didn't actually deliver OMP support

I rebased this PR onto the current `main`, which now includes #1067 ("make OMP lifecycle safe and loadable"). While reconciling I looked closely at what `installOmp` (`bin/install.js`) actually ships for OMP users, and it's worth flagging before this PR lands.

**What #1067 hardened is real and valuable** — the ownership journal, atomic staging, symlink/foreign-registration refusal, and Windows Bun-shim launching are all solid safety work. But the thing being installed and hardened was a standalone ~20-line hand-rolled CommonJS stub (`ompExtensionSource()`), not this package. Compared to what every other native integration in this installer actually delivers, the gap is stark:

| Integration | Delivers |
|---|---|
| opencode (`src/plugins/opencode/plugin.js`, 277 lines + shared `caveman-config.js`/`caveman-parse.js`) | Dynamic mode tracking: slash commands, natural-language activation/deactivation, per-turn reinforcement |
| Pi / OMP via `@caveman-ai/pi` (this PR) | The actual product: proxy routing, `caveman_retrieve` recovery, Core injection, tool-output shrinking |
| OMP via `caveman install` pre-fix | A `session_start` UI status label and one static rule string re-appended to the system prompt every turn — no toggling, no compression, no recovery |

The old `installOmp` closed issue #743 (ruleset doesn't persist past the first prompt) but didn't give OMP users anything close to feature parity with opencode, let alone with what this package ships. A user running `caveman install --only omp` got a permanently-injected instruction block with zero compression mechanism behind it — arguably worse than not installing anything, since it spends tokens for no savings.

Also worth noting: the copied `commands/`, `agents/`, and `rules/` directories in the old generated plugin package were never wired into OMP's manifest (`omp.extensions`/`omp.skills` are the only fields OMP reads — confirmed against a real installed `agent-skills` plugin manifest) — they were inert files nobody discovered.

**What I changed** (new commit on this branch, `fix(install): install the real @caveman-ai/pi extension for OMP`):

`installOmp` now delegates to OMP's own plugin manager instead of hand-rolling a payload:
- **Running from a local checkout** (the common case — this installer already requires `src/hooks`/`agents`/`skills` siblings for every native provider) builds `packages/pi-extension` once (skipped once already built) and links it by path, so a checkout always installs exactly what's in that checkout — not whatever last shipped to npm.
- **Without a local checkout** it falls back to `omp plugin install @caveman-ai/pi` from the npm registry.
- Since `bin/install.js` and `packages/pi-extension` release on separate schedules, the registry path **verifies the installed version actually declares a loadable `omp.extensions` entry** before reporting success. Right now `@caveman-ai/pi@0.1.1` on npm predates this PR and has no `omp` field at all — I checked this empirically against the real `omp` CLI and confirmed the guard reports an honest, actionable failure instead of a false "installed" success. It'll start succeeding automatically once a release with OMP support ships.
- Uninstall checks `omp plugin list` first so a host that never had caveman's OMP plugin installed exits cleanly instead of erroring on "not installed" (this was a real regression I introduced mid-patch and caught via the existing `uninstall-safety` test suite — fixed before pushing).

This drops the local ownership journal entirely for the OMP path — there's no local payload left to own now that `omp plugin install` (registry or local-path) owns persistence itself, the same trust already extended to `claude plugin install` and `gemini extensions install`.

Verified end-to-end against the real `omp` CLI (not just the test shim): local-checkout install/uninstall round-trip, and the registry-fallback honest-failure path against the actually-published `@caveman-ai/pi@0.1.1`. Full installer suite (`node --test tests/installer/*.test.mjs tests/hooks/*.test.mjs`) passes: 386 passed, 13 skipped (platform-gated), 0 failed.

Happy to split this into its own PR against `main` if you'd rather land the installer fix independently of the OMP-native-extension work — let me know.

JuliusBrussee:
**Still a maintainer decision** — a second published entry point in `@caveman-ai/pi` plus a new officially-supported host is a distribution call, not a bug fix, so this routine does not adopt it. Re-flagged in #1103.

One factual claim from your new comment I could check, and it holds: `@caveman-ai/pi@0.1.1` is `latest` on npm and carries no `omp` field at all. So your registry-path guard would report an honest failure today rather than a false success, exactly as you describe.

Your direct question — whether to split the `installOmp` change into its own PR against `main` — is @JuliusBrussee's to answer, not mine, and I have not acted on it either way. Worth noting it is the more separable half: an installer fix against merged #1067 can be reviewed on its own terms, while the native-extension work still needs the distribution decision. I have not re-reviewed the installer commit for correctness; `packages/pi-extension` still has no `node_modules` in this container, so I could not run `npm run test:omp` to check your results rather than take them on trust.

---

palprateek:
Heads-up from the installer side: #782 — the installer proposal your Related-work section references — was superseded by #1067 and merged in ab5f6ce.

Two practical notes:

1. This PR touches `bin/install.js` and `tests/installer/omp.test.mjs`, both of which changed on main with #1067. Even where git merges cleanly, the #782-era package shape is no longer what main ships — worth updating to align with (or defer to) the landed layout: `package.json#omp` entry, extension entry, skills/commands/agents staging at `~/.omp/caveman-plugin`, delegated to `omp plugin install`.
2. The Related-work reference should point at #1067 now.

The runtime lifecycle core here (`src/omp.ts`, session-owned recovery, generation-safe transitions, provider-override restore) is untouched by #1067 and complementary — triage on #1067 explicitly recorded the two PRs as complementary, not competing. So this stays the natural follow-on once realigned. 👍

andrebrait-bux:
@palprateek thanks for bringing it to my attention. I did not update the PR description, but the code has already been updated against #1067.

andrebrait:
@palprateek thanks for bringing it to my attention. I did not update the PR description, but the code has already been updated against #1067.

PR #1038, opened 2026-09-12
fix(stats): accept the /caveman stats space-arg form

nathanhbsimmons:
**What:** `/caveman stats` and `/caveman:caveman stats` now route to the stats script, same as `/caveman-stats`.

**Why:** The space-arg form is what users reach for, and it was a silent no-op. `statsMatch` only matched the hyphenated command, so `/caveman stats` fell through to `parseModeChange`, where `stats` is not a `VALID_MODE`. Flag untouched, no stats block injected, prompt ran as an ordinary turn — no error, no output, nothing to indicate the command was even seen.

**Change:** one regex in `src/hooks/caveman-mode-tracker.js`:

```js
-const statsMatch = /^\/caveman(?::caveman)?-stats(?:\s+(.*))?$/.exec(prompt);
+const statsMatch = /^\/caveman(?::caveman)?(?:-stats|\s+stats)(?:\s+(.*))?$/.exec(prompt);
```

Tail flags (`--share`, `--all`, `--since`) work on the space-arg form too — same capture group.

**Tests:** three added to `tests/test_mode_tracker.py` — space-arg routing across all forms plus `--share`, the hyphenated forms still routing, and `/caveman ultra` still switching level (proving the widened alternation doesn't swallow real mode args). The suite uses an isolated `CLAUDE_CONFIG_DIR` with no session log, so they assert on routing (the stats branch's preamble), not on numbers; the numbers stay covered by `tests/test_caveman_stats.js`.

Also updated the mirrored copy of the regex in `tests/installer/slash-commands.test.mjs`, which is there specifically to track the live hook.

**Green:** `npm test` 279/279, `tests/test_mode_tracker.py` 37/37, `test_caveman_stats.js` 55/55, `test_mode_tracker_stdin.js` 18/18, `test_symlink_flag.js` 15/15, `test_caveman_init.js` 12/12. `tests.test_compress_safety` fails on my machine only — Python 3.9.6 can't parse `str | bytes` at def time in `compress.py:342`. Unrelated to this change, fails the same on a clean checkout.

No source-of-truth SKILL.md touched, so no CI resync needed.

https://claude.ai/code/session_018DgWoFStF2HEaW775zhGka

### DIFF
--- a/src/hooks/caveman-mode-tracker.js
+++ b/src/hooks/caveman-mode-tracker.js
@@ -197,7 +197,10 @@ function handle(raw) {
     // as additionalContext (#618), instructing the model to relay it
     // verbatim. The script reads the active session log, so we pass
     // transcript_path through when Claude Code provides it.
-    const statsMatch = /^\/caveman(?::caveman)?-stats(?:\s+(.*))?$/.exec(prompt);
+    // The space-arg form (/caveman stats) is accepted too: it used to fall
+    // through to parseModeChange, where "stats" is not a VALID_MODE, so the
+    // flag was left untouched and the prompt ran with no stats at all.
+    const statsMatch = /^\/caveman(?::caveman)?(?:-stats|\s+stats)(?:\s+(.*))?$/.exec(prompt);
     if (statsMatch) {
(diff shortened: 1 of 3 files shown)


nathanhbsimmons:
Opened by mistake

PR #1041, opened 2026-09-13
fix(hooks): repair win32 junction handling in the flag write/delete guards

Mehael:
## Symptom

Caveman mode drifts back to verbose prose mid-session on Windows machines where `$CLAUDE_CONFIG_DIR` (or `~/.claude`) is a directory junction to another drive — a common setup for anyone who keeps their home directory on a small system drive.

Two independent mechanisms both fail through the junction and compound into that symptom:

1. The per-turn reinforcement in `caveman-mode-tracker.js` is gated on a flag file that `safeWriteFlag` refuses to write, because its win32 symlink-parent check compares the resolved junction target against `os.homedir()` — the wrong proxy, since a junctioned config dir resolves to a realpath on a different drive with no home-directory prefix at all.
2. Even when the flag *can* be written elsewhere, `writeSessionMode`'s legacy-mirror cleanup deletes it with a bare `fs.unlinkSync`, which happily follows the same junction `safeWriteFlag` just refused to write through — so the one case where the write guard is doing its job is also the case where the delete path bypasses it and the file can never come back.

The SessionStart injection (`caveman-activate.js`) is not blocked by either bug — its own comment already explains why the per-turn reminder matters as much as it does:

> ```
> // 2. Emit full caveman ruleset, filtered to the active intensity level.
> //    The old 2-sentence summary was too weak — models drifted back to verbose
> //    mid-conversation, especially after context compression pruned it away.
> //    Full rules with examples anchor behavior much more reliably.
> ```
> — `src/hooks/caveman-activate.js:326-328`

That comment is upstream's own evidence for why this matters: the SessionStart ruleset is exactly what gets pruned by context compaction, and the per-turn reminder is the only thing that re-anchors behavior between SessionStart events. When the flag it depends on can never be written (or is destroyed and can't be recreated) on a junctioned config dir, that reinforcement silently stops firing for the rest of the session.

## Fixes (two commits, each independent)

1. **`fix(hooks): safeWriteFlag win32 guard uses writable-parent check, not home-prefix`** — replaces the `startsWith(homedir)` check with `fs.accessSync(dir, fs.constants.W_OK)`, testing what the guard actually cares about (can the current user write here) instead of a path shape that a junction breaks.
2. **`fix(hooks): guard the legacy flag delete path the same way its write path is guarded`** — adds `safeDeleteFlag`, mirroring `safeWriteFlag`'s junction-resolve + ownership/writable + no-symlink-target checks, and routes `writeSessionMode`'s legacy-mirror deletion through it instead of a bare `unlinkSync`. Before this, any prompt matching "stop caveman" / "normal mode" could destroy the flag through a junction even where the write path (correctly) refuses to touch it.

## A third defect that turned out to already be fixed

The original diagnosis (done against a plugin-cache install pinned at `655b7d9`) also found a SKILL.md-path off-by-one in `caveman-activate.js` — `path.join(__dirname, '..', 'skills', ...)` needing a second `'..'`. Checking current `main` before opening this PR: that one is already fixed, via #591 (`6d71b9e`, merged as `4520554`) — `skillPathCandidates()` in `caveman-config.js` now tries the correct two-levels-up path first and only falls back to the one-level-up path for the standalone-hook-install layout where it's actually correct. No action needed here; noted for the record since the branch this PR is based on is well past `655b7d9`.

## Testing

- `node --check` on the modified file.
- `node --test tests/hooks/caveman-config.test.mjs` — 24/25 pass; the one failure (`gcSessionStore refuses to follow a symlink out of the sessions dir`, `EPERM` creating a symlink without elevated privileges) reproduces identically on unmodified `main` in a separate worktree — pre-existing Windows-environment limitation, not a regression.
- `node tests/test_mode_tracker_stdin.js` — 17/18 pass; the one failure (`brevity trigger ("be brief") activates caveman at the default mode`, expects `full` got `ultra`) also reproduces identically on unmodified `main` in a separate worktree — a pre-existing environment leak (a repo-local `.caveman` config being picked up from an ancestor directory), not caused by this change.
- `src/hooks/checksums.sha256` regenerated for `caveman-config.js` in both commits, per the repo's own hook-integrity manifest.


src/hooks/caveman-config.js
```diff
@@ -190,12 +190,18 @@ function safeWriteFlag(flagPath, content) {
...
+          // The home-prefix check used to live here, comparing the resolved
+          // symlink target against os.homedir(). That is the wrong proxy on
+          // win32: a directory junction (e.g. ~/.claude junctioned to another
+          // drive, the same "legitimate symlinked config dir" case this branch
+          // exists to allow) resolves to a realpath that does not start with
+          // the home directory, so the write was refused through the very
+          // junction this code's own comments say it should tolerate. Test
+          // what actually matters instead: can the current user write there.
+          try {
+            fs.accessSync(realFlagDir, fs.constants.W_OK);
```
AmirF194, line 202: I ran this against `236e701` in a clean `node:20-slim` container, comparing it to `origin/main`.

`safeWriteFlag`'s win32 branch used to require the resolved symlink/junction target to sit under `os.homedir()` (added deliberately in a660db0, "verifies ownership on Unix (uid match) or under-home on Windows": the two checks were meant to be equivalent trust proxies, not just "is it writable"). This PR replaces that with `fs.accessSync(realFlagDir, fs.constants.W_OK)`, so any writable directory now passes regardless of where it is. `safeDeleteFlag` (new in this PR, line 314) has the same pattern.

Repro: junction `~/.claude` to a directory outside the home tree, with `process.getuid` unset (that's the actual gate this code uses for the win32 branch, not the real OS):

```
=== origin/main ===
wrote through the junction to a directory OUTSIDE home: false

=== 236e701 (this PR) ===
wrote through the junction to a directory OUTSIDE home: true
```

`tests/test_symlink_flag.js` still reports 15/15 passing on this branch, including the "no longer has blanket symlink parent refusal" audit test at the bottom of that file. It only checks that `realpathSync`, `getuid`, or `normalizedHome` appears somewhere in the source, and `getuid` is still there from the untouched Unix branch, so it can't see that the win32-specific containment was dropped.

If accepting any writable target is intentional to unblock cross-drive junctions, the win32 branch still needs some boundary narrower than "writable by me" (e.g. requiring it to be either under home or under the same drive/root the user explicitly configured), and the audit test should check the win32 code path specifically rather than a substring match over the whole file.

JuliusBrussee:
The inconsistency you found is real and I confirmed it by reading: on POSIX `safeWriteFlag` accepts a symlinked config dir anywhere the current user owns it, while the win32 branch requires the resolved target under `os.homedir()` — so a `~/.claude` junctioned to another drive is refused through the very junction that branch exists to tolerate. Good catch, and correctly located.

Not adopted as written, for one reason: `fs.accessSync(realFlagDir, W_OK)` on win32 is true of essentially anywhere an attacker could point the junction, so it does not narrow the hole the home-prefix check covered — it removes it. Windows exposes no `fs.stat` ownership to substitute, which makes this a design decision rather than a patch, so it is raised as an open item on #1015. Two candidates there: honor the junction only when `CLAUDE_CONFIG_DIR` explicitly names it, or keep the rule and document junctions as unsupported.

Two smaller notes. This runner cannot create a Windows junction, so no test here can fail first for the right reason — that is the other reason I stopped short. And the `safeDeleteFlag` argument is thinner than stated: `unlinkSync` never follows a symlink at the target, so it is the parent-directory handling that carries the weight.

---

AmirF194:
Thanks for confirming the finding, and for tracking the Windows containment question on #1015 that makes sense, since there is no ownership check to fall back on there.

JuliusBrussee:
**Adopted into #1055** — both commits cherry-picked, authorship preserved. Thanks, this was a good catch and the write-up made it easy to verify.

I reproduced both defects before adopting: the junction case by dropping `process.getuid` (Node reports a win32 junction as `isSymbolicLink()`, so the no-getuid branch is reachable on a POSIX runner), and the delete asymmetry by staging a foreign-owned symlinked parent. Added the two regression tests the PR was missing.

One sibling you missed: `appendFlag` carries a byte-identical copy of the same home-prefix guard, so the lifetime stats log (`.caveman-history.jsonl`) stayed silently dead on exactly the machines this fixes. #1055 fixes that too.

Flagged for the maintainer rather than decided by me: `accessSync(W_OK)` is weak on Windows. I kept it because the check it replaces wasn't a real boundary either — anyone who can plant a junction at `~/.claude` can point it *inside* the home dir and pass `startsWith(homedir)`.

---

JuliusBrussee:
**Already landed — your fix is on `main`.** It came in through the 2026-09-15 triage merge (`ed37ab1`, #1055): `src/hooks/caveman-config.js` now carries the win32 writable-parent check in place of the home-prefix comparison, with your reasoning in the comment, plus `safeDeleteFlag` and the `writeSessionMode` call site that uses it.

Nothing left to change here. Flagged in #1082 so the maintainer can close it; I don't close PRs.

Worth recording why the write/delete asymmetry mattered, since it is the less obvious half: without `safeDeleteFlag`, a caller could delete straight through a junction that `safeWriteFlag` would refuse to write through, so the flag could be destroyed but never recreated on that machine. Good catch.

---

JuliusBrussee:
Superseded — this already landed on main via #1055.

Current `src/hooks/caveman-config.js` has both halves of this PR: `safeDeleteFlag` is defined and exported, `writeSessionMode` clears the legacy mirror through it, and the win32 branch of `safeWriteFlag` uses the `fs.accessSync(realFlagDir, W_OK)` check in place of the old home-prefix comparison — with your junction reasoning preserved in the comment.

The remaining gap your work exposed was `clearSessionPrev`, fixed today in #1089 via #1086.

Thanks — the "home-prefix is the wrong proxy on win32" diagnosis is the part that stuck.

---

JuliusBrussee:
**Landed via #1055.** The win32 junction handling is on `main`: both `safeWriteFlag` and `safeDeleteFlag` take the junction branch (writable-parent check on win32, instead of the home-prefix check that rejected a legitimately junctioned config dir).

I did not diff your branch against the landed version line by line. If yours still covers a case the landed one misses, point at it and I'll pick it up next run.

Thanks for the patch — the diagnosis was right, and it shipped.

---

PR #1042, opened 2026-09-13
fix(install): disable native agent integrations on uninstall

### LINKED ISSUE #1040: npx --uninstall leaves the ANTHROPIC_BASE_URL route in ~/.claude/settings.json, so Remote Control stays broken after uninstall
**What happen**

`caveman setup --install` writes a persistent route into the user-scoped Claude Code settings:

```json
"env": {
  "ANTHROPIC_BASE_URL": "http://127.0.0.1:8787/w/claude",
  "_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL": "1"
}
```

The integration record at `~/.caveman/integrations/claude.json` owns exactly this: `owned.route = http://127.0.0.1:8787/w/claude`, `owned.assume_first_party = "1"`, `previous_route = null`.

As already documented in #947 and in `docs/technical/agent-wrapping.md` (section "Claude Code Remote Control"), this route makes Claude Code Remote Control unavailable. To get Remote Control back I uninstalled Caveman with the command both uninstall sections document:

```bash
npx -y github:JuliusBrussee/caveman -- --uninstall
```

After that, both env keys were still in `~/.claude/settings.json`, and Remote Control stayed broken. It only came back after I removed `ANTHROPIC_BASE_URL` and `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL` from `settings.json` by hand.

**Why it happens (docs)**

- `INSTALL.md#uninstall` lists what `--uninstall` removes (hook entries, hook files, plugins, mode state) and what it does not remove. The env route written by `caveman setup --install` / `caveman enable claude` is in neither list.
- `docs/technical/install-and-update.md#uninstall` gives the same `npx … --uninstall` plus `npm uninstall -g @caveman-ai/cli`, and neither touches `settings.json` env.
- The only place that mentions `caveman disable claude` is the Remote Control section of `agent-wrapping.md`, which a user following the uninstall docs will not read.

So a user who uninstalls exactly as documented keeps a dead (or still-proxied) `ANTHROPIC_BASE_URL`, and keeps the Remote Control breakage from #947, with nothing in the uninstall output pointing at the cause.

**Expected**

One of:

1. `--uninstall` also removes the CLI-owned route. The integration record already knows the owned keys and that `previous_route` was `null`, so there is nothing to restore, only two keys to delete.
2. Or the uninstall sections in `INSTALL.md` and `install-and-update.md` say to run `caveman disable claude` (or `caveman setup --agent-native claude --remove`) before `--uninstall` / `npm uninstall -g`, and `--uninstall` warns when it finds a Caveman route still in `settings.json`.

Separately, it would help if the install docs (not only `agent-wrapping.md`) mention up front that native routing disables Claude Code Remote Control.

**Before/after example**

```
Input:  npm install -g @caveman-ai/cli && caveman setup --install
        npx -y github:JuliusBrussee/caveman -- --uninstall
        claude remote-control
Got:    ~/.claude/settings.json env still contains ANTHROPIC_BASE_URL=http://127.0.0.1:8787/w/claude
        and _CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL=1; Remote Control stays unavailable (#947)
Want:   both keys removed (previous_route was null), Remote Control available again
```

I did not test whether `caveman disable claude` removes the keys correctly; this report is about the documented uninstall path.

**Platform**
- [x] Claude Code
- [ ] Codex
- [ ] Other: ___

**Version / install method**

- `@caveman-ai/cli` 1.3.3 (`binary_release: bin-v1.1.6`), installed via `npm install -g @caveman-ai/cli`, then `caveman setup --install`
- Integration `pack_version` 2.2.0
- Claude Code 2.1.269 at install time, 2.1.270 now
- macOS

AmirF194:
`caveman enable claude` writes ANTHROPIC_BASE_URL (and the MCP entry) straight into settings.json and journals the previous value at ~/.caveman/integrations/claude.json so `caveman disable` can put it back. uninstall() never reads that journal, only the hook-based install state, so those keys survive `--uninstall` verbatim and Remote Control stays routed at a proxy that's gone.

Rather than re-implementing the journal restore in JS, uninstall() now shells out to `caveman disable --all` when the CLI is on PATH, before the hook cleanup runs. That restores every native agent (claude, codex, gemini, ...) in one call through the same code `caveman disable` already uses, and it's a no-op when nothing was ever enabled natively.

One gap this doesn't close: if `caveman` itself was already removed from PATH before `--uninstall` runs, there's nothing to shell out to and the journal is left alone. Same blind spot `claude plugin uninstall`/`gemini extensions uninstall` already have for their own missing binaries.

Added a test that stubs `caveman` and asserts uninstall calls `disable --all` (and that install itself never touches it). It fails on main, since nothing invokes the binary today. Ran the full `npm test` suite before and after: same 16 pre-existing failures on unrelated profile-scope tests either way, 261 passing plus the 2 new ones.

Fixes #1040


JuliusBrussee:
Adopted into #1015 — both commits cherry-picked with authorship preserved, including the Windows node-shim fix. Delegating to `caveman disable --all` instead of re-deriving journal restoration inside `bin/install.js` is the right call, and the comment saying so is worth keeping.

Verified the delegation actually does the job rather than trusting the recording stub: `enable claude` writes the route and journal, `disable claude` removes the journal and restores `settings.json` to its pre-install state.

Extended with the branch the guard creates: `hasCmd('caveman')` means a user who ran `npm uninstall -g` first, or whose `disable --all` failed, keeps the route silently. `--uninstall` now names any agent still journaled and the command to run. A surviving journal is exact evidence, so it cannot false-positive on a user's own `ANTHROPIC_BASE_URL`. Plus the docs half #1040 asked for.

---

AmirF194:
The hasCmd('caveman') gap is exactly the one I'd left unhandled. Naming the surviving journal instead of guessing at PATH state is the right fix, and it stays correct even against a user's own ANTHROPIC_BASE_URL.

JuliusBrussee:

AmirF194:
And this one too, thanks for folding all three into #1015.

PR #1043, opened 2026-09-13
fix: move native-core.md out of pi's skills scan

tkoyama010:
## Problem

When caveman is installed as a [pi](https://github.com/badlogic/pi-mono) package, pi collects every **direct `.md` child of the package's `skills/` directory** as a skill candidate (`SKILL.md` files from subdirectories, plus loose root-level `.md` files). `skills/native-core.md` is the native pack build source — consumed by `skills/compile.mjs` via `native_pack.core_source`, not a skill, and has no frontmatter — so pi warns `description is required` for it under the **[Skill conflicts]** banner on every launch.

## Fix

- Move `skills/native-core.md` → `skills/native/native-core.md`. pi only reads `SKILL.md` files from subdirectories, so the subdirectory is invisible to it.
- Relax the `native_pack.core_source` validator in `skills/compile.mjs` to allow subdirectory paths (still rejects `..` and absolute paths).
- Regenerate the compiled packs.

Instructions are embedded in the generated artifacts (`native-pack.generated.json`, `native-pack.generated.ts`, `skills/generated/*/pack.json`), so no consumer reads the source file at runtime.

## Verification

- `node skills/compile.mjs` passes; regenerated artifacts differ only in the `source` field.
- pi skill loader on a machine with this package installed: **34 skills, 0 diagnostics** (previously 1 warning).
- `packages/cli` native-enable test suite: same result before and after this change (pre-existing environment failures unrelated to this PR).


JuliusBrussee:
Needs info before this can be adopted — no linked issue, no test, and no symptom described, so I could not reproduce anything to verify the move against.

What I checked: `skills/native-core.md` is the native pack's `core_source` (`skills/registry.json`), not a skill. It is not a `skills/*/SKILL.md`, so Claude Code's wholesale discovery does not pick it up, and `pi` is not among `native_pack.targets` (claude, codex, hermes, gemini, opencode, aider). So I could not find the scan the title refers to.

The move may well be correct — `commands/*.md` shadowing same-named skills was a real bug of exactly this family, and it is documented in CLAUDE.md. But this routine does not ship a fix it cannot reproduce. Could you say which scan picks the file up, what the user-visible symptom is, and ideally add a test that fails before the move? Happy to pick it up next run with that.

---

JuliusBrussee:
**Adopted** — both commits cherry-picked into #1089 with your authorship preserved.

Verified rather than assumed: I restored the loose `skills/native-core.md` and confirmed your new `verify_repo.py` check fails on it with the intended message, then confirmed `node skills/compile.mjs` regenerates the packs byte-identically after the move. The only surviving mentions of the old path are the cosmetic `// GENERATED by … from native-core.md` header strings, which no consumer resolves at runtime.

Two things I appreciated: shipping the regression guard with the fix instead of leaving it to be written later, and deriving `SKILLS` from the directory rather than pinning a stale list. The `includes("/")` → `startsWith("/")` relaxation still rejects `..` and absolute paths, so the validator did not lose anything.

#1089 is currently red on an unrelated pre-existing check (`profiles`) that is blocked on a maintainer reply — nothing to do with your change.

---

PR #1044, opened 2026-09-14
evals: isolate runs from local config, per-language prompts, pt-BR snapshot

edubraqd:
## What

Three changes to `evals/`, all MIT-side, no CI impact (no workflow reads the snapshot).

1. **Isolation.** `llm_run.py` called `claude -p` with the user's settings, hooks, CLAUDE.md, skills and MCP servers loaded. Measured with a probe prompt on my machine: `~/.claude/CLAUDE.md` was present in the *baseline* arm. Anyone with a "be terse" rule in their CLAUDE.md (or a hook/plugin that injects style rules) was measuring that in every arm. Every call now passes `--setting-sources "" --disable-slash-commands --tools "" --strict-mcp-config --no-session-persistence`. `--bare` would be simpler but it skips keychain reads and cannot authenticate.

2. **`--system-prompt-file`.** On Windows `shutil.which("claude")` resolves to `claude.CMD`; cmd.exe re-parses the multi-line SKILL.md argument (quotes, backticks, `&`) and the skill arm dies with `Input must be provided either through stdin or as a prompt argument when using --print`. The baseline and terse arms pass because their prompts are one short line, so the failure only shows up on the arm that matters. A temp file sidesteps every shell.

3. **Per-language prompts.** `CAVEMAN_EVAL_LANG=pt` reads `prompts/pt.txt`, uses a translated terse control (`Responda de forma concisa.`) and writes `snapshots/results.pt.json`. `CAVEMAN_EVAL_SKILLS=caveman` limits the run to named skills (20 skills × 10 prompts is a lot of calls for a language snapshot). English keeps its file names, so nothing that reads `results.json` changes.

## Why a Portuguese snapshot

SKILL.md promises "compress the style, not the language", and the rules are written against English function words (a/an/the, just/really/basically). Whether that transfers to a language with gendered articles and a different filler vocabulary was unmeasured; #812 reports language drift. This gives it a number.

## First pt-BR result (haiku-4-5, 10 prompts, single run)

| Arm | Tokens | vs baseline |
|---|---|---|
| baseline (no system prompt) | 3192 | — |
| terse control (`Responda de forma concisa.`) | 3096 | +3% |
| caveman | 2460 | +23% |

caveman on top of the terse control: **median +19%**, mean +19%, stdev 34%, range −43%..+69%. Same noise profile as the English snapshot: one prompt got longer, one got 69% shorter.

Language drift: zero English function words (`the/is/are/because/with/your`) in any of the 30 Portuguese replies, all three arms. With the current SKILL.md and haiku, the PT/ES leak in #812 did not reproduce.

## How to reproduce

```bash
CAVEMAN_EVAL_LANG=pt CAVEMAN_EVAL_MODEL=claude-haiku-4-5 CAVEMAN_EVAL_SKILLS=caveman uv run python evals/llm_run.py
CAVEMAN_EVAL_LANG=pt uv run --with tiktoken python evals/measure.py
```

The English snapshot was **not** regenerated; it still carries `caveman-cn`/`caveman-es` arms that no longer exist as skills. Happy to refresh it in a follow-up if wanted, but that is a separate diff.


JuliusBrussee:
**Looks right; recommending the maintainer merge it.** Evals-only, MIT side, no workflow reads the snapshot, and the three changes are each independently justified.

Verified rather than trusted: every isolation flag exists on the current CLI (`--setting-sources`, `--disable-slash-commands`, `--tools`, `--strict-mcp-config`, `--no-session-persistence`, `--system-prompt-file`), and the pt snapshot's metadata (haiku-4-5, CLI 2.1.251, 10 prompts, translated terse control) matches the run described. The Windows `claude.CMD` re-parse is the same class of bug this repo has hit in the installer, so the temp-file route is the right shape.

One small thing, non-blocking: `TERSE_PREFIXES[LANG]` raises a bare `KeyError` for a language with a prompt file but no terse control; a one-line `SystemExit` naming the missing entry would match how `CAVEMAN_EVAL_SKILLS` already fails. Leaving the stale English snapshot for a follow-up is the correct call — it is a known item and needs its own run.

edubraqd:
**Done in cde4852: a language with `prompts/<lang>.txt` but no `TERSE_PREFIXES` entry now exits with a named error instead of a bare `KeyError`.**

```
CAVEMAN_EVAL_LANG=es: no terse control in TERSE_PREFIXES; add one next to prompts/es.txt
```

Same shape as the `CAVEMAN_EVAL_SKILLS` failure, and it fires at import, before the first `claude -p` call, so a typo in the env var costs zero API budget. Five lines, nothing else changed; both snapshots are untouched, so the metadata in `results.pt.json` still describes the run that produced it.

Checked both branches: `pt` resolves to the translated control, a `LANG` with no entry hits the `SystemExit`. Nothing else imports `llm_run.py` (`measure.py` and `benchmarks/run.py` only mention it in comments), so CI sees exactly what it saw before.

The English snapshot refresh stays a separate PR. It needs its own run under the same isolation flags so `results.json` and `results.pt.json` are comparable, and an LLM re-run doesn't belong in the same diff as three code changes already verified.

edubraqd:
@JuliusBrussee status, three days after the triage bot's "recommending merge": the one nit is fixed in cde4852 (named `SystemExit` instead of a bare `KeyError`), and the branch still applies cleanly on `main`. Both workflows are waiting for a maintainer to approve the run, so CI has not executed yet.

Merge as is, or fold it into a daily triage PR like the others, whichever is less work for you. The English `results.json` refresh is a separate follow-up once this lands.

PR #1046, opened 2026-09-14
Fix codex native/wrap route missing /v1; auto-repair stale auth-mode routing

### LINKED ISSUE #1045: Codex integration (pack v2.2.0): base_url missing /v1, and chatgpt-auth tokens sent to platform API cause 401 missing scope api.responses.write
**Versions:** caveman CLI 1.3.3, binary_release bin-v1.1.6, caveman-proxy running dev build on 127.0.0.1:8787. codex-cli 0.154.0. Codex integration pack_version 2.2.0.

**Bug 1 — wrong base_url written by the codex integration installer**

`caveman <agent> codex` (or whatever installs the integration) writes this to `~/.codex/config.toml`:

```toml
[model_providers.caveman]
base_url = "http://127.0.0.1:8787/w/codex"
wire_api = "responses"
```

Codex's OpenAI-Responses client appends `/responses` directly to `base_url`, producing `http://127.0.0.1:8787/w/codex/responses`. The proxy doesn't recognize that path and 404s:

```
{"error":{"type":"cave_gateway_error","code":"cave_route_not_found","message":"Proxy path is not recognized."}}
```

Confirmed via direct curl: `/w/codex/responses` → 404 `cave_route_not_found`, `/w/codex/v1/responses` → routes correctly (gets a real upstream response). So `base_url` should be written as `http://127.0.0.1:8787/w/codex/v1` (matching the `/v1` the client expects to already be present, same convention as OpenAI's own base_url).

**Bug 2 — chatgpt-auth (OAuth) sessions get forwarded to the platform API and rejected**

After fixing base_url locally, requests reach the proxy and get forwarded upstream, but fail with:

```
401 Unauthorized: You have insufficient permissions for this operation. Missing scopes: api.responses.write.
url: http://127.0.0.1:8787/w/codex/v1/responses
```

`~/.codex/auth.json` has `auth_mode: "chatgpt"` (ChatGPT-plan OAuth login via `codex login`, tokens under `.tokens.access_token`, no `OPENAI_API_KEY`). It looks like the proxy is forwarding this ChatGPT session token straight to `api.openai.com/v1/responses` (the platform API), which requires `api.responses.write` — a scope ChatGPT-plan OAuth tokens don't carry (the codex login OAuth flow only requests `openid profile email offline_access api.connectors.read api.connectors.invoke`).

Reproduced with a fresh `codex login` (ruled out stale token) — same error both times.

**Workaround confirmed:** disabling the caveman `model_provider` override in `~/.codex/config.toml` (so Codex talks to OpenAI directly, provider `openai`) works immediately with the same ChatGPT login — confirming the OAuth session itself is fine, and the issue is specifically in how the proxy forwards `auth_mode: chatgpt` sessions.

**Suggested fix:** when `auth_mode` is `chatgpt`, the proxy should forward to the ChatGPT backend endpoint (e.g. `chatgpt.com/backend-api/codex/responses`) instead of `api.openai.com/v1/responses`, same as native Codex does — rather than treating the OAuth access_token as if it were a platform API key.

**Bonus (non-blocking):** `GET /w/codex/v1/models` also 404s with `Proxy path is not recognized` — Codex logs this as an ERROR on every `codex exec` (background model-list refresh), doesn't block the session but is noisy.

g0GobliN:
Fixes #1045.

**Bug 1: api-key route missing `/v1`**

`codexCavemanProviderToml` writes `base_url = ".../w/codex"` for api-key-mode codex sessions (both native install and ephemeral `wrap`). Codex's Responses client appends `/responses` directly onto that, giving `.../w/codex/responses` — but the proxy's own route table (`proxy/providers/openai/openai.go`) only matches `/v1/responses` (or `/openai/v1/responses`), never a bare `/responses`. Every api-key codex request 404'd with `cave_route_not_found` before reaching OpenAI.

Fixed by appending `/v1`, same as the existing `/w/aider/openai/v1` convention. Updated the two other places that construct/compare this route (`codexNativeMutations`, `nativeIntegrationStatus`) so install and `doctor` drift-checking stay consistent, plus the one test that hardcoded the old string.

**Bug 2: native install goes stale when auth mode changes**

`detectCodexWrapAuthMode()` is checked once, at install time, and baked into `config.toml`. If someone runs `codex login` afterward (switching from api-key to ChatGPT-subscription auth, or vice versa), nothing re-checks it — `caveman doctor codex` correctly reports `degraded`, but only if you think to run it. Added a small self-repair in the codex `SessionStart` hook: if status comes back degraded, it just runs the same repair `doctor codex --fix` does, instead of leaving a silent 401 waiting for the next session.

**Testing**

- `tsc --noEmit` clean.
- `node --test tests/codex-subscription-wrap.runtime.mjs tests/native-enable.runtime.mjs`: 38/39 pass — the 1 failure (`doctor reports Codex routing degraded when auth lane changes`) reproduces identically on unpatched `main`, so it's pre-existing and unrelated to this change.
- Verified end-to-end against a real Codex CLI session (both api-key-shaped and ChatGPT-subscription-shaped `auth.json`), confirming the route the proxy now receives actually resolves instead of 404ing.


g0GobliN:
Pushed a follow-up commit addressing two gaps in the auto-repair part:

- Narrowed the SessionStart self-repair from any `degraded` state to specifically `components.routing === false`. `degraded` also covers pack-version drift, missing hooks, MCP recovery being down, etc. — none of those should get a silent config rewrite just because a session started; they still only surface through `caveman doctor codex` as before.
- Added test coverage for both drift directions (api-key → subscription and back), plus a case proving an unrelated degraded reason (simulated stale `pack_version` in the install journal) leaves `config.toml` untouched.
- Fixed an env-hygiene bug in the `native-enable.runtime.mjs` fixture: it spread `...process.env` unfiltered, and `detectCodexWrapAuthMode()` falls back to reading `OPENAI_API_KEY` from the process env. Anyone running the suite with a real key exported (normal for actual Codex use) got every codex fixture silently forced into api-key mode regardless of what the fixture's `auth.json` said — which is exactly what made the existing "doctor reports Codex routing degraded when auth lane changes" test fail locally until I tracked it down. Not a bug in this PR's logic, but a real false-negative risk for anyone else running this file, so fixed it alongside.

Full suite (`tests/*.runtime.mjs`) with `OPENAI_API_KEY` unset: 919/951 pass, 28 skipped, same 4 pre-existing failures as on unpatched `main` (unrelated to codex — claude native-launch, kilocode routing, raw wrap base URL union).

g0GobliN:
Closing for now — still finishing verification (CI, broader test coverage) before this is ready for review. Will reopen once it's solid.

PR #1047, opened 2026-09-14
codex: self-heal stale auth-lane routing after login

g0GobliN:
Refs #1045.

Rebased onto current `main` after #1055. The original `/v1` route-fix half is intentionally dropped here because `main` already carries `CODEX_API_KEY_ROUTE` via #1015.

What remains in this PR:

- On Codex `SessionStart`, if the native integration is degraded and its routing component is unhealthy, run the same repair path as `caveman doctor codex --fix`. This repairs stale `config.toml` routing after `codex login` switches between API-key and ChatGPT/subscription auth. Degradation with healthy routing is left alone, and the hook stays best-effort/fail-open.
- Isolate the native-enable fixture from a developer shell's real `OPENAI_API_KEY`, so fixture auth-mode detection comes only from the fixture's own `auth.json`.
- Add regression coverage for both auth-switch directions plus a guard proving unrelated degradation does not rewrite `config.toml`.

Validation on the rebased patch:

- Typecheck/build clean.
- `native-enable.runtime.mjs`: 44/44 pass.
- Full `packages/cli` suite: 1046 pass, with one `kilo-wrap` failure reproduced unchanged on clean `main`.
- Root hook/mode-tracker/verify suites green.
- Mutation check on pristine `main`: both auth-switch regression tests fail without the implementation at the stale-route assertion, while the unrelated-degradation guard still passes.


JuliusBrussee:
Route half adopted into #1015, re-implemented through a shared constant so the provider TOML, the install journal and the doctor's expected route cannot drift — you are credited as co-author. Your diagnosis of both halves of #1045 was correct.

Self-heal half **not** adopted, and this is a maintainer decision rather than a request for changes. The problem it solves is real: `codex login` flips the auth lane and `config.toml` keeps the old one. But it would have a SessionStart hook take the integration lock and restore-then-reapply the user's `~/.codex/config.toml` unasked, on a hook budget. `caveman doctor codex --fix` already repairs it and `doctor codex` already reports it (`routing: false`, with the repair command), so this buys convenience, not capability. Raised as an open item on #1015; if it lands, a hook that *warns* rather than writes is the narrower version.

Worth adding: the conformance harness was green through all of this because its Codex stub appended `/v1/responses` and supplied the segment your config lacked. That is fixed too, so the harness now reproduces your 404.

---

JuliusBrussee:
**Not a duplicate of #1052 — recommending this one for review next.**

Checked both: #1052 fixes "proxy never started after `enable`" (#1051), yours fixes "api-key lane routes to `/w/codex` with no `/v1`, so the proxy 404s with `cave_route_not_found`" (#1045). Different root causes, complementary, both real.

Heads-up on sequencing: you and #1052 edit the same regions of `packages/cli/src/index.ts` and `native-enable.runtime.mjs`, so whichever lands second needs a rebase. #1055 adopted #1052 today, so that's likely you — sorry, no reflection on this PR's quality.

Separately: your fixture fix for the inherited `OPENAI_API_KEY` is the same defect class I hit independently with `ANTHROPIC_BASE_URL` leaking into the `agent-shortcut` fixtures (two tests fail on any machine with a gateway configured). Fixed there in #1055; good find on your side.

---

JuliusBrussee:
**Correction to my comment above — half of this has already landed on `main`.**

My earlier reply was written against a stale checkout. Re-checked after merging current `main`:

- **The `/v1` route fix is already in.** `main` now has `const CODEX_API_KEY_ROUTE = "/w/codex/v1";` (`index.ts:6277`), landed via `ca9d420` / `8bc01d6`. So the `cave_route_not_found` half of this PR is fixed — and, usefully, it was hoisted into a single named constant rather than the three separate literals this PR patches, so that part will conflict and should just be dropped on rebase.
- **The SessionStart self-heal is still novel.** Nothing on `main` calls `repairNativeAgent("codex")` on routing drift, so the "`codex login` silently leaves config.toml stale" half is unfixed and still worth landing.
- Your fixture fix for the inherited `OPENAI_API_KEY` also still applies.

So: still worth reviewing, but as a smaller PR than it is today. Sorry for the misdirection in my previous comment.

---

g0GobliN:
Rebased onto current `main` and narrowed this per your latest comment.

- dropped the `/v1` route half since it already landed via #1015
- kept the SessionStart self-heal
- kept the `OPENAI_API_KEY` fixture isolation
- kept regression coverage for both auth-switch directions and unrelated degradation

The branch is now a single commit on top of post-#1055 `main`.

JuliusBrussee:
**Reviewed, no action needed from triage — this is ready for the maintainer's review.**

Deliberately *not* absorbed into today's triage PR (#1070). It is already rebased onto current `main` (`ed37ab1`, the same base), the scope is correctly narrowed now that `CODEX_API_KEY_ROUTE` landed via #1015, and you are actively iterating on it. Pulling the commits into a triage PR would only duplicate your work and fragment the review.

Two notes from reading it rather than running it: dropping the `/v1` half after #1015 is the right call, and gating the repair on *routing* being unhealthy rather than on degradation generally is the distinction that keeps this from rewriting `config.toml` on unrelated problems — the guard test for that is the one I would have asked for.

One thing to look at: `mergeable_state` currently reports `unstable`, so a check is failing or still pending on `bdc09d1`. Worth a glance before it goes up for review.

---

Issue #1037, opened 2026-09-12
Codex re-prompts for already-approved commands after Caveman shrink rewrite


Alexey-Matjuk:
**What happen**

With Caveman's Codex integration enabled, a command already approved by the user is rewritten by `shrink-hook`. Codex sees a different command and asks for another approval.

Relevant `config.toml`:

```toml
model_provider = "caveman"
sandbox_mode = "workspace-write"
approval_policy = "on-request"

[model_providers.caveman]
base_url = "http://127.0.0.1:8787/chatgpt"
wire_api = "responses"
requires_openai_auth = true
```

Relevant `hooks.json`:

```json
{
  "hooks": {
    "PreToolUse": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "<caveman-bin> shrink-hook"
          }
        ]
      }
    ]
  }
}
```

Existing Codex approval rule:

```python
prefix_rule(pattern=["git", "add"], decision="allow")
```

**Expected**

A command already covered by a saved Codex approval rule should not ask for approval again when Caveman shrinking is active.

**Before/after example**
```
Input: git add -A
Got:   caveman shrink -- git add -A
       Codex asks for approval again.
Want:  The existing approval for git add applies without another prompt.
```

**Platform**
- [ ] Claude Code
- [x] Codex
- [ ] Other: ___

**Version / install method**

- Caveman CLI: `1.3.3`
- Caveman native binary: `bin-v1.1.6`
- Native pack: `2.2.0`
- Codex: `0.153.4`
- Installation: global npm installation of `@caveman-ai/cli`
- OS: macOS

JuliusBrussee:
**Reproduced and fixed** in #1015 (awaiting maintainer review). Your paste was exact — here it is from the hook itself:

```
$ echo '{"tool_name":"exec_command","tool_input":{"command":"git add -A"}}' | caveman shrink-hook
… "updatedInput":{"command":"'/opt/node22/bin/node' '/…/index.js' shrink -- git add -A"}
```

[...]

PR #1048, opened 2026-09-14
fix: preserve Windows commands when shrink hook cannot identify the shell

IdealX-dev:
Windows coding hosts can report PowerShell commands as `tool_name: Bash` with only a `command` field. The shrink hook currently prepends a quoted executable path using Bash syntax; PowerShell treats it as a string and fails with `Unexpected token shrink`, blocking ordinary commands such as `git diff`.

Preserve commands on Windows when the hook cannot identify the execution shell. POSIX rewriting is unchanged, and explicit manual compression remains available. This applies the automatic-by-default-safe rule in `packages/cli/AGENTS.md`: optional compression must not change command execution semantics.

Validation: CLI build passed. Targeted shrink-hook tests passed 17 with four POSIX rewrite tests skipped on Windows. Regression cases exercise five host tool names, filenames with spaces, a successful diff, and a failing Git command through real PowerShell, preserving output and exit status. The formerly failing command also passed through an actual Windows Codex tool invocation after applying the guard. The full CLI suite was not run.


JuliusBrussee:
Early feedback since this is a draft. The mechanism is right — the rewrite is Bash-quoted (`'…node' '…index.js' shrink -- …`) and PowerShell parses that as a string expression, hence `Unexpected token shrink`.

Two things, though. Your regression cases loop over `exec_command` / `shell` / `shell_command`, and since #1037 landed `shrinkHook` already emits **nothing** for those three — verified against a build of current main. So the Codex reproduction can no longer occur, which leaves the premise unproven: which host reports a **PowerShell** command as `tool_name: Bash`? On Claude Code for Windows that tool is Git Bash, where the quoting is correct, so `process.platform === "win32"` also disables the hook for users it works for today.

My guess at the exposure that does survive #1037: Gemini CLI's `run_shell_command`, which on Windows goes through `cmd.exe` — and `cmd.exe` does not understand single quotes either. If that is your host, a `run_shell_command`-scoped fix would be narrow and provable.

Also, every new test is `skip: process.platform !== "win32"`, so CI verifies none of it. Whatever lands wants at least one assertion about the emitted string that runs on Linux.

---

JuliusBrussee:
**Real problem, needs changes on the approach.**

The premise is worth fixing — a host labelling a PowerShell command as `Bash` with no way to identify the shell is a genuine gap, and mangling someone's command is worse than not compressing it.

But `if (process.platform === "win32") process.exit(0)` turns shrink off for *every* Windows user, including the Git Bash majority where it works correctly today. That trades a correctness bug affecting some Windows users for a feature regression affecting all of them, and the `{ skip: process.platform === "win32" }` markers on the existing tests mean nothing would catch the loss.

What would land: detect the shell (or detect that you can't) and skip only that case, keeping the Git Bash path working. The existing tests should keep running on Windows to prove it.

Still marked draft, so this may already be where you were heading.

---

PR #1050, opened 2026-09-14
fix(cli): honor think.shrink in the native install, not just the launcher

### LINKED ISSUE #1049: Native install always adds shrink-hook; think.shrink=false is not honored by enable/doctor --fix
## Summary

`caveman enable <agent>`, `caveman doctor <agent> --fix` and `repairNativeAgent` always write the `caveman shrink-hook` PreToolUse entry into the agent's settings, regardless of the `think.shrink` config value. The only place `think.shrink` / `CAVEMAN_SHRINK` is consulted is the ephemeral launcher plugin (`caveman claude`), so there is no persistent way to run the native integration without the command-output rewrite.

## Where

`packages/cli/src/index.ts` (v1.3.3):

- `nativeMutationsFor` for Claude: `nativeHooksDocument("claude", true, settings)` — `includeShrink` is the literal `true` (around line 6932). Same for Gemini (`:7002`) and Codex (`:7391`).
- `nativeHooksDocument` (`:6434`) appends `${cavemanBinForHook()} shrink-hook` whenever `includeShrink` is true, with replace-not-accumulate semantics, so a manually removed entry comes back on the next repair.
- `think.shrink` is read into `noShrink` (`:3507`) and used only at `:5458` for `buildClaudeEphemeralPlugin`.

## Why it matters

Claude Code runs every matching PreToolUse hook in parallel and executes only one `updatedInput`. A user who already has a shell-command rewriter (in my case an `rtk hook claude` wrapper) gets it silently displaced by `shrink-hook`. On Codex, every `updatedInput` must carry `permissionDecision: "allow"`, so `shrink-hook` also widens the permission grant to everything on its allowlist (`gh`, `docker`, `kubectl`, `git push`, ...).

Removing the entry by hand works until `caveman doctor <agent> --fix` — which `caveman doctor` itself recommends as soon as the install is `degraded` — restores it. The only stable workaround today is `caveman disable <agent>`, which drops the lifecycle hooks and Core injection as well.

## Proposed fix

Have the native writer honor the persisted config: `nativeHooksDocument(agent, cfg.shrink, ...)` (or an explicit `--no-shrink` on `enable` / `doctor --fix` that is recorded in the integration journal), so `caveman config set think.shrink false` (or `CAVEMAN_SHRINK=0`) yields a native install with lifecycle hooks but no command rewrite, and repairs keep that choice.

## Repro

```bash
caveman config set think.shrink false   # or CAVEMAN_SHRINK=0
caveman enable claude
grep -c shrink-hook ~/.claude/settings.json   # expect 0, get 1
```

Related: #1037 (Codex re-prompt after the shrink rewrite) is a consequence of the same grant.

L4XB:
Closes #1049.

## What was wrong

`nativeMutationsFor` passed a literal `true` for `includeShrink` on all three native agents, so `caveman enable <agent>`, `caveman doctor <agent> --fix` and `repairNativeAgent` wrote the `shrink-hook` PreToolUse entry regardless of `think.shrink`. The only reader of that config (or `CAVEMAN_SHRINK`) was `buildClaudeEphemeralPlugin`, i.e. `caveman claude`. A hand-removed entry held until the next repair — and `caveman doctor` recommends that repair as soon as the install reads `degraded`.

The issue spells out why it is not just a preference: Claude Code runs every matching PreToolUse hook in parallel and applies **one** `updatedInput`, so an existing shell rewriter is silently displaced; on Codex every `updatedInput` must carry `permissionDecision: "allow"`, so the entry also widens the grant to everything on its allowlist.

## What changed

One helper reads the persisted value, and four call sites use it:

- `claudeNativeMutations`, `geminiNativeMutations`, `codexNativeMutations` — the three writers.
- **`nativeHookEntriesHealthy`** — the same helper, deliberately. Without it `doctor` keeps calling a correctly shrink-less install `degraded` and keeps recommending the `--fix` that used to reinstate the entry. That loop is most of the reported pain, and fixing only the writers would leave it in place.

Two sites keep the literal `true`, each with a comment saying why:

| site | why |
| --- | --- |
| `assertNativeHooksShape` | only reads event **names** to refuse a non-array; the superset is the safer set to check |
| `removeNativeHookEntries` | `disable` must withdraw an entry an earlier install wrote, whatever the config says now |

## What I wrote and then removed

A purge branch in `nativeHooksDocument` that dropped an already-present managed `shrink-hook` entry when the switch is off. It reads like the missing half of "replace, never accumulate" — but **no test could distinguish it**: `doctor --fix` rebuilds the document from the pre-install baseline, so the entry is never in `base` on that path, and the health check plus the repair already converge without it. I removed it rather than ship code no cell exercises. Happy to bring it back if you know a path that reaches it.

## Behaviour after this change

```
caveman enable claude                      # shrink-hook present (unchanged)
caveman config set think.shrink false
caveman doctor claude                      # degraded — honest: the install no longer matches the config
caveman doctor claude --fix                # entry removed, route + lifecycle hooks + Core kept
caveman doctor claude                      # installed
caveman doctor claude --fix                # no-op, not a reinstatement
```

`caveman enable claude` refuses while the install is degraded (pre-existing behaviour, and it names `doctor claude` in the message), so `--fix` is the route. The env form works the same and needs no repair because there is nothing to converge from:

```
CAVEMAN_SHRINK=0 caveman enable claude     # shrink-hook never written
```

## Tests

`packages/cli/tests/native-enable.runtime.mjs`, two cells, both starting from an accept control that proves the entry IS written with the rewrite on:

1. the persisted `think.shrink: false` path end to end — degraded → `--fix` → entry gone, `native-hook claude` and `ANTHROPIC_BASE_URL` intact → `installed` → a second `--fix` is a no-op → flipping back on restores it;
2. the `CAVEMAN_SHRINK=0` env form.

| mutation | red |
| --- | --- |
| writers back to the literal `true` (i.e. `main` today) | both new cells |
| health check back to the literal `true` | the persisted-config cell |

```
node --test tests/native-enable.runtime.mjs    34 pass, 0 fail
tsc --noEmit -p packages/cli/tsconfig.json     clean
```

One unrelated cell in the wider `pnpm test` run fails on this machine for a locale reason (`provider tokens 1.200 input` vs the expected `1,200`) — `de_DE` number grouping, present with or without this branch.


JuliusBrussee:
Adopted into #1015, cherry-picked so your authorship is preserved. This is the right shape: one reader means the writers and the health check that judges them cannot disagree, and your call on which three sites keep the literal `true` is correct — I kept both comments explaining why.

One case the tests could not reach: `enable` merges into the host's live file, and the shrink block only ran under `if (includeShrink)`, so an entry a standalone or pre-switch install already left in `settings.json` was never touched. With the switch off that meant exit 0, rewrite still live, and the fresh install born `degraded`. Follow-up commit extends the replace-not-accumulate rule to our own path when the switch is off, with a test seeding a foreign PreToolUse hook to pin that nothing else is withdrawn.

Also picked up the second half of the report: the degraded gate pointed at bare `caveman doctor <agent>`, which dead-ends. It names `--fix` now. Thanks — clean diagnosis and the comments were genuinely useful to read.

---

L4XB:
Thanks — and the `enable` case is a real gap in what I wrote, not a nicety. My cells drove the install path, where `includeShrink` is what decides whether the block runs at all; I never exercised the merge into a host file that already carried an entry from a standalone or pre-switch install. With the switch off that is exit 0, a live rewrite, and a fresh install born `degraded` — which is the failure the health check exists to catch, reached by the one route the check could not see. Seeding a foreign `PreToolUse` hook to pin that nothing else is withdrawn is the right shape for that test; a replace-not-accumulate rule is only safe if it can be shown not to replace anyone else's entry.

Happy to close #1050 whenever #1015 lands, or now if you would rather keep the queue short — the commit is in your branch with the authorship, which is what matters. Say the word either way.

JuliusBrussee:

PR #1052, opened 2026-09-14
fix(cli): caveman enable <agent> doesn't start the local proxy it just wired config to

### LINKED ISSUE #1051: Caveman Causing codex config.toml to stop working
After updating/installing caveman I noticed it also updated my codex config.toml

so when I try to start any conversation I'm getting 

"stream disconnected before completion: stream closed before response.completed"

This is because it added this line 

```
`# >>> caveman:native-root
model_provider = "caveman"
# <<< caveman:native-root`
```

Removing it does solve it but this is not normal behaviour

SyN-droMe:
Fixes #1051.

`caveman enable codex` writes config.toml to route Codex through the local caveman-proxy, but never starts or checks that the proxy is running. If nothing's listening yet (fresh install, reboot, or just running `enable` before ever launching through `caveman <agent>`), every Codex session after that hits a dead endpoint. The "stream disconnected before completion" error in #1051. The `caveman <agent>` shortcut already pre-starts the proxy after enabling; `enable` on its own never got that safety net.

Fix: best-effort, synchronous, fire-and-forget proxy spawn right after the config write succeeds. Not a reuse of `startWrapProxy`'s readiness-poll that loop's timer isn't `unref()`'d, so awaiting it would add up to a 2s hang to every `enable` call.

Verified: `pnpm lint` clean; `native-enable.runtime.mjs` 32/32 pass under WSL2 (fixtures don't run on native Windows, unrelated to this change); manual before/after on Windows confirms the proxy is spawned with the fix and isn't without it.


JuliusBrussee:
**Right gap, needs changes before it can land.** Confirmed the premise: on `main` the only two proxy spawn sites are `caveman start` and `startWrapProxy` (wrap), and the native SessionStart revive only fires once Codex has approved the installed hooks — so `enable codex` on a cold machine does leave config.toml pointing at a port nothing listens on. Nothing on the maintainer's pending branch covers it either.

Four things on the diff as sent:

1. **`CAVEMAN_RECOVERY` is inherited instead of stamped.** Both existing spawn sites set it explicitly in both directions (see the comment block in `startWrapProxy` — an exported `CAVEMAN_RECOVERY=mcp` must not survive into a proxy that has no retrieve tool to back it). `{ ...process.env, … }` here lets a stray export through, and the proxy publishes it into run state as `RecoveryViaMCP`, which is what a later wrap uses to decide whether it may claim compression. Stamp it the same way the SessionStart revive does for this agent.
2. **`CAVEMAN_PROXY_OWNER: "enable"` means "never exit".** The proxy treats every owner other than `wrap` as `start`, so this proxy skips the 30-minute native idle-exit that a hook-revived proxy gets. If that is intended (a persistent daemon from `enable`), say so; otherwise use `wrap` so it behaves like the revive it is standing in for.
3. **This is `startWrapProxy` minus the readiness poll, copied.** Extract the spawn (`cavemanBin` → `which` → `spawn` detached + `unref`) into one helper both call, and keep the poll out of this caller as you describe. The un-unref'd sleep timer is a fair reason not to `await` it; it is not a reason to duplicate the block.
4. **No test.** `native-enable.runtime.mjs` already drives `enable` against stub binaries; point `CAVEMAN_PROXY_BIN` at a recording stub and assert one spawn with the expected `CAVEMAN_LISTEN` and recovery value, and zero spawns when `wrapMode` is not `local`.

Minor: a second `enable` while a proxy is already up spawns a proxy that fails to bind and exits — harmless because run state is only written after the listener binds, but a `gatewayListening` probe first would make the common case free.

SyN-droMe:
Thanks for the detailed review. All four addressed, plus the optional 5th:

1. Extracted `spawnLocalProxyProcess()` out of `startWrapProxy` (env assembly + spawn, no readiness poll) so both call sites share one implementation. `CAVEMAN_RECOVERY` is stamped explicitly the same way in both, never inherited.

2. `enable` now passes `purpose: subscription ? "codex-subscription" : "standard"`, matching what `agentShortcut` already does right after calling `enableNative`,  so `CAVEMAN_PROXY_OWNER` ends up "wrap" in the normal case (30-min idle-exit, same as a hook-revived proxy), not the immortal "start" bucket.

3. Duplication gone. See (1).

4. Added a test to `native-enable.runtime.mjs`: points `CAVEMAN_PROXY_BIN` at a stub that logs its own env when spawned bare, plants a stray `CAVEMAN_RECOVERY` in the parent env to prove it doesn't leak through, and asserts `owner=wrap` + the correct `listen` address.

5. Added a `portListening` probe before spawning, so a second `enable` while a proxy's already up skips the redundant spawn.

33/33 in `native-enable.runtime.mjs` (32 existing + the new one above). It ran under WSL2, since the POSIX shell-script fixtures in this suite don't execute under native Windows `execFileSync`, which I confirmed is pre-existing and unrelated to this change before switching environments.

JuliusBrussee:
**Adopted into #1055** — all three commits cherry-picked, authorship preserved. The four review points from the previous round are genuinely addressed; nice turnaround.

Verified rather than taken on trust: your new test fails without the `index.ts` change (`not ok 8`) and passes with it, 33/33 across three consecutive runs with no flake. Also confirmed `env.CAVEMAN_MODE === mode` at the assignment site, so swapping the ready-message to `mode` is equivalent, not a behaviour change.

One follow-up, not blocking: the spawn is gated on `agent !== "aider"`, so aider keeps the dead-endpoint window this closes for everyone else. Deliberate, or worth a second look?

---

SyN-droMe:
Thanks for the review and for adopting this into #1055. Appreciate you re-verifying. Good catch on the noProxy / repeated-enable gaps. It makes sense to keep aider excluded given it has no MCP binary to back the recovery flag.

JuliusBrussee:
**Already landed — your fix is on `main`.** The 2026-09-15 triage merge (`ed37ab1`, #1055) carries `spawnLocalProxyProcess`, the split out of `startWrapProxy`, and the fire-and-forget spawn at the end of `enableNative` with the `portListening` pre-probe.

Nothing left to change. Flagged in #1082 for the maintainer to close.

One note for the record, in fairness to the review process rather than to your diff: I could not have verified this one independently in the triage environment. `packages/cli` needs a full TypeScript build and its `*.runtime.mjs` suite is not part of root `npm test`, so your test would not have run here. That is a gap in the triage setup, not in the PR.

---

PR #1056, opened 2026-09-15
fix(shrink): finish the shutdown sequence when the upstream ignores stdin EOF

Dhevenddra:
Follow-up to #995, which you asked for as its own PR: the stdin-EOF orphan I reported there but deliberately left out.

## The bug

`caveman-shrink` is an MCP **client** to the upstream it spawns, so teardown is the sequence the spec gives that client ([lifecycle, Shutdown > stdio](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#stdio)):

> 1. First, closing the input stream to the child process (the server)
> 2. Waiting for the server to exit, or sending `SIGTERM` if the server does not exit within a reasonable time
> 3. Sending `SIGKILL` if the server does not exit within a reasonable time after `SIGTERM`

`endInput()` does step 1 and stops there. #995 added steps 2 and 3, but only behind `process.on(SIGTERM|SIGINT|SIGHUP)`, which is the path where the host signals us. Closing our stdin is the *first* thing a host does to shut an stdio server down, and on Windows it is the only thing: the comment at `index.js:108` already says teardown there "goes through the stdin EOF", and that is exactly the path with nothing behind it.

So an upstream that does not exit on EOF keeps itself and the wrapper alive, and the host waits on both.

## Repro, against current main (4df4b03)

An upstream that ignores EOF and would honour a signal if one ever arrived:

```js
// stubborn-server.js
process.stdin.resume();
process.stdin.on('end', () => { /* deliberately ignored */ });
```

Spawn the wrapper around it, close the wrapper's stdin, wait 10s:

```
$ node repro.js src/mcp-servers/caveman-shrink/index.js
client closed the wrapper stdin
STILL ALIVE after 10000ms - wrapper hung, upstream orphaned
```

Same repro on this branch:

```
client closed the wrapper stdin
wrapper exited after 2844ms  code=143 signal=null
```

800ms to the EOF, one 2s grace, then SIGTERM. This upstream does not trap it, so the SIGKILL step was not needed. 143 is `128 + SIGTERM` through the existing `onClose` contract, and it is accurate: that upstream did not shut down on its own.

## The change

`closeInput()` passes the EOF on, then arms the same grace period and falls into the existing `forward('SIGTERM')`, which already escalates to SIGKILL. Same constant, same escalation, reached from either entry point.

`forward` moves out of the returned object literal to a named function so `closeInput` can call it without `this` inside a timer callback. The body is unchanged apart from folding the two `exitCode/signalCode` checks into a shared `upstreamGone()`.

## Nothing is signalled on the normal path

A server that exits on EOF fires `close`, which clears the wait before anything is sent. Measured end to end against a server that answers `tools/list` and exits on EOF:

| | main | this branch |
|---|---|---|
| exit | 452ms, code 0, no signal | 448ms, code 0, no signal |
| response | compressed identically | compressed identically |

## Tests

34 pass, 0 fail (28 on main). The six new ones use the injected timers from #995, so they stay synchronous.

Each was checked against a broken variant rather than taken on trust:

| mutant | caught by |
|---|---|
| EOF arms no wait at all | escalates EOF through SIGTERM to SIGKILL; a repeated EOF does not arm a second wait |
| EOF signals immediately instead of waiting | does not signal an upstream that exits on EOF (+3 more) |
| `onClose` leaves the EOF wait armed | close clears a pending EOF wait |
| EOF is never passed on to the upstream | passes client EOF on to the upstream; EOF on an already-exited upstream arms nothing |
| the repeat guard is dropped | a repeated EOF does not arm a second wait |
| a wait is armed for an already-dead upstream | EOF on an already-exited upstream arms nothing |
| the escalation goes straight to SIGKILL | escalates EOF through SIGTERM to SIGKILL |

No mutant survived, and the detector itself reports nothing against the unmutated file.

`npm test` is unchanged by this: 351 pass on both main and the branch, and the same 11 installer tests fail on both (one symlink case, nine `powershell.exe` manifest cases, one `pwsh.exe` case). None of them are in caveman-shrink, and the two failure lists are byte-identical.

## Limits, unchanged from #995

The escalation still reaches the direct child only. An upstream that is really a launcher can leave the server running as a descendant holding the inherited stdout pipe, which also keeps `close` from firing. Owning the whole process tree is the larger change you and I both described on #995 and it is not attempted here.

Also: `tests/test_mcp_shrink.js` is touched by #1055, but its hunk is at line 113 in the compress tests and mine start at 360, so the two merge without a conflict.


JuliusBrussee:
**Adopted into #1070** as-is, by cherry-pick so your authorship is preserved.

I re-ran your repro independently rather than taking the numbers on trust, both directions: pre-fix, `STILL ALIVE after 10000ms - wrapper hung, upstream orphaned`; with the commit, `wrapper exited after 2315ms code=143`. `tests/test_mcp_shrink.js` goes 28 → 35, all passing, and it cherry-picked onto current `main` with a clean auto-merge exactly as you predicted.

The spec citation is what sold it — framing the wrapper as an MCP *client* makes `closeInput` obviously step 1 of a sequence that already existed, rather than a new code path. Pulling `forward` out of the object literal so a timer callback can reach it without `this` is the right call too. And the mutation table meant I could check the tests were load-bearing instead of guessing.

Noted in the PR that the launcher/descendant limitation is unchanged from #995 and not attempted here.

---

PR #1058, opened 2026-09-15
fix(bedrock): splice cache markers instead of decode-remarshal

### LINKED ISSUE #1057: Bedrock cache-points optimizer corrupts integers >2^53 anywhere else in the request body
## What

`ApplyProviderNativeTransforms` in the Bedrock adapter's cache-points optimizer decodes the whole request body into `map[string]any`, mutates the map to add one `cachePoint`/`cache_control` marker, then re-marshals the entire document with `json.Marshal`. Go decodes untyped JSON numbers into `float64`, which only represents integers exactly up to 2^53 (9007199254740992). Any integer literal anywhere else in the body at or above that bound is silently rounded to the nearest representable double on the way back out, with no error and no signal.

A tool-result payload carrying a large id, a snowflake, a timestamp, or a monetary amount in cents has nothing to do with the cache marker being injected, but gets corrupted whenever the optimizer fires.

## Where

`proxy/providers/bedrock/cache_points.go`: `ApplyProviderNativeTransforms` decodes at line 53-54 (`json.Unmarshal(data, &root)`), the `injectConverseCachePoint`/`injectAnthropicCacheControl` helpers mutate the decoded map, and line 75 (`json.Marshal(root)`) re-marshals the whole document.

The rest of the codebase already avoids this: `proxy/providers/anthropic/cache_breakpoints.go`'s `injectBreakpointRaw` and `proxy/providers/openai/cache_key.go` decode only to make decisions, then splice the marker into the original byte slice via `proxy/providers/jsonsplice`, leaving every other byte untouched. `cache_points.go` has no splice attempt at all.

## Repro

```go
var root map[string]any
json.Unmarshal([]byte(`{"amount_cents":9007199254740993}`), &root)
out, _ := json.Marshal(root)
// out == `{"amount_cents":9007199254740992}`
```

Against the actual adapter method: a Converse request with a stable `system` prefix (so the optimizer fires) and a `toolResult` containing `amount_cents: 9007199254740993` comes back with `amount_cents: 9007199254740992` in the bytes that get signed and sent to Bedrock.

## Scope

Requires the operator to have the `bedrock-cache-points` optimizer enabled (off by default) and PAYG Converse/Invoke traffic on a cache-eligible Claude-on-Bedrock model with a stable tools/system prefix. Any deployment that does enable it (its documented purpose) is exposed on every matching request that happens to carry a >2^53 integer anywhere in its body, which is plausible for agentic tool-call traffic.

AmirF194:
Fixes #1057

## Root cause

`ApplyProviderNativeTransforms` decoded the request into `map[string]any`, mutated the map to add a cache marker, then re-marshaled the whole document. Go decodes untyped JSON numbers into `float64`, exact only up to 2^53, so any larger integer anywhere else in the body (a tool-result id, a timestamp) was silently rounded on the way out.

## Fix

Splice the marker into the original bytes instead, matching the pattern `proxy/providers/anthropic/cache_breakpoints.go` and `proxy/providers/openai/cache_key.go` already use via the `jsonsplice` package. `jsonsplice` had no array-append primitive (only `AppendObjectFields`, for inserting fields into an object), needed here because both injection paths append a whole new `cachePoint` element rather than augmenting an existing one, so this adds `AppendArrayElements` alongside it.

## Verified

- Added a regression test that asserts a >2^53 integer survives byte-for-byte, for both the Converse and Invoke injection paths. Without this fix, it fails.
- `go build ./...`, `go vet ./...`, and `go test ./...` all pass across the full repo, run in a golang:1.26-bookworm container.
- `gofmt -l` is clean on every changed file.
- I have not run this against the Windows or macOS legs directly; the change is plain `encoding/json` and `jsonsplice`, no OS-specific paths.


JuliusBrussee:
**Adopted into #1070** by cherry-pick, so your authorship is preserved. Verified your regression test fails on `main` and passes with the splice; full proxy suite green.

Good change — `AppendArrayElements` was the missing primitive and it matches the existing `AppendObjectFields` contract exactly. The string-`system` branch is better than what it replaced, since splicing the original bytes preserves the caller's escapes.

Two things I added on top, not corrections to yours: the same defect exists at three sibling sites that can't splice (each mutates a nested field), so those are fixed at the decode with `UseNumber`; and the repro showed the old path also **reordered the whole body**, so the cache optimizer was rewriting its own cache prefix.

Thanks — the issue write-up was what made this a 20-minute verification instead of a day.

---

AmirF194:
Thanks for catching the three sibling sites and the body-reordering side effect, both were real gaps in my fix. Glad the write-up saved you time on verification.

AmirF194:
Closing this one: the fix landed via #1070's cherry-pick, so main already has it.

PR #1060, opened 2026-09-15
fix(repointel): recheck the 500-symbol cap before every append, not after

### LINKED ISSUE #1059: repointel's 500-symbol cap doesn't hold across sibling declarations in the CGO_ENABLED=0 build
scanSymbols in symbols_nocgo.go tries to enforce the package's own 500-symbol-per-file cap by returning `len(out) < 500` from the ast.Inspect callback. That return value only controls whether Inspect recurses into that node's own children, it says nothing about later siblings. Top-level func/type declarations in a Go file are siblings in file.Decls, so once the cap trips inside one declaration, returning false only stops the rest of that declaration: every subsequent top-level decl in the file is still visited and appended.

A Go file with 600 top-level funcs comes back with 600 symbols under this build, not 500. symbols_cgo.go doesn't have this bug, it checks len(*out) >= 500 at the top of every recursive call before appending. build-release-binaries.mjs always sets CGO_ENABLED=0, so it's the broken nocgo path every shipped binary actually runs. Bites hardest on generated or vendored Go with lots of top-level declarations: protobuf stubs, mockgen output, big const tables.

AmirF194:
Fixes #1059

scanSymbols in symbols_nocgo.go checked the 500-symbol cap with `return len(out) < 500` from the ast.Inspect callback. That return value only prunes descent into the current node's own children, not later siblings, and top-level func and type decls in a Go file are siblings in file.Decls: once the cap is hit inside one declaration, returning false only skips the rest of that one, every later top-level decl still gets visited and appended. symbols_cgo.go already gets this right, checking len(*out) >= 500 before appending, and it's the nocgo variant that every shipped release binary actually runs (build-release-binaries.mjs always sets CGO_ENABLED=0).

Moved the check to the top of the closure, same shape as the cgo file.

Test builds a source string with 600 top-level funcs and asserts scanSymbols caps at 500: red on main, green here. Ran it under CGO_ENABLED=0 since that's the tag this file needs. The repo's `go test ./...` CI job links symbols_cgo.go by default, which was already correct, so this file isn't otherwise exercised under the build config that actually ships.


JuliusBrussee:
**Adopted into #1070** as-is, by cherry-pick so your authorship is preserved.

Verified both directions before taking it: your test returns 600 symbols for 600 top-level decls under `CGO_ENABLED=0` and passes under `CGO_ENABLED=1`, which pins the cgo/nocgo split rather than just the cap. Moving the check to the top of the closure is the right shape — it matches `symbols_cgo.go`, so the two implementations now fail the same way if either drifts again.

---

AmirF194:
Thanks for verifying both cgo and nocgo before merging. Appreciate you keeping the check symmetric with symbols_cgo.go, that was the right call.

AmirF194:
Closing this one: the fix landed via #1070's cherry-pick, so main already has it.

PR #1061, opened 2026-09-15
test(evals): keep semantic taxonomy IDs out of runtime

alexis-ag:
## Summary

- define stable semantic contract IDs in an eval-only taxonomy
- label source cases with readable keys and resolve IDs when preparing eval output
- reject duplicate IDs, unknown keys, embedded IDs, and duplicate cases
- verify that annotation preserves prompt text and that runtime `SKILL.md` contains no taxonomy IDs

## Why this boundary

This is the focused eval-only follow-up invited in #944. Stable IDs help compare failures across revisions, but putting them in runtime headings spends prompt context on test bookkeeping. Source cases therefore stay readable, and `evals/annotate_taxonomy.py` adds IDs only to evaluation output.

This PR does not change the runtime skill, generated plugin mirrors, or `dist/`.

## Validation

- `python3.12 -m unittest tests/test_eval_taxonomy.py -v`
- `python3.12 -m unittest discover -s tests -p 'test_*.py' -v` (146 tests)
- `python3.12 tests/verify_repo.py`
- `git diff --check`

The macOS system `python3` is 3.9 and cannot import an existing repository module that uses `str | bytes`; validation therefore used Python 3.12, matching the repository's Python 3.11+ CI lanes.

Follow-up to #944.


alexis-ag:
Supporting write-up is now live: [A Semantic Core Instead of a Phrasebook](https://relux.works/en/blog/semantic-core-instead-of-phrasebooks/). It records the #944 runtime-ID objection, corrects the earlier description of that unmerged candidate, and explains the eval-time taxonomy boundary together with the independent matrix-validation and runner-isolation follow-ups. Claims and limitations are separated explicitly.

PR #1062, opened 2026-09-15
test(evals): reject incomplete snapshot matrices

alexis-ag:
## Summary

- validate committed eval snapshots before reporting token metrics
- require complete metadata, prompts, both control arms, and at least one skill arm
- require exact prompt/output cardinality and raw string output in every arm
- cover missing controls, truncated arms, malformed output, metadata mismatch, and controls-only evidence with negative tests

## Why

The earlier #944 harness could accept incomplete evidence. This independent follow-up applies the same fail-closed principle to the existing output-length snapshot: a partial matrix must not produce a plausible-looking report.

No snapshot data, runtime skill, generated mirror, or distribution artifact changes in this PR.

## Validation

- `python3.12 -m unittest tests/test_eval_snapshot_contract.py -v` (7 tests)
- `python3.12 -m unittest discover -s tests -p 'test_*.py' -v` (148 tests)
- `python3.12 tests/verify_repo.py`
- `python3 evals/measure.py` using installed `tiktoken` 0.14.0
- `git diff --check`

The documented `uv run` wrapper was unavailable on this workstation; the report smoke used the equivalent direct Python entry point. CI remains authoritative for the documented wrapper.

Follow-up to #944.


evals/snapshot_contract.py
```diff
@@ -0,0 +1,89 @@
...
+        if not isinstance(arm, str) or not arm:
+            raise SnapshotContractError("arm names must be non-empty strings")
+        if not isinstance(outputs, list):
+            raise SnapshotContractError(f"arm {arm} must contain a list of outputs")
+        if len(outputs) != n_prompts:
+            raise SnapshotContractError(
+                f"arm {arm} contains {len(outputs)} outputs; expected {n_prompts}"
+            )
+        for position, output in enumerate(outputs):
+            if not isinstance(output, str):
```
AmirF194, line 76: I ran `evals/measure.py` in a clean container against a snapshot that passes `validate_snapshot`, at head `6c8a1d0`.

`snapshot_contract.py:76` only checks that each output is a `str`, not that it's non-empty. An all-empty-string `__baseline__` arm (plausible if `run_claude`'s stdout comes back blank) is accepted, and then `measure.py:89` divides by `sum(baseline_tokens)`, which is 0, crashing with an unhandled `ZeroDivisionError` instead of the "Invalid snapshot" message this PR is built to produce.

```
$ python evals/measure.py
Traceback (most recent call last):
  File "/repo/evals/measure.py", line 89, in main
    f"({fmt_pct(1 - sum(terse_tokens) / sum(baseline_tokens))} vs baseline)"
ZeroDivisionError: division by zero
```

The prompts check two lines up already rejects an empty string (`isinstance(prompt, str) and prompt`). The output check at line 76 could use the same pattern.

PR #1065, opened 2026-09-15
fix(middleware): reclaim a revoked scope's tombstone after its own grace period

### LINKED ISSUE #1064: middleware: revoked scope tombstones are never reclaimed, wedging the shared capacity cap
Revoking a session (`sessions/delete`) writes a tombstone that `Expire` can never reclaim, so revocation traffic alone can permanently exhaust the shared 100,000-row / 64MB capacity cap across every principal using the store.

`Delete` marks a revoked scope by setting `expires_at=0` (`proxy/internal/store/middleware.go`, `MiddlewareTx.Delete`). `Expire`'s reclaim query only ever selects rows where `expires_at>0 AND expires_at<=now` (`middlewareExpire`'s `dead` selection and its final scope-row delete). A row with `expires_at=0` can never satisfy `expires_at>0`, so a revoked scope's row, and the choice rows `purge()` leaves behind with their payload zeroed but the row intact, are never selected by any of `Expire`'s statements. They sit forever.

This is the same class of bug `c41b81f` fixed for elapsed scopes (an unreachable reclaim branch wedging the row cap), but for the other trigger: revocation, which is the protocol's own normal way to end a session early, needs no elapsed timer and no malformed input.

Reproduced against a real `Store`/`MiddlewareTx` (no mocks) on HEAD `4df4b03`:

```go
s, _ := store.Open(path, nil)
s.InitMiddleware(ctx)
s.WithMiddleware(ctx, func(tx *store.MiddlewareTx) error {
	tx.SaveScope(store.MiddlewareScope{ID: "s", Authority: "auth", Manifest: []byte("[]"), ExpiresAt: 1 << 40})
	return nil
})
s.WithMiddleware(ctx, func(tx *store.MiddlewareTx) error { return tx.Delete("auth") })
// 50 Expire() passes at an arbitrarily far-future time
for i := 0; i < 50; i++ {
	s.WithMiddleware(ctx, func(tx *store.MiddlewareTx) error { return tx.Expire(1 << 62) })
}
scope, err := ... tx.Scope("s") // still returns the row, err == nil, unchanged
```

The row survives every `Expire` call regardless of how far into the future `now` is. Filling a store past the row cap with only revoked scopes (no elapsed ones) reproduces `SaveScope` returning `ErrMiddlewareCapacity` forever, with no `Expire` pass ever draining it.

Go 1.26, `go test ./proxy/internal/store/...` at HEAD `4df4b03ba538751131d24415c59a1fad0d8c2c18`.

AmirF194:
Fixes #1064.

`Delete` marks a revoked scope with `expires_at=0`, and `Expire`'s reclaim query only ever selects `expires_at>0` rows. A revoked scope's row, and the choice rows `purge()` leaves behind with a zeroed payload, can never be selected, so every `sessions/delete` call leaves one permanent row against the shared capacity cap. This is the same wedge `c41b81f` fixed for elapsed scopes, for the trigger that fix didn't cover.

`Delete` now records the revocation time as `-now` instead of the bare `0`. `Expire` treats any `expires_at<=0` as immediately payload-dead (same schedule as an elapsed scope, so choices/plans/an authority's dropped credit reclaim right away) and its final scope-row delete reclaims a revoked row once `-expires_at` is past one `MiddlewareGraceSeconds` window, mirroring the elapsed-scope tombstone exactly. `previousPlan` and `Grant`, the two readers of the old `expires_at==0` sentinel, move to `expires_at<=0` so a scope still answers "deleted" for the whole grace window, never earlier.

Verification (Go 1.26, Docker, HEAD `4df4b03`):
- New `TestMiddlewareRevocationTombstoneSurvivesGraceThenReclaims`: with the old `Delete`/`Expire` the tombstone never leaves the store; with this change it survives one tick short of grace and is gone exactly at the deadline.
- New `TestMiddlewareExpiryReclaimsCapacityFromRevokedScopes` mirrors the repo's own elapsed-scope capacity test with revoked scopes as the sole trigger: seeds past the 100,000-row cap with `expires_at<=0` rows only, confirms admission stays capped, then confirms `Expire` drains the store to empty and admission recovers.
- `go build ./...`, `go vet ./...`, and `go test ./...` all pass repo-wide; a coverage run of `go test ./proxy/internal/store/... ./proxy/internal/middleware/...` confirms every changed line is exercised.
- Not run under coverage instrumentation: `TestMiddlewareAcceptanceTwentyTurnsAcrossProcesses` re-execs the test binary as a subprocess and fails whenever `-coverprofile` is enabled, a pre-existing interaction with this repo's own test that has nothing to do with this change. I measured coverage with that one test skipped instead.

A related, smaller gap I did not fix: an authority whose only scope was ever revoked (never elapsed) still keeps its `middleware_originals` credit row forever, since the drop for that row was never wired to a grace window at all. Flagging it rather than folding it into this diff.


JuliusBrussee:
**Adopted into #1070** by cherry-pick (authorship preserved), with one follow-up commit. The mechanism is right — storing revocation time as a negative `expires_at` is what makes the grace deadline computable at all, and the `<= 0` updates in `recovery.go`/`runtime.go` are the correct companions.

One thing it missed: the new `dead` selects revoked scopes *immediately*, and `dead` drives `DELETE FROM middleware_choices`. `recovery.retrieve` reads the typed `"deleted"` answer off the **choice** row via `Grant`, the same way `previousPlan` reads it off the scope row — which is why `Delete` keeps that row and only zeroes its payload. Since `Expire` runs in front of every optimize request, the first pass after a revocation deleted it, and `Grant` then returns `ErrNoRows` → `"not_found"` instead of `"deleted"`. Net effect was a zero-length revocation grace on the recovery path and a full week on the optimize path.

Your `TestMiddlewareRevocationTombstoneSurvivesGraceThenReclaims` doesn't catch it because it asserts on `tx.Scope`, not `tx.Grant`. I added a `Grant`-path test and made revoked rows wait out the same grace in `dead` that they already wait out in your scope delete. Nothing is collected early as a result — `Delete` calls `purge` synchronously, so the payloads are already gone and only the rows remain.

Your capacity test still passes unchanged. Nice catch on the bug itself; the `c41b81f` parallel was the right framing.

---

AmirF194:
Good catch, my test asserted on tx.Scope and missed the Grant-path read entirely, so the revocation grace was only half fixed. Thanks for tracing it through to Grant and adding the test that actually pins it.

AmirF194:
Closing this one: the fix landed via #1070's cherry-pick and the follow-up commit that closed the Grant-path gap, so main already has the full mechanism. Nothing left for this branch to carry.

PR #1072, opened 2026-09-16
fix(hooks): restore "go (back) to normal mode" deactivation

pedrohlmelo:
**Problem**
Saying `go back to normal mode` or `go to normal mode` leaves caveman active. The tracker keeps the mode flag and still injects the per-turn `CAVEMAN MODE ACTIVE (full)` reminder on that same turn, so the user gets no feedback that the request was dropped. `back to normal mode`, `switch back to normal mode` and `return to normal mode` all work. README and SKILL.md both promise that `normal mode` turns caveman off.

**Reproduce**
```
D=$(mktemp -d); printf full > "$D/.caveman-active"
echo '{"prompt":"go back to normal mode"}' | CLAUDE_CONFIG_DIR="$D" node src/hooks/caveman-mode-tracker.js
cat "$D/.caveman-active"   # before: "full" (still active)   after: file removed
```
Same result with `go to normal mode`.

**Cause**
Regex introduced in #598. The prompt-initial "normal mode" pattern in `caveman-parse.js` lists `go\s+` and `back\s+to\s+` as separate, mutually exclusive prefixes, so `go back to` matches nothing, and `go to` has no alternative at all. This is a regression: running the tracker from the commit before #598, both phrases deactivated caveman.

**Fix**
Make `back` and `to` independently optional after `go`: `go\s+(back\s+)?(to\s+)?`. One regex change. Mid-sentence mentions (`how do I exit vim normal mode`, `switch from normal mode to debug mode`) still do not deactivate. Both Claude Code and the opencode plugin consume this parser, so both get the fix. `src/hooks/checksums.sha256` regenerated.

**Tests**
New case in `tests/test_caveman_parse.js` fails on `main` and passes here. `npm test`, every `tests/test_*.js`, `python -m unittest discover -s tests` and `tests/verify_repo.py` all pass.

Written with AI assistance (Claude Code); reproduction and test runs were executed locally.


JuliusBrussee:
**Adopted into #1082** — cherry-picked, so your authorship and sign-off are preserved.

Confirmed on `main` first: `"go back to normal mode"` and `"go to normal mode"` both returned `null`, which is the worst shape a deactivation bug can take — silent, on the most natural phrasing, leaving the user unable to get out.

Your diff needed nothing added. Checked the two things I would have flagged: the regex stays prompt-anchored so `"how do I exit vim normal mode"` still does not deactivate, and `caveman-parse.js` is the single parser (the opencode plugin loads this same file via `new Function`, and Codex/Gemini/OpenClaw are prompt-driven), so there is no sibling host left blind. Checksums were already regenerated.

The maintainer merges; I don't.

---

JuliusBrussee:
Already on main — superseded.

`caveman-parse.js` currently carries the alternation this PR adds:

```js
/^(please\s+)?(go\s+(back\s+)?(to\s+)?|back\s+to\s+|switch\s+(back\s+)?to\s+|return\s+to\s+)?normal\s+mode\b/
```

I checked out this branch's `tests/test_caveman_parse.js` against the unmodified parser on main and all 55 tests pass, including your four new phrasings — so there's nothing left for #1089 to fix here.

Good catch originally; the "go" / "back to" alternatives being mutually exclusive was a subtle one.

---

JuliusBrussee:
**Not reproduced on current `main`** — the phrasings this PR targets already deactivate.

Driving `src/hooks/caveman-parse.js` on `main`, all of "go back to normal mode", "go to normal mode", "back to normal mode", "please go back to normal mode", "switch back to normal mode" and "return to normal mode" return `{action:'clear'}`. The false-positive guard still holds too: "how do I exit vim normal mode" returns `null`.

The prompt-initial `normal mode` branch already carries the optional `go (back) (to)` prefix.

If you have a spelling that still slips through, paste it and I'll check it next run — a concrete miss is easy to act on.

---

PR #1078, opened 2026-09-16
fix(engine): compress real-world C++ (ifdef sniff, recovered tokens)

sebastianrueckerai:
Stacked on #1077: the first commit here is that PR's, because both change the same parse gate in `Compress`. Happy to rebase once either lands.

## Problem

Two things kept a Godot GDExtension, and much other C++, out of the code compressor:

1. `#ifdef`, `#ifndef` and `typedef` all contain "def ", and the Python sniff was `has("def ") && has(":")`, so a guarded C or C++ file was parsed as Python and passed through.
2. Class-body macros without a trailing semicolon (`GDCLASS(X, Y)`) and typed anonymous enums (`enum : int64_t {`) make tree-sitter insert a token. `root.HasError()` was true and the whole file passed through with every method body intact.

## Change

1. The Python signal is now a `def name(` that starts a line.
2. `parseAccepted` replaces the `HasError` gate: no ERROR node anywhere, and MISSING nodes are tolerated only when the invented token is a `;` or a name, because the block structure the elision relies on was parsed from real tokens. A MISSING brace or parenthesis still rejects. The re-parse guard holds the output to at most as many MISSING nodes as the input, so an edit can never make the tree worse.

## Measured

o200k tokens on a real GDExtension: the kernel `.cpp` (1,449 lines, `GDCLASS`, two typed anonymous enums) 18,112 → 3,847. Its header is declarations only and stays 1,065. `register_types.cpp` still passes through: `GDE_EXPORT` in declaration position is a real ERROR node, and that rule is unchanged.

## Tests

`TestSniffIfdefIsNotPython`, `TestCodeCppToleratesRecoveredTokens` (macro line, enum and signatures survive, bodies elided), `TestCodeCppRealErrorsStillPassThrough`. `go test ./engine/...` passes.


sebastianrueckerai:
The red `go` check is `TestNewHTTPClient_ProxyConnectTunnelsHTTPSWithoutResolvingTarget` in `shared/platform/ssrf`, not this change: the PR touches only `engine/compressors`, the same diff passed that job on run 35130604888 and failed on 35130692891 and 35132609922, and the test passes locally 600 of 600 with and without `-race`. The tunnel handler in that test drops the buffered reader `Hijack` returns, so the first byte of the ClientHello is lost whenever it arrives between the header flush and the cancelled background read. Fix in #1079, one test file.

PR #1080, opened 2026-09-16
fix(learn): stop overcounting SessionsScanned and undercounting Agent spawns

### LINKED ISSUE #1075: learn: subagent_pressure never fires (spawn tool is now Agent, denominator counts every transcript)
`source_claude.go` counts `"name":"task"`, but spawns are now `tool_use` blocks named `Agent`: 135 to 379 real spawns a week in my transcripts, 0 detected (checked at ed37ab1).
`SessionsScanned` counts every transcript opened, not sessions with turns in the window (5,178 vs about 240 for one 7-day run), so pressure rounds to zero even with detection fixed.
Before repairing the count: is a spawn waste? Delegating to a cheaper model saves cost, and over the same month spawns per session fell while spawns per turn rose.
A measure that separates the two cases: the share of subagent tokens on the parent's tier or a more expensive one, plus the receipt size that re-enters the parent.

AmirF194:
`sessionEventConsumer.consume()` called `recordSession()` (SessionsScanned++) on every session's `sessionStart` event, before the per-line `since` cutoff in each source's `scanSession` ever runs. A transcript whose real turns all predate the analysis window still got counted as scanned, so `subagent_pressure = TaskSpawns / SessionsScanned / 5.0` divides by a number much larger than the sessions actually in range and the score component stays near zero regardless of real usage.

Moved the increment to the first turn event a source actually emits inside the window; `sessionStart` no longer counts anything. This is the shared path all five sources go through, so one change covers all of them.

Separately, `source_claude.go`'s `TaskSpawns` only matched `"name":"task"`. The subagent-spawn tool_use block is now named `Agent`, not `task`, so real spawns in a recent transcript went uncounted. Added the second name to the same match.

Added two tests in `source_claude_test.go`: one builds a session whose only turn predates `since` and checks it's excluded from `SessionsScanned`, the other checks an `Agent`-named tool_use block counts as a spawn. Both are red on main. `go build ./...`, `go vet ./...` and `go test ./...` are clean at the repo root.

I didn't build a fixture for the other four sources (codex, aider, opencode, gemini); they go through the same `consume()`, so the fix should cover them, but I only exercised claude's.

Fixes #1075


JuliusBrussee:
**Adopted into #1082** — cherry-picked, authorship preserved — **and extended.**

Both your tests fail on `main` for the right reason (`SessionsScanned = 2, want 1`; `TaskSpawns = 0, want 1`). I verified your central assumption before taking the `sessionStart` change, since `session_source.go` is the shared consumer: all five sources (`claude`, `codex`, `gemini`, `opencode`, `aider`) apply the `since` cutoff before emitting a real turn, and `c.opened` is still used by `finish()`, so moving `recordSession` to the first in-window event is safe. `aider` filters at session level and never emits `sessionStart` when stale, so it is a no-op there.

What it missed: the rename is hardcoded at **four** sites. `subagentSpendTracker.observeSpawn` and `claudeStructuredSkillReferences` still matched `"task"` alone, so subagent *spend* and `SpawnsByType` stayed at zero on exactly the transcripts your counter fix had just started detecting. Folded the name into one `isSubagentSpawnTool` predicate in a follow-up commit, with two more tests.

Left your raw-line counter as a substring count so it stays a cheap per-line heuristic; a comment ties the names it accepts to the predicate.

---

AmirF194:
Good catch on the other two call sites, `subagentSpendTracker.observeSpawn` and `claudeStructuredSkillReferences` were still on the old name and I missed both while focused on the counter. Folding it into one predicate is the right call. Thanks for verifying the `sessionStart` assumption against all five sources too, that's more than I checked before pushing.

AmirF194:
Closing, this landed via #1082 (merged, cherry-picked with authorship preserved and extended for the two call sites I missed). Nothing left to carry here.

PR #1084, opened 2026-09-17
feat(install): add native Grok Build support

### LINKED ISSUE #547: Add support for Grok Build agent
**What you want**

Please add Grok build support and Installation guide.

**Before/after example**
```
Before: [current behavior]
In INSTALL.md no guide in "Per-agent install" section for Grok build agent.

After:  [desired behavior]
In "Per-agent Install" section in INSTALL.md, Command for Grok build agent is also present.
```

**Why good**

Grok released "Grok Build" agent and It is a good agent (using composer 2.5 + super grok).

Thank you for your attention.

SomSamantray:
## Summary

The unified installer (`bin/install.js --only grok`) now installs owned skill copies for Grok Build natively into `GROK_HOME/skills` (default `~/.grok/skills`), following the same owned-copy convention as Continue/AiderDesk/iFlow. Grok Build appears in `--list` and in the install docs, with the vendor discovery matrix recording its sources and limits.

Closes #547

## Test report

- `npm test` (full `tests/installer` + `tests/hooks` suite): 388 tests, 375 pass, 0 fail, 13 skipped.
- `tests/installer/grok.test.mjs`: 4/4 pass.
- `tests/installer/provider-skills.test.mjs` + `provider-skills-integration.test.mjs`: 39/39 pass.
- `node bin/install.js --list` shows the grok row.
- All verified without a grok binary installed.

## Evidence strength

The `~/.grok` default and `GROK_HOME` override come from third-party setup docs; no live-binary run was possible, so no native loading success is claimed.

## Review findings — all resolved

- **P2** — `bin/lib/provider-skills.js:52` — a relative `GROK_HOME` installed from one working directory would be orphaned by `--uninstall` from another. Fixed by documenting the requirement explicitly in `INSTALL.md`: a relative override (this applies to the pre-existing Continue/AiderDesk/iFlow overrides too, not just Grok) must be uninstalled from the same working directory it was installed from, since the path resolves against `cwd`.
- **P2** — `tests/installer/grok.test.mjs:25` — the round-trip tests pinned a hardcoded 7-skill subset while the generic native path installs every `skills/` dir with a `SKILL.md`. Fixed by deriving the expected skill list dynamically via `readdirSync`, matching the pattern already used in `provider-skills-integration.test.mjs`.
- **P3** — `CLAUDE.md:360` — the agent-support table omitted Grok Build from the owned-copy agents row. Fixed by adding it alongside Continue/AiderDesk/Antigravity, including the `GROK_HOME` override note.

## CI note

The `go` check failure on this PR is in `shared/platform/ssrf/ssrf_test.go` (`TestNewHTTPClient_ProxyConnectTunnelsHTTPSWithoutResolvingTarget`), a TLS-handshake-timing test unrelated to this diff — this PR touches only the JS installer, docs, and installer tests. `main` is green on the same suite, confirming this isn't a regression introduced here. Watching the CI run on the latest push to confirm it clears (or reproduces as a flake unrelated to this diff).


SomSamantray:
## CI status after the latest push

All checks pass except `go`, which fails on the same pre-existing test both before and after this PR's fix commit:

```
--- FAIL: TestNewHTTPClient_ProxyConnectTunnelsHTTPSWithoutResolvingTarget (0.00s)
    ssrf_test.go:824: CONNECT request failed: Get "https://example.com/v1/models": EOF
FAIL	github.com/JuliusBrussee/caveman/shared/platform/ssrf	0.018s
```

This PR's diff touches only `bin/install.js`, `bin/lib/provider-skills.js`, `INSTALL.md`, `docs/technical/installer-provider-discovery.md`, `CLAUDE.md`, and `tests/installer/*.mjs` — no Go code. The `shared/platform/ssrf` package and its tests are untouched.

I don't have admin rights on this repo to re-run the job directly (`gh run rerun` refused with "Must have admin rights to Repository"), but recent CI history on unrelated branches shows this same test flaking and clearing on the very next run without a code change — e.g. `fix/clearsessionprev-symlink-delete-guard` failed at run `35260260628` and passed at `35260894990` a few minutes later with no commits in between. That's consistent with a runner-timing/local-TLS-handshake race in the test itself, not a regression from this change.

Flagging so a maintainer can re-run the `go` job when convenient — happy to push again if a fresh run is preferred instead.

JuliusBrussee:
**Reviewed, not adopted — needs a maintainer decision, not a change from you.**

The engineering here is solid: install, uninstall, dry-run and cwd-relative-override tests, the whitespace-preservation case, the discovery-matrix row, and `SKILLS` derived from the directory rather than pinned. The uninstall-symmetry test is exactly the right instinct.

Two reasons it isn't something this run lands. First, it implements a feature request (#547), and this daily-triage run triages feature requests rather than shipping them. Second, and the one that actually needs the maintainer: you're upfront that no live binary was available, so `GROK_HOME` and `~/.grok/skills` come from third-party setup docs rather than verified vendor behaviour. `CLAUDE.md` treats the install table as product surface — "One broken install command costs real user" — so that row wants one confirmation against a real Grok Build install before it ships.

Flagged for decision in #1089. Nothing to fix meanwhile; if you can get output from a real `grok` install showing where it reads skills from, that's the single thing that would unblock it.

---

PR #1086, opened 2026-09-17
fix(hooks): delete the per-session prev flag through safeDeleteFlag

### LINKED ISSUE #1085: clearSessionPrev deletes through a symlinked parent with a bare unlinkSync
`clearSessionPrev` in `src/hooks/caveman-config.js` deletes both the per-session and legacy prev-mode files with a bare `fs.unlinkSync`:

```js
function clearSessionPrev(claudeDir, sessionId) {
  const p = sessionPrevPath(claudeDir, sessionId);
  if (p) {
    try { fs.unlinkSync(p); } catch (e) {}
    return;
  }
  try { fs.unlinkSync(path.join(claudeDir, PREV_BASENAME)); } catch (e) {}
}
```

`safeWriteFlag` refuses to write a flag through a parent symlink owned by another user, but a bare `unlinkSync` follows that same symlink without any check, so a file the write path would never have created there can still be deleted through it.

#1055 fixed the identical asymmetry for `writeSessionMode`'s legacy-mirror delete by routing it through `safeDeleteFlag`, which is already exported from this file. `clearSessionPrev`'s two delete sites never got the same treatment.

Practical impact is limited: nothing currently reads either prev file back as a trust signal, so this is the same class of hardening #1055 already covered, not a new attack in itself.

AmirF194:
`clearSessionPrev` deletes the per-session and legacy prev files with a bare `fs.unlinkSync`, so it follows a parent symlink `safeWriteFlag` would refuse to write through. #1055 already fixed this same write/delete asymmetry for `writeSessionMode`'s legacy mirror by routing it through `safeDeleteFlag`; this does the same for `clearSessionPrev`'s two call sites, both of which had the identical gap.

`safeDeleteFlag` is already exported from `caveman-config.js`, so this is a drop-in swap, no new mechanism.

Added a source-code audit test (holds on every runner) plus a behavioral one that stages a symlinked `.caveman-sessions` dir owned by another uid and asserts the file survives (needs chown privileges, skips otherwise, same pattern as the existing `writeSessionMode` symlink test). Ran the full `tests/test_*.js` suite and `npm test` in Docker across node 18/20/22: my new tests fail on the pre-fix code and pass after, nothing else regressed. `npm test` has 11-16 pre-existing failures on main unrelated to hooks (`profile scope ...`), unaffected by this change either way.

Not touched: `caveman-mode-tracker.js`'s own `removeFlag` (bare `unlinkSync`) is a fallback used only when `caveman-config.js` fails to load or lacks these exports, so it can't call `safeDeleteFlag` from the sibling it's standing in for without duplicating the symlink-safety logic. Leaving it as is unless that's wrong.

Fixes #1085


JuliusBrussee:
Adopted into #1089, cherry-picked so your authorship is preserved.

Verified before/after: the behavioural test fails on main for the right reason (the delete follows the foreign-owned symlinked sessions dir) and passes with the fix; the source audit holds on runners without chown privileges. `src/hooks/checksums.sha256` re-checked against file contents after the pick.

No changes needed. Nothing further for you to do here — leaving this open for @JuliusBrussee to close or merge as they prefer.

---

AmirF194:
30366ff is in main now (via #1089, authorship preserved). Closing, this one's redundant.

Issue #1081, opened 2026-09-16
caveman learn doesn't work on OpenCode 1.18.31


DSBloom:
**What happen**
Caveman learn reports only Claude usage. Not OpenCode.

**Expected**
Read my OpenCode session history.

**Before/after example**
```
Input: $ caveman learn
Got:  caveman learn scanned 38 local Claude Code sessions from the last 30 days.
Result: no repeated token-heavy block appeared in at least 3 sessions, so it has no optimization recommendation yet.
Want:  OpenCode usage stats.
```

**Platform**
caveman 1.3.4
opencode 1.18.31
caveman learn --sources opencode --plain finds 0 sessions
OpenCode database exists at ~/.local/share/opencode/opencode.db
caveman doctor opencode reports installed components active but version_probe_failed

**Version / install method**
Installed via install.md

Root Cause
Caveman’s OpenCode learner expects the old file-based store:
~/.local/share/opencode/storage/session/.../*.json
Your OpenCode 1.18.31 uses a SQLite database instead:
~/.local/share/opencode/opencode.db
The database has the session data Caveman needs:
- session: session metadata, directory, provider/model and aggregate token counts
- message: per-message payload and timestamps
- part: text and tool-call payloads
Caveman therefore sees no JSON sessions and silently reports zero OpenCode sessions.

JuliusBrussee:
**Not reproduced — needs info, but your root-cause reading matches the code.** `source_opencode.go` walks a file-based JSON session store; nothing in it opens a SQLite database, so an install that keeps sessions in `~/.local/share/opencode/opencode.db` yields zero sessions and reports that silently. That also explains `--sources opencode` finding 0 while `caveman doctor` still shows components active.

I did not write a fixture or a fix, because I would have to invent the `session` / `message` / `part` schema and would then be testing my guess rather than your opencode.

What makes this reproducible here:

```
sqlite3 ~/.local/share/opencode/opencode.db '.schema session' '.schema message' '.schema part'
```

plus one redacted row per table (token counts and timestamps matter, prompt text does not). With that I can build a fixture DB and fix it against a real shape.

Two notes: `version_probe_failed` from `caveman doctor` is a separate detection miss worth its own issue if it persists. And `modernc.org/sqlite` is already a dependency, so reading the DB needs no new one.

---
_Generated by [Claude Code](https://claude.ai/code)_

JuliusBrussee:
[...]

PR #1087, opened 2026-09-18
fix(opencode): generate a V2 native plugin on OpenCode 2 hosts

### LINKED ISSUE #1083: OpenCode 2: generated native plugin uses incompatible V1 API
## What happened

The native plugin generated by `caveman enable opencode` fails to load in OpenCode 2.0.5. The installed generator still emits a V1 named export returning a Hooks object.

The matching OpenCode server log entry reports:

```text
2026-09-17T09:01:23.371Z
role=server
ref=err_ac6c5897
message="failed to load plugin"
target=~/.config/opencode/plugins/caveman-native.js
cause="PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup function. (cause: SchemaError(Missing key at [\"default\"])))"
```

The installed `@caveman-ai/cli/dist/index.js` generator contains:

```js
// Contract: @opencode-ai/plugin 1.17.8 Hooks (installed local type source).
export const CavemanNative = async () => ({
  // event, chat.message, experimental.chat.system.transform,
  // tool.execute.before, tool.execute.after, experimental.session.compacting, dispose
});
```

An isolated import of the generated file succeeds, but its only export is `CavemanNative`; `typeof module.default` is `undefined`. The failure occurs at plugin-definition validation, before native hooks or Caveman subprocesses run. Both Caveman executables exist, the CLI runs, and the separately configured Caveman MCP server connects successfully.

## Expected

Generate a plugin compatible with the detected OpenCode major version, or report unsupported V2 explicitly instead of installing a plugin that cannot load.

V2 requires a default definition with a stable `id` and `setup(ctx)` or `effect`. Adding a default export alone is insufficient: V1 returned hooks must be ported to domain hook registrations, event subscriptions, and cleanup. A no-op setup would hide the load error without restoring functionality.

Official migration documentation: https://opencode.ai/v2/docs/build/plugins/migrate-v1

## Reproduction

1. Use OpenCode 2.0.5 and Caveman CLI 1.3.4.
2. Run `caveman enable opencode`.
3. OpenCode discovers `~/.config/opencode/plugins/caveman-native.js` automatically.
4. Inspect the server plugin status and matching server log entry.

## Platform

- Other: OpenCode 2.0.5 on macOS
- Caveman CLI: 1.3.4, installed as `@caveman-ai/cli` in Bun's global packages
- Caveman binary release: bin-v1.1.7
- Installed OpenCode plugin packages: `@opencode-ai/plugin` and `@opencode/plugin`; their presence does not translate V1 plugin code

## Suggested fix

Port the generator to the V2 API, preserving prompt/context/compaction hooks, before/after tool processing, event handling, and cleanup. Account for changed event payloads, structured system and tool content, and the `bash` to `shell` tool rename. Add a generated-plugin loading test and focused behavior tests against a supported OpenCode V2 release. If retaining V1 support, use separate implementations or the documented dual-version entrypoint.

ranokay:
Fixes #1083.

`caveman enable opencode` emitted the V1 hook map on every host. OpenCode 2 rejects it at load time (`PluginModule.LoadError`, missing `default`), confirmed against a live 2.0.5 server log (`ref=err_ac6c5897`), so the native integration is dead on arrival there.

This change probes `opencode --version` at enable time (reusing the existing `nativeHostProbe`/`parsedSemver` helpers) and emits the implementation matching the host major:

- OpenCode 1.x keeps the existing V1 template untouched.
- OpenCode 2+ gets a ported `{ id: "caveman-native", setup(ctx) }` plugin preserving every native capability: session lifecycle, prompt profiling, stable/dynamic context injection, pre/post tool observation, shrink rewrite, and compaction hooks, adapted to domain hook registration, structured system parts and tool results, the `session.compaction.ended` event, and both `bash`/`shell` tool names.
- An unreadable version gets the current API rather than a plugin that cannot load.

Verified:

- `tsc --noEmit` clean; full CLI `build` passes.
- `tests/native-enable.runtime.mjs`: existing OpenCode 1.x test now locks the V1 branch; new test drives the generated V2 plugin end to end (setup registration, event lifecycle, prompt profiling with continuation flags, context injection, shell rewrite, output replacement, compaction, cleanup). Both pass.
- The generated V2 file was also validated against a live OpenCode 2.0.5 server: it loads with the plugin reported `active` and zero load failures (local backport; upstream generator output is byte-equivalent in behavior).
- Two `codex` proxy-spawn tests fail in this sandbox both with and without this change (pre-existing, environmental).

Follow-up, out of scope here: `src/plugins/opencode/plugin.js` (the `bin/install.js --only opencode` path) is V1-only too and will need its own port.


JuliusBrussee:
Adopted into #1089, cherry-picked so your authorship is preserved. Thanks — the version gate and the test that actually drives `setup()` end to end (event subscription, all three session hooks, both `bash` and `shell`, cleanup) made this straightforward to verify. Confirmed the new test fails on main for the right reason and passes with the change.

One follow-up committed on top, which I'd value your read on: the unreadable-version case now falls back to **V1** instead of V2. `nativeHostProbe` returns `version: null` for an empty/non-zero/unspawnable `opencode --version` and reports it as `version_probe_failed` — and #1081 records exactly that state on a live OpenCode **1.18.31** host. So "unknown" isn't evidence of "new": defaulting it to V2 hands a 1.x user whose probe merely flaked a plugin their host can't load, while a 2.x user in that state is no worse off than this issue already describes.

Your #1088 is reviewed separately — not adopted in this run, details there.

---

PR #1088, opened 2026-09-18
fix(opencode): dual V1/V2 entrypoint for the installer plugin

ranokay:
Stacks on #1087 (this branch contains #1087 plus one commit; merge after #1087). Same root cause as #1083, other install path: the `install.js --only opencode` plugin (`src/plugins/opencode/plugin.js`) was V1-only and fails to load on OpenCode 2 (`PluginModule.LoadError`, missing `default`).

The default export now serves both hosts from one file, following the documented dual-entrypoint shape:

- V2 calls `setup()`: prompt/context domain hooks plus an event subscription. Reinforcement injection grew a parts-aware twin for V2 structured system parts with the same idempotency contract.
- V1 (>= 1.18.29, which supports object entrypoints) calls `server()` with the untouched legacy hook map. Older V1 releases only accept a factory function and cannot load this file; no single file serves all three, and this is noted on the export.
- Flag writes, mode parsing, drift degradation, and the createRequire/caveman-config loading trick are shared and unchanged.

The installer also ships `index.js` (same bytes as `plugin.js`): OpenCode 2 auto-discovers package directories by their `index.js` entrypoint and ignores `package.json` `main`, while 1.x keeps loading the explicit `plugin.js` config entry, so one layout serves both hosts with no version detection. The legacy `plugin` config entry is silent on V2; V2 file entries warn, so the entry keeps pointing at `plugin.js`.

Verified:

- `tests/installer/opencode.test.mjs`: 16/16 pass. The behavior suite now drives the V2 hooks (prompt/context/subscription); a new compact test locks the dual shape (`id`/`setup`/`server` plus one legacy round-trip); the installer asserts `index.js` ships byte-identical and uninstall removes it.
- `tests/verify_repo.py` passes.
- Live-verified against OpenCode 2.0.7: the directory plugin loads `active` with zero failures, `setup()` writes the mode flag on session init, and the legacy `plugin` file entry stays silent. Verification scaffold removed afterwards; no live-config changes remain.


JuliusBrussee:
Not adopted in today's run (#1089) — needs a maintainer decision, not changes from you.

Your first commit is #1087, which #1089 does adopt. The second one changes `src/plugins/opencode/plugin.js`'s default export from a factory function to `{ id, setup, server }`, and as your own comment notes, object entrypoints need opencode >= 1.18.29 — so it drops pre-1.18.29 V1 hosts. That's a support-policy call for @JuliusBrussee, not something a triage run should decide.

Worth knowing: #1089 bumps the opencode pin from 1.18.27 to **1.18.31** (probed), which removes the specific objection that the floor sat above the pinned target. So this is more viable now than when you filed it.

The duplicated `index.js` entrypoint in `bin/install.js` looks independently sound, and the parts-aware `injectReinforcement` keeps the same idempotency contract as the V1 body.

---

Issue #1011, opened 2026-09-07
Invariant summary sheds facts without saying so


rjmendez:
summarizeElided drops entries until the summary fits summaryBudget, and renderInvariants joins whatever is left. Nothing records that anything went, so a trimmed `all ...` list is indistinguishable from a complete one.

Constants are shed 4th of 6 (invariants.go:379), which is the awkward part, since each one is a claim about every elided unit.

Repro is 120 log lines with 6 constant fields, then adding class-enumeration fields to put the budget under pressure:

    extra enums   constants still in the marker
    0             6/6
    1             5/6
    2             2/6
    3             1/6

Same marker shape every time, no sign of the drop.

Reading a field's absence as "wasn't constant" is most of what the marker is good for, so this is where it can mislead. A `; +N omitted` suffix would cover it. Can send a patch if useful.

JuliusBrussee:
Reproduced — your table lands exactly (constants go 6/6 → 5/6 → 2/6 → 1/6 as enumeration pressure rises, same marker shape throughout).

Fixed in #1012 with the `+N omitted` suffix you suggested. Your framing is what made it clearly a defect rather than a tuning question: `enumerateField` already refuses a partial value list one level down for precisely this reason, so the shed loop was breaking a rule the file states about itself.

One design note. I charged the notice to the same budget as the facts rather than exempting it. Exempting reads better in isolation, but it breaks both caps that govern a marker — the 160-byte ceiling and the rule that a summary never costs more than half the bytes it replaced — and a marker that stops shrinking what it stands in for is the one thing it may never do. Shed priority is unchanged; an untrimmed summary is byte-identical to before.

No patch needed, but thanks for offering — the repro table did the work.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1095, opened 2026-09-21
fix(engine): stop non-finite values riding the numeric track

pedrohlmelo:
**Problem**

A CSV or logfmt export spells a missing measurement `NaN` — pandas, Postgres `numeric`, R — and `strconv.ParseFloat` accepts it, along with `Inf` and `Infinity`. The value parses without error, so the field stays on the numeric track; every comparison against NaN is false, so the unit neither widens the range nor takes the field off that track.

It is elided into a run `summarizeElided` then describes as `range amount=lo..hi` — a bound that unit does not satisfy — and the compressed table comes out **byte-identical** to the same table holding a clean number in that cell:

```
region,status,note,amount
eu,ok,settled invoice for the quarterly subscription renewal,110
eu,ok,settled invoice for the quarterly subscription renewal,111
eu,ok,settled invoice for the quarterly subscription renewal,112
… 56 rows elided (caveman): all region=eu status=ok; range amount=113..168 …,,,
…
"… caveman: elided rows resemble shown rows and match the stated invariants — answer from this view; if you truly cannot, one broad caveman_retrieve …",,,
```

One of those 56 rows had `amount=NaN`. Nothing in the view says a row had no amount at all, under the contract line that tells the agent to answer from this view rather than recover. That is the claim `invariants.go` opens by forbidding — a marker may never imply a fact it did not verify — and the same shape as #1011.

**Reproduce**

```
go test ./engine/compressors -run 'NonFinite|KeepNumericExtremaIgnoresNonFinite'
```

Three tests, all failing before this change. The end-to-end one prints the two byte-identical outputs on failure.

**Cause**

`ParseFloat` is used as the "is this a number" test at all three sites in the package. It is not that test: it accepts the non-finite spellings, and NaN's comparison behaviour then does the rest.

`keepNumericExtrema` has the same root cause and loses more. It latches `minIndex`/`maxIndex` on the first numeric cell and only displaces them on a strict comparison, so the first NaN in a column pins both ends and no later row can ever move them — the column's real extrema stop being kept, which is that pass's whole job (`[NaN 10 70 30]` keeps only the NaN row). End to end the damage is usually masked, because `keepNonRedundant` runs afterwards and independently rescues outlier rows; I could not build a case where the extrema actually disappear from the output. The defense is switched off in silence either way, which is why it is fixed here rather than reported separately.

**Fix**

One `invariantNumber` helper, used by all three sites, that counts only finite values as numbers. A non-finite value leaves the numeric track and the field falls through to enumeration or coverage, which count every unit (`amount: 500×26 NaN×1`), so the NaN becomes visible instead of vanishing. A run whose values are all `NaN` is still constant and still reported as such — that much is verified. Finite numbers are untouched: `range amount=100..400` still renders exactly as before.

Ten lines of logic, at the one choke point the three callers share.

**Tests**

`TestNonFiniteValuesLeaveTheNumericTrack` (NaN mid-run, NaN last, `Infinity`, `-Inf`, and a NaN in an enumerable run, plus the all-NaN-is-still-constant and finite-numbers-unchanged cases), `TestKeepNumericExtremaIgnoresNonFinite`, and `TestTabularNonFiniteCellIsNotAbsorbedIntoARange`, which pins that a NaN row may not compress to the same bytes as a numeric one. All three fail on `main` and pass here.

`go test ./...` (0 failures across the repo), `npm test` (353 pass, 0 fail), `python -m unittest discover -s tests` (141 tests, OK) and `tests/verify_repo.py` all pass. `gofmt` and `go vet` clean.

Written with AI assistance (Claude Code); the reproduction, the fuzzing pass over the TOON round-trip that ruled out neighbouring defects, and every test run above were executed locally.


JuliusBrussee:
**Reviewed and independently verified — recommending merge as-is.** Nothing needed from you.

I checked the claim rather than trusting the write-up: reverting `engine/compressors/{invariants,tabular}.go` to main while keeping your tests, `TestTabularNonFiniteCellIsNotAbsorbedIntoARange` fails and prints the two byte-identical outputs — the NaN row really does compress to the same bytes as `amount=120`. Restoring the fix turns all three green, and `go test ./engine/... ./proxy/... ./shared/...` is clean across 62 packages.

Your framing holds up too, including the honest note that the `keepNumericExtrema` damage is usually masked by `keepNonRedundant` downstream — fixing it at the shared choke point is the right call regardless.

I deliberately did **not** cherry-pick this into today's triage PR (#1103): it is standalone, already carries its tests, and bundling it would duplicate your work without letting the PR be closed properly. It should land on its own with your authorship. Over to @JuliusBrussee.

---

PR #1107, opened 2026-09-21
fix(caveman-compress): run-aware inline-code pairing survives mid-line backtick runs

dilchenko:
## Problem

A backtick run in the middle of a prose line — e.g. `Put the result inside a ```diff fence.` — is matched by neither `FENCE_OPEN_REGEX` (not at line start) nor `FENCE_MARKER_LINE_REGEX` (not a whole-line marker), so its backticks leak into inline-code pairing. The naive `` `([^`]+)` `` pattern opens a span at the run's trailing backtick and closes it at the next real span's opener, shifting every following pair.

On a real ~/.claude/CLAUDE.md containing that sentence, `extract_inline_codes` produced 42 garbage multi-line "spans" of plain prose. The compressor can never reproduce those, so validation failed on every attempt and the file was permanently uncompressible — same failure family as #820, different leak path (mid-line run instead of indented fence marker).

## Fix

Pair spans the CommonMark way: a span opens with a backtick run and closes with an equal-length run.

```python
return [m.group(2) for m in re.finditer(r"(?<!`)(`+)([^`]+)\1(?!`)", text_without_fences)]
```

An unclosed run matches nothing instead of poisoning downstream pairs. Multi-line spans still extract (the #820 fix's contract), fence stripping and marker blanking are untouched, and the change is applied to both copies (`skills/` and `plugins/caveman/skills/`).

## Tests

New `TestMidLineFenceRun` in `tests/test_validate_inline.py`: pairing unaffected by a mid-line run, self-validation passes, and a span dropped after a mid-line run is still an error. Full suite: 144 tests, OK.

Verified end to end: the previously uncompressible CLAUDE.md now compresses and validates cleanly.


skills/caveman-compress/scripts/validate.py
```diff
@@ -263,7 +263,11 @@ def extract_inline_codes(text):
         "" if FENCE_MARKER_LINE_REGEX.match(line) else line
         for line in text_without_fences.split("\n")
     )
-    return re.findall(r"`([^`]+)`", text_without_fences)
+    # Run-aware pairing (CommonMark): a span opens with a backtick run and
+    # closes with an equal-length run. A mid-line ``` in prose ("inside a
+    # ```diff fence") has no equal closer, so it matches nothing — instead of
+    # its trailing backtick shifting the pairing of every following span.
+    return [m.group(2) for m in re.finditer(r"(?<!`)(`+)([^`]+)\1(?!`)", text_without_fences)]
```
AmirF194, line 270: I ran `extract_inline_codes` against `fa736f3c7` in a clean container.

The content class in the new pattern, `[^`]+`, still forbids any backtick inside a span, so a shorter complete backtick pair nested inside a longer delimiter's interior gets matched first and splits the outer span instead of becoming its literal content. That is the same "permanently uncompressible" failure class this PR fixes, just triggered by a different input shape. It happens on the exact idiom this PR's own body uses to show a backtick-quoted token:

```
>>> extract_inline_codes("The naive `` `([^`]+)` `` pattern is bad.")
['([^']
>>> extract_inline_codes("see `` `a` `` here")
['a']
```

CommonMark's rule is "closed by the next run of *equal* length," not "closed by the next run of any length with no backtick in between." A `[^`]+` content class can't express that, since it excludes backticks unconditionally.

A fix probably needs to split on backtick runs first (`re.split(r"(`+)", text)`) and walk the pieces, pairing each opening run with the next run of the same length and treating everything between as literal content, including any shorter/longer runs in it.

JuliusBrussee:
**Adopted** — cherry-picked into the open daily-triage PR #1103 with your commit and tests intact, plus `Co-authored-by`.

I reproduced the mis-pairing before taking it: on `main`, `extract_inline_codes` returns `['diff fence. Then state:\n\n- count of ', ' matches\n- check ']` for your example — two stretches of ordinary prose, and neither real span. Run-aware pairing fixes it, and the three tests fail without the change.

One change: I dropped the edit to `plugins/caveman/skills/caveman-compress/scripts/validate.py`. That path is the CI-synced mirror of `skills/`, which CLAUDE.md says not to hand-edit — `sync-skill.yml` republishes it from the source on merge. Worth knowing for next time; the `skills/` + `tests/` half was exactly right.

---

AmirF194:
Thanks for cherry-picking this in. Good to know about the plugins/caveman mirror, I did not realize that path gets regenerated from skills/ on merge.

PR #1108, opened 2026-09-22
chore(proxy): delete the dead routing package

JuliusBrussee:
`proxy/routing` (Router interface, FrugalGPT cascade, session-value policy validator, benchmark helpers; 2,833 lines across 10 files) has no importer anywhere in the module. The standalone proxy never rewrites a model, and the live consumer of this code is the private Caveman router service, whose copy is about 1,500 lines ahead. A stale duplicate here only invites a third fork.

No behaviour change: `go build ./...`, `go vet ./proxy/...`, `go test ./proxy/... ./shared/...` are green with the package gone.


JuliusBrussee:
Verification note: `go test ./proxy/... ./shared/...` was green except one run of `proxy/internal/nativehook` TestRepositoryStateChangesForExternalEditIndexAndWorktree ("initial repository state unavailable"). It passes on untouched `main` and passed on two consecutive re-runs here; the package does not import `proxy/routing`. Pre-existing flake under parallel test load, not related to this deletion.

Issue #1105, opened 2026-09-21
Cache-breakpoint planner never stabilizes for Claude Code native integration — every request cache-busts


alex-libra:
Running `caveman-proxy` (mode: compress) wrapping Claude Code (Windows). Every `/v1/messages` request logs:
```
"derived cache epoch prefix diverged; forwarding original bytes"
"session frozen-cache prefix did not extend prior request; possible cache bust"
```
back-to-back on the same `session_id`, across multiple sessions, persisting after repairing MCP recovery (`caveman mcp install claude --server caveman`, which fixed the separate "MCP recovery missing" issue and flipped native integration `claude` from `unavailable` → `degraded · newer_unknown`).

Effect: `caveman_stats` reports `tokens_before:0, tokens_after:0, requests:0` even with `compress on` and heavy usage — no compression is landing, and the repeated cache-prefix divergence means Anthropic prompt caching is being defeated on every turn (net token cost increase, not decrease).

Environment: Windows 11 (64-bit), Claude Code installed natively (`claude.exe` v2.1.278) at `~/.local/bin`, `@caveman-ai/cli` v1.3.4, `diverging_component_index` consistently `1`.

Reproduction: any multi-turn Claude Code session with the proxy wrap active; check `~/.caveman/proxy.log` for the WARN pairs above.

easy21865-jpg:
> Running `caveman-proxy` (mode: compress) wrapping Claude Code (Windows). Every `/v1/messages` request logs:
> ```
> "derived cache epoch prefix diverged; forwarding original bytes"
> "session frozen-cache prefix did not extend prior request; possible cache bust"
> ```
> back-to-back on the same `session_id`, across multiple sessions, persisting after repairing MCP recovery (`caveman mcp install claude --server caveman`, which fixed the separate "MCP recovery missing" issue and flipped native integration `claude` from `unavailable` → `degraded · newer_unknown`).
> 
> Effect: `caveman_stats` reports `tokens_before:0, tokens_after:0, requests:0` even with `compress on` and heavy usage — no compression is landing, and the repeated cache-prefix divergence means Anthropic prompt caching is being defeated on every turn (net token cost increase, not decrease).
> 
> Environment: Windows 11 (64-bit), Claude Code installed natively (`claude.exe` v2.1.278) at `~/.local/bin`, `@caveman-ai/cli` v1.3.4, `diverging_component_index` consistently `1`.
> 
[...]

Issue #1094, opened 2026-09-20
Claude Code interactive: derived cache epoch gate busts at the system component every turn, 0 tokens compressed across 117 requests


srad:
Claude Code interactive: `derivedEpochAllows` busts at the `system` component on every request after the first, so compression is skipped session-wide (0.0% reduction)

## Summary

On Claude Code subscription (OAuth) in **interactive** mode, every request after the first in a
session logs `derived cache epoch prefix diverged; forwarding original bytes` with
`diverging_component_index: 1`. Compression is skipped for the whole session. `caveman stats`
reports `before_tokens == after_tokens` and `saved_tokens: 0` over 117 requests.

This looks like the symptom reported in #1066, which was closed on 2026-09-15 as "Retracted by
author. Original diagnosis was incorrect." There are no comments on that issue, so it is not
possible to tell whether the symptom was invalid or only the explanation was. Filing fresh with
measurements.

It is **not** #1020. There the engine is handed nothing and `compression_tokens_before` stays at 0.
Here `before_tokens` advances to 8,984,616, so the engine is reached — it just returns the input
unchanged because the gate refused the request.

## Environment

- Windows 11 Education, x64
- `@caveman-ai/cli` 1.3.4, `binary_release` bin-v1.1.7
- **proxy binary reports `"version":"dev"`** in `~/.caveman/run/8787.json`, not a tagged release.
  Possibly relevant; flagging in case it changes which code is running.
- Claude Code 2.1.278
- Claude subscription OAuth, no API key — `auth_mode: "subscription"` on every request
- `caveman.yaml`: `mode: compress`, `subscription_compress: live_zone`, `breakpoint_plan: frontier`,
  `toolschema_strip: false`
- `caveman status` → `no off-states — everything the layer can do is on`

## Symptom

`caveman stats --json` totals over 30 days:

```json
"requests": 117,
"input_tokens": 10200906,
"cache_read_tokens": 9481087,
"cache_write_tokens": 717897,
"cache_write_1h_tokens": 717897,
"measured_requests": 103,
"before_tokens": 8984616,
"after_tokens": 8984616,
"saved_tokens": 0,
"observe_would_save_tokens": 0
```

`before_tokens` and `after_tokens` are identical to the byte. 103 requests were measured, so the
comparison path runs; it simply never produces a smaller payload.

## Proxy log

One pair of lines per request, on every request after the first in a session:

```
[...]

PR #1110, opened 2026-09-22
fix(proxy): match JSON-escaped session markers; skip home dir and .Trash in repo walk

mnott:
Two proxy fixes found while running the native Claude Code integration on a machine whose agent sessions start in the home directory.

## 1. Session markers never matched, so no request was exactly correlated

`markerPattern` in `proxy/internal/nativeruntime/marker.go` expected `sid="…" sig="…"` with raw quotes. The marker is injected into a user message and reaches the proxy inside a JSON string, where every quote is `\"`. The regex therefore never matched:

- `StripSessionMarkers` returned no session id and the gateway fell back to `unique_recent_session` / `unique_recent_time_model`;
- receipts reported `not_attested_due_approximate_session_correlation`;
- the prefix monitor logged `session frozen-cache prefix did not extend prior request; possible cache bust` on every request (233 of 233 in one run).

The existing test built an invalid JSON body with unescaped quotes, so the escaped form was never exercised.

Fix: accept the raw and the JSON-escaped form as two full alternatives (RE2 has no backreferences, and an optional backslash would let `sid` and `sig` disagree and leave a stray byte). New test embeds a real marker via `encoding/json`, asserts the id is returned, the body is still valid JSON and no longer contains the marker.

Observed after swapping the binary on a live proxy: `session_correlation_basis` went from `unique_recent_time_model` to `signed_marker` and `cache_bust` from 1 to 0 on the next request.

## 2. Repository intelligence walked the whole home directory, including `~/.Trash`

With cwd = home and no git checkout, `listFiles` in `proxy/internal/repointel/index.go` fell back to `walkFiles` over the entire home tree. A discarded `Xcode.app` in `~/.Trash` then produced "Likely implementation path" items pointing at SDK headers in the trash on every prompt, several hundred tokens of noise per turn.

Fix:
- the home directory is treated as not a repository (`ListingHomeSkipped = "home_directory_not_a_repository"`, no files, no walk), consistent with the comment above `listFiles` that an empty answer must not be reported as a complete map;
- `.trash` is added to `dependencyDirectoryNames` (matched case-folded, so `.Trash` is covered).

Tests cover both: a walk over a temp root must not list `.Trash/x.h`, and `listFiles` with root = HOME (via `t.Setenv`) returns zero files with the new reason. After the change the prompt hook reports "no path claims injected".

## Checks

`go test ./proxy/internal/nativeruntime/... ./proxy/internal/repointel/... ./proxy/internal/gateway/...` passes. Commits are DCO signed.


JuliusBrussee:
**Adopted — both commits cherry-picked into #1103 with your authorship and tests intact.** Thank you; this is the highest-blast-radius fix in that PR and it arrived with no issue filed against it.

Reproduced the marker miss on `main` first by building the body with `encoding/json`: `StripSessionMarkers` returns `session=""`, and the marker is forwarded to the provider verbatim because it is never stripped. Your RE2 reasoning is exactly right and I kept it in the comment — no backreferences, so the two forms must be full alternatives or `sid` and `sig` could disagree and leave a stray byte.

Worth noting for the record: this is a **second, independent** root cause for the `possible cache bust` warnings in #1105 (the first, #1094, was fixed by `99224f7`). Without exact correlation, the session id falls through to the `unique_recent_*` heuristics, which bind a main thread and a subagent to one id — which is what the prefix monitor then sees as a non-extending prefix.

On the home-dir walk: `gitListFiles` still runs first, so a dotfiles repo at `$HOME` keeps its git listing. Verified.

---

PR #1112, opened 2026-09-22
fix(middleware): isolate async observe() to its own receipt pool

### LINKED ISSUE #1111: AsyncMiddlewareRuntime.observe() shares optimize()'s executor, so a receipt burst can starve it into a false deadline bypass
`AsyncMiddlewareRuntime.observe()` sends receipts through the same 4-worker
executor as `optimize()`/`ready()`/`retrieve()`, instead of the dedicated
single-worker pool `observe_background()` uses. A burst of concurrent
receipts can occupy all 4 workers long enough that a concurrent `optimize()`
call never gets a worker before its own deadline expires, and bypasses with
`reason="deadline"` without ever attempting its capabilities fetch.

The sync API already isolates this correctly: `MiddlewareRuntime.observe_background()`
submits to its own lazily-created `ThreadPoolExecutor(max_workers=1)`, gated by
a separate semaphore, so receipt traffic can never delay `optimize()`. The
async wrapper's `observe()` instead calls `self._submit(self._runtime.observe, receipt)`,
which goes through `self._executor` (the same pool `optimize()` uses) and
`self._slots`, the shared admission semaphore.

Built and ran this from source (HEAD `3ee70a1`), faking only the HTTP boundary:
4 concurrent `observe()` calls with a slow receipts endpoint occupy all 4
shared workers, and a concurrent `optimize()` call started right after them
comes back `reason="deadline"` in about 2s, its capabilities fetch never even
starting. Routing the same 4 receipts through `observe_background()` instead
leaves `optimize()` unaffected.

AmirF194:
Fixes #1111.

`AsyncMiddlewareRuntime.observe()` was routing through `self._submit()`,
the same 4-worker executor and admission semaphore `optimize()`,
`ready()`, `retrieve()` and `delete_session()` share. A receipt burst
could occupy every worker long enough that a concurrent `optimize()`
call sat queued until its own deadline had already passed, so it never
even reached its capabilities fetch.

`MiddlewareRuntime.observe_background()` already solves this on the
sync side with its own lazily-created single-worker pool and a separate
semaphore. `observe()` now delegates to it directly instead of
duplicating the isolation.

Added `test_observe_burst_does_not_starve_concurrent_optimize`, styled
after the existing `test_queued_work_counts_against_one_deadline`: four
concurrent `observe()` calls hold a fake slow receipts endpoint, then a
concurrent `optimize()` is started and given time to miss its deadline
while queued. Without the fix it comes back `reason="deadline"`; with
observe() isolated, it doesn't. Ran the full `packages/sdk/python` suite
(277 tests) on Python 3.13 in Docker, matching the
`python (packages/sdk/python)` CI job's own install and pytest command.


JuliusBrussee:
**Diagnosis and reproduction adopted; fix written differently in #1103, with you credited as co-author.** Your analysis is correct and I reproduced it exactly as described — `deadline_ms=200`, four receipts held at the HTTP boundary, and an unrelated `optimize()` comes back `reason="deadline"`.

Why not the delegation as written: routing async `observe()` into `observe_background()` makes the coroutine return without awaiting delivery, which makes the two public methods identical and silently drops a receipt when the semaphore is full. The TypeScript twin (`packages/sdk/typescript/src/middleware/runtime.ts`) already solves this the other way — one serialized delivery chain, a separate 16-record cap, delivery still awaited, with the comment *"Receipt bursts must neither occupy all optimizer slots nor delay native output"* — and the two SDKs are required to mirror each other.

So on the branch `observe()` keeps its `await` and gets a dedicated single-worker pool with its own `BoundedSemaphore(16)`, through a `_submit_to()` that `_submit()` also routes through; both pools shut down in `_shutdown_now()` and `aclose()`. Your isolation point stands unchanged — only the mechanism differs.

---

AmirF194:
Good call moving the delivery to a dedicated pool instead of the delegation I wrote. Thanks for the credit; closing this one since #1103 covers it.

